PR #323, opened 2024-02-26
Price Impact Helper

brunoguerios:
# Description

This PR aims to create a helper contract to calculate price impact for add liquidity unbalanced operations.

Price impact is currently being done off-chain, within the SDK, through the ABA method (described [here](https://www.notion.so/Price-Impact-49545b95b98243e3ae624be6bcea2700))
For most operations (e.g. swaps, remove liquidity single token, etc..) this is ok to be done off-chain, because they contain a respective "reverse" operation and ABA method requires only 2 on-chain queries (A -> B and B -> A) to calculate price impact.
Add Liquidity Unbalanced is the only operation that requires an adapted version of ABA because Remove Liquidity Unbalanced is not available in v3.
More details on the adapted ABA logic can be found [here](https://www.notion.so/Price-Impact-49545b95b98243e3ae624be6bcea2700?pvs=4#419a8635787a4c7db14528198495d46a)

A few items still open for discussion:
1. I created a new package called standalone-utils to place this new helper contract. The goal was to have something similar to what we had on monorepo-v2. If there's any better alternative approach, feel free to suggest.
2. I initially thought of interfacing with the Router, in the same way the SDK does, but after some feedback from @mkflow27, I decided to remove the intermediary and query the vault directly. This has led to some code duplication. Let me know if you think there's a better way of doing this.
3. I created `standalone-utils` as a new package by copy/paste from another existing package. It might have some dependencies (or other things) that are not needed. Could use a help on identifying what is ok to remove or not.

Depends on #383.

## Type of change

- [ ] Bug fix 
- [x] New feature 
- [ ] Breaking change 
- [ ] Dependency changes
- [ ] Code refactor / cleanup
- [ ] Other

## Checklist:


pkg/standalone-utils/contracts/CallAndRevert.sol
```diff
@@ -0,0 +1,39 @@
+// SPDX-License-Identifier: GPL-3.0-or-later
+
+pragma solidity ^0.8.24;
+
+import { Address } from "@openzeppelin/contracts/utils/Address.sol";
+
+import { RevertCodec } from "@balancer-labs/v3-solidity-utils/contracts/helpers/RevertCodec.sol";
+
+contract CallAndRevert {
```
jubeira, line 9: maybe we can make a library out of this? There's no state to save or to access, so a library would seem better suited. Agree about making it a standalone module.

pkg/standalone-utils/contracts/PriceImpactHelper.sol
```diff
@@ -0,0 +1,251 @@
...
+        uint256[] memory exactAmountsIn,
+        address sender
+    ) external returns (uint256 priceImpact) {
+        uint256 bptAmountOut = _queryAddLiquidityUnbalanced(pool, exactAmountsIn, sender);
+        uint256[] memory proportionalAmountsOut = _queryRemoveLiquidityProportional(pool, bptAmountOut, sender);
+
+        // get deltas between exactAmountsIn and proportionalAmountsOut
+        int256[] memory deltas = new int256[](exactAmountsIn.length);
+        for (uint256 i = 0; i < exactAmountsIn.length; i++) {
+            deltas[i] = int(proportionalAmountsOut[i]) - int(exactAmountsIn[i]);
```
jubeira, line 38: Nit: let's use full types.
```suggestion
            deltas[i] = int256(proportionalAmountsOut[i]) - int256(exactAmountsIn[i]);
```

jubeira: (here and elsewhere)

pkg/standalone-utils/contracts/PriceImpactHelper.sol
```diff
@@ -0,0 +1,251 @@
...
+        for (uint256 i = 0; i < exactAmountsIn.length; i++) {
+            deltaBPTs[i] = _queryAddLiquidityUnbalancedForTokenDeltas(pool, i, deltas, sender);
+        }
+
+        // zero out deltas leaving only a remaining delta within a single token
+        uint256 remainingDeltaIndex = _zeroOutDeltas(pool, deltas, deltaBPTs, sender);
+
+        // calculate price impact ABA with remaining delta and its respective exactAmountIn
+        // remaining delta is always negative, so by multiplying by -1 we get a positive number
+        uint256 delta = uint(-deltas[remainingDeltaIndex]);
```
jubeira, line 52: I'd still just use safe cast here. Everything is static so we don't care about gas, and we have plenty of bytecode to spare.

pkg/standalone-utils/contracts/CallAndRevert.sol
jubeira: Ah this is to reuse the hook right? Then I guess it's fine as it is.

pkg/standalone-utils/contracts/PriceImpactHelper.sol
```diff
@@ -0,0 +1,251 @@
...
+    /*******************************************************************************
+                                    Helpers
+    *******************************************************************************/
+
+    function _queryAddLiquidityUnbalancedForTokenDeltas(
+        address pool,
+        uint256 tokenIndex,
+        int256[] memory deltas,
+        address sender
+    ) internal returns (int256 deltaBPT) {
```
jubeira, line 161: I think I don't get the back and forth here. Why is this a signed value if we always change the result to be positive?

pkg/standalone-utils/test/foundry/PriceImpact.t.sol
```diff
@@ -0,0 +1,116 @@
...
+        uint256 spotPrice = infinitesimalAmountIn.divDown(infinitesimalBptOut);
+
+        // calculate effectivePrice
+        uint256 bptOut = router.queryAddLiquidityUnbalanced(pool, amountsIn, address(0), bytes(""));
+        uint256 effectivePrice = amountIn.divDown(bptOut);
+
+        // calculate expectedPriceImpact for comparison
+        uint256 expectedPriceImpact = effectivePrice.divDown(spotPrice) - 1e18;
+
+        // assert within acceptable bounds of +-1%
```
jubeira, line 110: hm isn't 1% quite large for this? Does it break with a lower tolerance?

jubeira (commented):
Thanks for pushing this @elshan-eth !

First pass done. Only a few minor questions and comments; otherwise LGTM!

pkg/standalone-utils/contracts/PriceImpactHelper.sol
joaobrunoah: It took me a while to see, but we check if delta is positive or negative, and then we return `result` or `-result`. `result` can still be negative.

pkg/standalone-utils/test/foundry/PriceImpact.t.sol
joaobrunoah: The issue is the calculation of the spot price. Since we're using add liquidity unbalanced to calculate it, it's very sensitive to the "infinitesimal amount in". I added some comments in the test explaining this better, but now the error of this test is around 0.7%.

joaobrunoah (approved):
Nice @elshan-eth ! LGTM!

jubeira (approved):
LGTM!

PR #696, opened 2024-06-27
add optimization for copyToScale function when memory array double init

### LINKED ISSUE #336: Consider scaling given amounts in place after calling `before` hooks
See https://github.com/balancer/balancer-v3-monorepo/pull/329#discussion_r1505892926.

Applies to before add / remove / initialize. This should be just a small gas optimization.

elshan-eth:
# Description

We can save gas if we don't create a new array when we use copyToScale... functions. With this PR we can reuse this array.

## Type of change

- [ ] Bug fix 
- [ ] New feature 
- [ ] Breaking change 
- [ ] Dependency changes
- [ ] Code refactor / cleanup
- [X] Other

## Checklist:

## Issue Resolution

Closes #336


EndymionJkb (commented):
Change looks ok, but it puts the Vault over bytecode; is it worth it? It should make the gas go down a bit, but can't really see it with all the other changes. Would have been a smaller diff if this had just branched off #731.

Also needs a description. PRs should always have descriptions, or you have to track down the issue and comment to see what it's doing.

elshan-eth:
no optimization

PR #710, opened 2024-07-02
Fix coverage reports

joaobrunoah:
# Description

Coverage reports doesn't work with via-IR flag enabled, so a few changes are required:
* Disable via-IR flag when coverage report is running;
* Do not save gas measurements and contract sizes because coverage reports are not optimized.

## Type of change

- [ ] Bug fix 
- [ ] New feature 
- [ ] Breaking change 
- [ ] Dependency changes
- [x] Code refactor / cleanup
- [ ] Other

## Checklist:


pvt/common/hardhat-base-config.ts
```diff
@@ -13,11 +13,14 @@ type SolcConfig = {
   };
 };
 
+// Coverage report don't work well with via-ir flags, so we disable it
```
EndymionJkb, line 16:
```suggestion
// The coverage report doesn't work well with via-ir flags, so we disable it
```

EndymionJkb (approved):
Looks like it works, even at the top-level.

It looks like we're back to a single script with soft links, so you can remove "Each pkg should have its own coverage.sh" from the description.

Also, it leaves the info-e files behind. Do we want that? If so, should add to the .gitignore

joaobrunoah:
> Looks like it works, even at the top-level.
> 
> It looks like we're back to a single script with soft links, so you can remove "Each pkg should have its own coverage.sh" from the description.
> 
> Also, it leaves the info-e files behind. Do we want that? If so, should add to the .gitignore

info-e files are added to .gitignore.

PR #723, opened 2024-07-02
Vault Pruning (Move BasePoolFactory)

EndymionJkb:
# Description

`BasePoolFactory` really sounds like something that should go in pool-utils. The main issue was PoolFactoryMock's dependence on it. The mocks in the Vault are very tightly coupled, to the point where it's really not practical to try to move them.

The wire I found to cut that didn't blow everything up was reimplementing the necessary parts of BasePoolFactory in PoolFactoryMock, so that it would not be dependent on it. I also needed to move the pause window base contract (which had no dependencies of its own) down to solidity-utils. Those changes enabled BasePoolFactory to move to pool-utils with minimal disruption.

## Type of change

- [ ] Bug fix 
- [ ] New feature 
- [ ] Breaking change 
- [ ] Dependency changes
- [X] Code refactor / cleanup
- [ ] Other

## Checklist:


EndymionJkb:
> Do we even need the disable functionality in the pool factory mock?

I don't think we use it, but I have to implement it as part of IBasePoolFactory, so might as well have it work.

Issue #724, opened 2024-07-02
Remove custom StorageSlot


EndymionJkb:
Use the released OZ version. It turns out we don't actually use the released version directly, but rather extend it with bytes32 types, etc. And the SlotDerivation thing we really need is still unreleased.

So a permanent solution has to wait until it comes out (probably Q3: hopefully before we launch, or we're going to be stuck with it). Until then, we can at least address the name clash by renaming it to StorageSlotExtension, and removing the unused parts. See #726

EndymionJkb:
Resolved with #726. (Not sure why this didn't close automatically.)

PR #726, opened 2024-07-02
Openzeppelin StorageSlot collision

EndymionJkb:
# Description

We are not actually using StorageSlot as currently released. Our version of it added support for Int256, but mostly defined bytes32 types and operations using them. We do use a version of SlotDerivation, but this is not yet in the OZ package we're using, so should not conflict.

Eventually, per OZ, they will release a full version including SlotDerivation. At that point, we can revisit / refactor our code to use it directly. Until then, this is a minimal-change update that should solve the name collision. Essentially, we rename "our" StorageSlot to StorageSlotExtension, and remove functions defined by OZ's native StorageSlot (which aren't used anyway).

## Type of change

- [ ] Bug fix 
- [ ] New feature 
- [ ] Breaking change 
- [ ] Dependency changes
- [X] Code refactor / cleanup
- [ ] Other

## Checklist:

## Issue Resolution

Addresses #724


jubeira (approved):
LGTM, thanks @EndymionJkb !

PR #729, opened 2024-07-03
Aggregate fee precision

### LINKED ISSUE #602: Review precision of aggregate fee computation
See [this comment](https://github.com/balancer/balancer-v3-monorepo/pull/561#discussion_r1619140441).

The aggregate fee percentage is computed based on two components: a protocol fee, and the pool creator fee. These are full precision numbers (uint256), though 18-decimal FPs. The computation in the ProtocolFeeCollector is full precision, yet the resulting value is truncated to 24-bit precision in the Vault.

This introduces some imprecision when fees are split. Should they be truncated in the Protocol Fee Collector as well, so that when they're stored in the Vault as 24-bits, it is lossless?

EndymionJkb:
# Description

Fees are stored at full 18-decimal 64-bit precision at the ProtocolFeeController (and calculations are performed on full words), but the aggregate fees are stored with 24-bit precision at the Vault. This corresponds to 0.00001% resolution on fees (which seems like plenty, even for computed values).

As it was, setting fees with a precision greater than this would cause a discrepancy between the ProtocolFeeController's computations (e.g., when splitting fees), and the Vault. If you tried to set a fee of 1.000005%, it would be truncated to 1% in the Vault, so that if you set the fee and then immediately retrieved it, you would not get the same value.

This PR checks for this case at the ProtocolFeeController, and reverts if you try to set a fee with "too high" precision.

## Type of change

- [ ] Bug fix 
- [X] New feature 
- [ ] Breaking change 
- [ ] Dependency changes
- [ ] Code refactor / cleanup
- [ ] Other

## Checklist:

## Issue Resolution

Resolves #602


pkg/vault/contracts/ProtocolFeeController.sol
```diff
@@ -188,8 +189,15 @@ contract ProtocolFeeController is
...
-        return protocolFeePercentage + protocolFeePercentage.complement().mulDown(poolCreatorFeePercentage);
+    ) internal pure returns (uint256 aggregateFeePercentage) {
+        aggregateFeePercentage =
+            protocolFeePercentage +
+            protocolFeePercentage.complement().mulDown(poolCreatorFeePercentage);
+
+        // Ensure it is not too high precision
+        if ((aggregateFeePercentage / FEE_SCALING_FACTOR) * FEE_SCALING_FACTOR != aggregateFeePercentage) {
+            revert IVaultErrors.FeePrecisionTooHigh();
+        }
```
jubeira, line 200: How about truncating and storing the truncated value instead?

jubeira (commented):
Yeah good point; I don't think it makes sense to accept greater precision in the user facing interface.

I think reverting might be a bit too harsh though. Maybe just truncating and storing the truncated value is enough?

pkg/vault/contracts/ProtocolFeeController.sol
EndymionJkb: I thought about that, but then it's kind of silently failing - not storing the fee you're sending it (which might be calculated and have high precision). If you really think that's ok, I can do it that way.

joaobrunoah: As I mentioned in the call this morning, maybe someone fork-testing us would be confused by seeing rounding issues in the collected fees and would take a while to realize it's because of 24-bit precision. If we move in that direction, maybe it's nice to document in the external function that the fee is stored as a 24-bit number. (Afaik, it's only in the internal function, but not in the setters that expose an API with a 256 bits number input)

danielmkm: I'd kind of lean towards the original implementation of reverting. Even though the impact is minimal, silently failing feels like the less desirable approach. The user has provided a bad input, and you don't tell them, you just correct it for them leaving them to think the value provided was correct.

joaobrunoah (approved):
LGTM!

PR #731, opened 2024-07-03
Add rest benchmark tests

elshan-eth:
# Description

This PR adds benchmarks test for BatchRouter and pool mock with hooks. Also, test files is renamed.

## Type of change

- [ ] Bug fix 
- [ ] New feature 
- [ ] Breaking change 
- [ ] Dependency changes
- [ ] Code refactor / cleanup
- [X] Other

## Checklist:


pkg/vault/contracts/test/EmptyPoolHooksMock.sol
```diff
@@ -0,0 +1,121 @@
...
+import { IVaultMock } from "@balancer-labs/v3-interfaces/contracts/test/IVaultMock.sol";
+import "@balancer-labs/v3-interfaces/contracts/vault/VaultTypes.sol";
+
+import { FixedPoint } from "@balancer-labs/v3-solidity-utils/contracts/math/FixedPoint.sol";
+import { ScalingHelpers } from "@balancer-labs/v3-solidity-utils/contracts/helpers/ScalingHelpers.sol";
+
+import { RateProviderMock } from "./RateProviderMock.sol";
+import { BaseHooks } from "../BaseHooks.sol";
+
+contract EmptyPoolHooksMock is BaseHooks {
```
jubeira, line 21:
```suggestion
/// @dev Simple hook contract that does nothing but return true on every call.
contract EmptyPoolHooksMock is BaseHooks {
```

pvt/benchmarks/src/PoolBenchmark.behavior.ts
```diff
@@ -180,6 +183,65 @@ export class Benchmark {
...
+        await saveSnap(
+          this._testDirname,
+          `[${this._poolType} - ${gasTag} - BatchRouter] swap single token exact in with fees - cold slots`,
+          receipt
+        );
+
+        await actionAfterFirstTx();
+
+        // Measure
+        tx = await batchRouter.connect(alice).swapExactOut(
```
jubeira, line 218: This one should also be exact in according to the gas file name below.

pvt/benchmarks/src/PoolBenchmark.behavior.ts
```diff
@@ -212,6 +274,9 @@ export class Benchmark {
       sharedBeforeEach('approve router to spend BPT', async () => {
         const bpt: IERC20 = await TypesConverter.toIERC20(this.pool);
         await bpt.connect(alice).approve(router, MAX_UINT256);
+        await bpt.connect(alice).approve(batchRouter, MAX_UINT256);
+        await bpt.connect(alice).approve(permit2, MAX_UINT256);
+        await permit2.connect(alice).approve(await bpt.getAddress(), batchRouter, MAX_UINT160, MAX_UINT48);
```
jubeira, line 279:
```suggestion
        await bpt.connect(alice).approve(permit2, MAX_UINT256);
        await permit2.connect(alice).approve(bpt, batchRouter, MAX_UINT160, MAX_UINT48);
```

`BatchRouter` always uses `permit2`, so you shouldn't need to give BPT allowance directly.

pvt/benchmarks/src/PoolBenchmark.behavior.ts
jubeira: Apparently this error was also there for the regular router; can we fix that one too?

jubeira (approved):
LGTM, nice one!

Just a few minor comments (some of them apply to existing issues, but if we can solve those here I think it'll be worth it).

pkg/vault/contracts/test/EmptyPoolHooksMock.sol
EndymionJkb: Also, as noted in #736: consider `EmptyHooksPoolMock` or `MinimalHooksPoolMock` for the name. It's the hooks contract that's "empty"; not the pool.

EndymionJkb (approved):
In general we try to avoid "stacking" PRs, but for those built on this one I think it would have helped see the real diff.

PR #734, opened 2024-07-04
Support permissionless pool creator fee withdrawal

EndymionJkb:
# Description

This was a partner request, to allow "public automation" of fee collection. If we wanted to do it, it's a minor change.

My reservation about this is it allows anyone to "force" collection of pool creator fees (though they would have to pay the gas for it, which wouldn't be completely trivial). What if there's a reason they need to be collected on a certain schedule (or at least not "unexpectedly")?

## Type of change

- [ ] Bug fix 
- [X] New feature 
- [ ] Breaking change 
- [ ] Dependency changes
- [ ] Code refactor / cleanup
- [ ] Other

## Checklist:


pkg/vault/test/foundry/ProtocolFeeController.t.sol
```diff
@@ -486,6 +486,10 @@ contract ProtocolFeeControllerTest is BaseVaultTest {
         feeController.withdrawPoolCreatorFees(pool, alice);
     }
 
+    function testPermissionlessWithdrawalByNonPoolCreator() public {
+        feeController.withdrawPoolCreatorFees(pool);
```
jubeira, line 490: Maybe check that they were effectively collected?

jubeira (commented):
Looks good!

@danielmkm WDYT? Should we include this feature?

pkg/vault/test/foundry/ProtocolFeeController.t.sol
EndymionJkb: I was afraid somebody might say that :) Both call a common routine, so it's pretty clear by inspection that it will work, but sure, I'll add the test. (After all, the code could change diverge later, and if does something should break.)

joaobrunoah (approved):
LGTM!

danielmkm (approved):
This looks good to me, makes sense to be able to collect permissionlessly for the sake of automation

PR #735, opened 2024-07-04
Always round balances down

### LINKED ISSUE #152: Review balances rounding direction
After #141 and #83 we have scaling factors moved to the vault, with rounding directions for amounts given, calculated, and balances.

For amounts given and calculated it's rather straightforward to reason about it.
For balances, at least using weighted pool math the rounding direction does not produce consistent results (both for swaps and for join-exit math). See e.g. https://github.com/balancer/balancer-v3-monorepo/actions/runs/6857314725/job/18646175815.

It might be the case that the rounding direction won't make any difference given the precision loss for the weighted math. Review together with #49.

EndymionJkb:
# Description

There are a couple places where we round balances up, and it's not clear this is the right thing to do. It also introduces a possible discrepancy between the public getter for live balances (which doesn't expose the rounding direction, and always rounds down), and what the live code would do. (Though we don't think this is consequential.)

## Type of change

- [ ] Bug fix 
- [ ] New feature 
- [ ] Breaking change 
- [ ] Dependency changes
- [X] Code refactor / cleanup
- [ ] Other

## Checklist:

## Issue Resolution

Resolves #152

### DIFF
--- a/pkg/vault/contracts/Vault.sol
+++ b/pkg/vault/contracts/Vault.sol
@@ -526,7 +526,7 @@ contract Vault is IVaultMain, VaultCommon, Proxy {
         //
         // Sets all fields in `poolData`. Side effects: updates `_poolTokenBalances`, and
         // `_aggregateFeeAmounts` in storage.
-        PoolData memory poolData = _loadPoolDataUpdatingBalancesAndYieldFees(params.pool, Rounding.ROUND_UP);
+        PoolData memory poolData = _loadPoolDataUpdatingBalancesAndYieldFees(params.pool, Rounding.ROUND_DOWN);
         InputHelpers.ensureInputLengthMatch(poolData.tokens.length, params.maxAmountsIn.length);
 
         // Amounts are entering pool math, so round down.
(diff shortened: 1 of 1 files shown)


pkg/vault/contracts/Vault.sol
```diff
@@ -526,7 +526,7 @@ contract Vault is IVaultMain, VaultCommon, Proxy {
         //
         // Sets all fields in `poolData`. Side effects: updates `_poolTokenBalances`, and
         // `_aggregateFeeAmounts` in storage.
-        PoolData memory poolData = _loadPoolDataUpdatingBalancesAndYieldFees(params.pool, Rounding.ROUND_UP);
+        PoolData memory poolData = _loadPoolDataUpdatingBalancesAndYieldFees(params.pool, Rounding.ROUND_DOWN);
```
jubeira, line 529: There is actually a justification for this in the comment above.
Strictly speaking, to the best of my understanding rounding up is the correct thing to do here. The question is whether this actually makes any difference or not when we apply minimum amounts.

PR #736, opened 2024-07-04
Clear _currentSwapTokensOut and _currentSwapTokensIn every each operations in BatchRouter

### LINKED ISSUE #707: Remove token information in batchRouter after operations
No we don't remove this information but we need to do it.

elshan-eth:
# Description

This PR fixed a problem with `_currentSwapTokensIn`/`_currentSwapTokensOut` token sets in `BatchRouter`.  We cleared only `_currentSwapTokenInAmounts`, `_currentSwapTokenOutAmounts`, and `_settledTokenAmounts` at the end of the operation but didn't do it for `_currentSwapTokensOut`/`_currentSwapTokensIn`. This problem happens every time we have multiple operations in one transaction.

In the first iteration, I tried to create an index for all mappings which we iterate each operation. It means that if the index changes, we don't need to clear every property. This way had a problem with `stack too deep` (without viaIR compilation) and gas was increased. 

This way is simple and cheap.

NOTE: merge this PR after #731

## Type of change

- [ ] Bug fix 
- [ ] New feature 
- [ ] Breaking change 
- [ ] Dependency changes
- [ ] Code refactor / cleanup
- [X] Other

## Checklist:

## Issue Resolution

Closes #707


pkg/vault/contracts/test/EmptyPoolHooksMock.sol
```diff
@@ -0,0 +1,121 @@
...
+import { RateProviderMock } from "./RateProviderMock.sol";
+import { BaseHooks } from "../BaseHooks.sol";
+
+contract EmptyPoolHooksMock is BaseHooks {
+    using FixedPoint for uint256;
+    using ScalingHelpers for uint256;
+
+    HookFlags private _hookFlags;
+
+    constructor(IVault vault) BaseHooks(vault) {}
```
EndymionJkb, line 27:
```suggestion
    constructor(IVault vault) BaseHooks(vault) {
        // solhint-disable-previous-line no-empty-blocks
    }
```

EndymionJkb (commented):
I think this is fine, but I'd like to see a title and description that says what this is doing (will be hard to find later, and especially now that it's public, we want to document what PRs are doing).

My understanding is that the original problem was only the transient token amounts, but not the transient token set, was getting cleared at the end of the operation, which would be a problem if you had multiple operations in the same transaction.

The first attempt was something involving storing the last index, which was complex, added a modifier, and cost more. This simply calls "remove" on the token set at the same time it sets the amount to 0.

In addition to that fix (which could have been factored out), this adds benchmark tests for the BatchRouter

pkg/vault/contracts/test/EmptyPoolHooksMock.sol
```diff
@@ -0,0 +1,121 @@
...
+import { IVaultMock } from "@balancer-labs/v3-interfaces/contracts/test/IVaultMock.sol";
+import "@balancer-labs/v3-interfaces/contracts/vault/VaultTypes.sol";
+
+import { FixedPoint } from "@balancer-labs/v3-solidity-utils/contracts/math/FixedPoint.sol";
+import { ScalingHelpers } from "@balancer-labs/v3-solidity-utils/contracts/helpers/ScalingHelpers.sol";
+
+import { RateProviderMock } from "./RateProviderMock.sol";
+import { BaseHooks } from "../BaseHooks.sol";
+
+contract EmptyPoolHooksMock is BaseHooks {
```
EndymionJkb, line 21: Probably would have been better to base this off #731, so the files aren't duplicated; would have reduced the diff.

Also, consider `EmptyHooksPoolMock` or `MinimalHooksPoolMock` for the name. It's the hooks contract that's "empty"; not the pool.

joaobrunoah (commented):
The code seems nice, but just a question: If amounts is cleared and the amount of token to transfer is 0, why do we need to clear tokensIn and tokensOut, if effectively that token won't be transferred? I mean, looks like we added at least 1k gas to do this. Isn't it better to check if amount is greater than 0 before transfering the asset from/to the user during settle?

elshan-eth:
> The code seems nice, but just a question: If amounts is cleared and the amount of token to transfer is 0, why do we need to clear tokensIn and tokensOut, if effectively that token won't be transferred? I mean, looks like we added at least 1k gas to do this. Isn't it better to check if amount is greater than 0 before transfering the asset from/to the user during settle?

if there will be several calls within 1 transaction (I don’t know if this makes sense), it would be the same. I mean we can remove elements and it will cost around 100 gas but also if we don't remove this the next call will read it to compare it with zero and also pay 100 gas more.

EndymionJkb (approved):
Looks like everything older was addressed - it's now a two-liner, except for all the gas files, etc. Please run `yarn test` from the root to run everything and resolve all the conflicts, but other than that looks good to go. Definitely way simpler than the first method.

joaobrunoah (approved):
Looks good!

jubeira (approved):
LGTM, thanks @elshan-eth !

PR #737, opened 2024-07-04
Fix ERC4626 buffer queries

joaobrunoah:
# Description

A query of a batch swap using boosted pools was returning a different result from the actual transaction because the preview methods for `_wrapWithBuffer` function were not implemented correctly.

Query transaction: [Tenderly](https://www.tdly.co/shared/simulation/4446ba22-9d59-41ac-aaca-ad33c7fdf773)
Actual transaction: [Tenderly](https://www.tdly.co/shared/simulation/9c9818f9-b287-43fb-97ac-b2f0408ab1e7)

This PR fixes that and modifies a test in hardhat to ensure that this error is caught again if it happens.

## Type of change

- [x] Bug fix 
- [ ] New feature 
- [ ] Breaking change 
- [ ] Dependency changes
- [ ] Code refactor / cleanup
- [ ] Other

## Checklist:


pkg/solidity-utils/contracts/test/ERC4626TestToken.sol
```diff
@@ -36,7 +36,7 @@ contract ERC4626TestToken is ERC4626, IRateProvider {
     }
 
     function getRate() external view returns (uint256) {
-        return _convertToAssets(FixedPoint.ONE, Math.Rounding.Floor);
+        return _convertToAssets(FixedPoint.ONE, Math.Rounding.Ceil);
```
EndymionJkb, line 39: Maybe add a comment on the rounding direction here?

pvt/helpers/src/test/relativeError.ts
```diff
@@ -3,42 +3,49 @@ import { Decimal } from 'decimal.js';
 import { BigNumberish } from 'ethers';
 import { bn, pct } from '../numbers';
 
-export function expectEqualWithError(actual: BigNumberish, expected: BigNumberish, error: BigNumberish = 0.001): void {
+export function expectEqualWithError(
```
EndymionJkb, line 6: Good generalization (closer to parity with forge); we should start using this now.

EndymionJkb (approved):
Looks good (symmetrical). Uses donation to manipulate the rate :) It adds a little bytecode of course, but not much, and we were already over.

jubeira (approved):
Great catch, thanks @joaobrunoah !

Good thing that we had already moved this piece of code to the extension, where we have some more room.

Merging now to deploy to testnet.

PR #745, opened 2024-07-05
Add a new change type for optimization

EndymionJkb:
# Description

I think it would be useful to be able to call out PRs that are optimizations; I've noticed we've just been marking them "Other" (or "refactor").

The difference is that a "primary" refactor, while it should ideally also result in less gas or bytecode, is motivated by a desire to reduce complexity, improve readability (including renaming), or just substitute a better implementation (e.g., simplify boolean logic, break out common code into a separate function, etc.) There are also "cleanup" refactors, which would be stuff like removing unused variables, functions, or imports.

Optimizations are logically equivalent changes, but motivated specifically by trying to save gas or bytes. A straight refactor should always decrease complexity/increase readability, while an optimization is allowed to *slightly* increase complexity/decrease readability, if the benefit is great enough. They should also be more extensively documented.

## Type of change

- [ ] Bug fix 
- [ ] New feature 
- [ ] Breaking change 
- [ ] Dependency changes
- [ ] Code refactor / cleanup
- [X] Other

## Checklist:

### DIFF
--- a/.github/pull_request_template.md
+++ b/.github/pull_request_template.md
@@ -10,7 +10,7 @@
 - [ ] Breaking change <!-- (would cause existing functionality to not work as expected) -->
 - [ ] Dependency changes
 - [ ] Code refactor / cleanup
-- [ ] Optimization ([ ] gas / [ ] bytecode)
+- [ ] Optimization: [ ] gas / [ ] bytecode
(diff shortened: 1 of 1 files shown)


jubeira (approved):
Sounds good!

PR #746, opened 2024-07-05
Shave bytecode off the Vault

jubeira:
# Description

Just targeting simple low-hanging fruits that **shouldn't change the logic at all**:
- Use `unchecked` math whenever it's obvious it won't overflow (i.e. right next to `if` / `else` clauses)
- Move external getters to extension
- Eliminate unnecessary operations by visual inspection

This is enough to make the Vault deployable again.

## Type of change

- [ ] Bug fix 
- [ ] New feature 
- [ ] Breaking change 
- [ ] Dependency changes
- [ ] Code refactor / cleanup
- [x] Optimization: [ ] gas / [x] bytecode
- [ ] Other

## Checklist:

## Issue Resolution

N/A


pkg/vault/contracts/VaultCommon.sol
```diff
@@ -332,10 +332,7 @@ abstract contract VaultCommon is IVaultEvents, IVaultErrors, VaultStorage, Reent
         }
 
         // Still has to be a valid percentage, regardless of what the pool defines.
-        if (
-            swapFeePercentage > ISwapFeePercentageBounds(pool).getMaximumSwapFeePercentage() ||
-            swapFeePercentage > FixedPoint.ONE
```
EndymionJkb, line 337: This is in case the pool returns > 1. Even if not malicious, it could have some kind of calculation (maybe with an external call somewhere) with a bug in it that causes it to return invalid values if we just blindly take whatever the pool returns.

Ahhh - but this is checked later in `PoolConfigLib.setStaticSwapFeePercentage`, so the check here is now redundant. (I don't think it was when this was written.) `InvalidPercentage` is also a more specific error message (though the "too high" one isn't wrong).

I suggest adding a comment on that library call saying it validates the percentage.

EndymionJkb (approved):
LGTM (after CI passes)! I think we were deliberately not thinking about optimizations yet while writing that buffer code.

The failing test is probably due to the new error message.

PR #748, opened 2024-07-06
Guardrails (Step 1: minimum trade amount)

EndymionJkb:
# Description

This is an aggressive form of enforcing a minimum trade amount in the critical path, at the Vault level. Per discussion, it's a fixed value applied to scaled amounts. It checks amountGiven/amountCalculated on swaps, and also non-zero amountsIn/amountsOut on liquidity operations (allowing zeros to support general unbalanced operations).

We had originally wanted to do this at the pool level (like the swap fee limits). The issue there is liquidity operations (and theoretically even the swaps) don't *have* to go through the pool, so there would be no way to enforce the limits across all pools. (So, for instance, an attacker could just make a pool that didn't have any limits, and potentially use it to exploit core pools.)

In the present implementation, is generally allowing zeros on liquidity operations too permissive? The alternative would be to treat the different kinds differently, which sounds like a lot of bytecode.

One consequence of this approach is we can no longer have a 100% fee (see changes to the hook tests). That would be a UX issue, as the devs would need to know that they can't go to 100% (and the exact upper limit would depend on the pool state). A solution to that would be to do swaps the same as liquidity operations: allow 0, and only enforce the minimum for 1 wei and above. This would "fix" all the hook tests and reduce the diff.

So the main question is whether we can provide sufficient protection against small "attack" trades, without compromising UX. A check that has holes in it isn't a good check. It's actually counterproductive: adding complexity and bytecode without providing the full benefit.

Another approach we'd considered earlier was a "fudge factor" - not trying to limit the trade size, but just adding 1, 10, or 100 wei to all amountsIn, and deducting the same from amountsOut. I'm a bit worried that could actually be exploited itself (e.g., an attacker trying to get 0 out for something in could craft a trade that results in 1 out, knowing that we will subtract one).

Should we have this at all, or a fudge factor instead, or both? The central idea here is simple: we are saying that there is no legitimate reason to trade tiny wei amounts - those are either fat fingered inputs or attacks - and blocking these trades should help prevent many classes of attacks (e.g., the V2 LinearPool exploit). The research question is whether we can do this effectively, fit within the bytecode limit, and not create more problems than we solve.

## Type of change

- [ ] Bug fix 
- [X] New feature 
- [ ] Breaking change 
- [ ] Dependency changes
- [ ] Code refactor / cleanup
- [ ] Other

## Checklist:


pkg/vault/contracts/Vault.sol
```diff
@@ -664,14 +668,24 @@ contract Vault is IVaultMain, VaultCommon, Proxy {
...
-            // amountsInRaw are amounts actually entering the Pool, so we round up.
-            // Do not mutate in place yet, as we need them scaled for the `onAfterAddLiquidity` hook
-            uint256 amountInRaw = amountsInScaled18[i].toRawUndoRateRoundUp(
-                poolData.decimalScalingFactors[i],
-                poolData.tokenRates[i]
-            );
-            amountsInRaw[i] = amountInRaw;
+            {
+                uint256 amountInScaled18 = amountsInScaled18[i];
+                if (amountInScaled18 > 0) {
```
joaobrunoah, line 676: Why do we consider 0 to be a valid trade amount? Removing this `if` would save some bytecode (maybe) and reduce gas for the majority of operations, or not?

joaobrunoah (commented):
Interesting to check this in scaled18 values. When I heard about the problem I thought of doing it the other way around (scale the minimum amount to raw, and not the opposite).

This solution does not prevent the trade of 1 wei USDC, which is a 6 decimals (1 wei USDC, when converted to scaled18, is 1e12, which is greater than 1e6) so this solution seems effective only for tokens with more than 12 decimals, but not sure if this is an issue.

pkg/vault/contracts/Vault.sol
danielmkm: This check is in `_addLiquidity`. We need to support 0 amounts because unbalanced joins will often times contain 0 values.

Example: I join the USDC/DAI/USDT pool UNBALANCED with only USDC & DAI

danielmkm:
> Interesting to check this in scaled18 values. When I heard about the problem I thought of doing it the other way around (scale the minimum amount to raw, and not the opposite).
> 
> This solution does not prevent the trade of 1 wei USDC, which is a 6 decimals (1 wei USDC, when converted to scaled18, is 1e12, which is greater than 1e6) so this solution seems effective only for tokens with more than 12 decimals, but not sure if this is an issue.

Trading 1 wei USDC is not inherently dangerous, all pool math is performed on scaled 18 values, so rounding errors will always be limited to the last digits in scaled 18. The goal of this check is to provide another layer of security that helps to ensure that rounding errors in the last digits do not impact the amount calculated in ways that could potentially have down stream impact (ie: pumping the rate). A token with 6 decimals is already "safe" because any sort of rounding errors will be wiped out when we down scale back to raw amounts

pkg/vault/contracts/Vault.sol
```diff
@@ -885,13 +899,23 @@ contract Vault is IVaultMain, VaultCommon, Proxy {
...
+                    _ensureValidTradeAmount(amountOutScaled18);
+
+                    // amountsOut are amounts exiting the Pool, so we round down.
+                    // Do not mutate in place yet, as we need them scaled for the `onAfterRemoveLiquidity` hook
+                    amountOutRaw = amountsOutScaled18[i].toRawUndoRateRoundDown(
+                        poolData.decimalScalingFactors[i],
+                        poolData.tokenRates[i]
+                    );
+                }
+            }
```
danielmkm, line 917: Do we need to also check `bptAmountIn`?

pkg/vault/contracts/Vault.sol
```diff
@@ -664,14 +668,24 @@ contract Vault is IVaultMain, VaultCommon, Proxy {
...
+                    // amountsInRaw are amounts actually entering the Pool, so we round up.
+                    // Do not mutate in place yet, as we need them scaled for the `onAfterAddLiquidity` hook
+                    amountInRaw = amountInScaled18.toRawUndoRateRoundUp(
+                        poolData.decimalScalingFactors[i],
+                        poolData.tokenRates[i]
+                    );
+                }
+
+                amountsInRaw[i] = amountInRaw;
+            }
```
danielmkm, line 688: Do we need to check `bptAmountOut`?

danielmkm (commented):
The biggest flag for me here is this point:
> One consequence of this approach is we can no longer have a 100% fee (see changes to the hook tests). That would be a UX issue, as the devs would need to know that they can't go to 100% (and the exact upper limit would depend on the pool state).

Think this would create a big headache for DX since knowing the upper limit of the fee is likely pretty challenging.

> A check that has holes in it isn't a good check. It's actually counterproductive: adding complexity and bytecode without providing the full benefit.

100% agree here, if this is in the vault, we need to make it complete, it'll otherwise just create a false sense of security.

pkg/vault/contracts/Vault.sol
EndymionJkb: I think so, yes. Will make sure, and add it.

pkg/vault/contracts/Vault.sol
EndymionJkb: Same as above

EndymionJkb:
> The biggest flag for me here is this point:
> 
> > One consequence of this approach is we can no longer have a 100% fee (see changes to the hook tests). That would be a UX issue, as the devs would need to know that they can't go to 100% (and the exact upper limit would depend on the pool state).
> 
> Think this would create a big headache for DX since knowing the upper limit of the fee is likely pretty challenging.

What do you think of only checking non-zero amounts on swaps then? Is that a hole? I'll do a PR on top of this one to see how that looks.

pkg/vault/test/foundry/DynamicFeePoolTest.t.sol
```diff
@@ -197,7 +214,8 @@ contract DynamicFeePoolTest is BaseVaultTest {
     }
 
     function testSwapChargesFees__Fuzz(uint256 dynamicSwapFeePercentage) public {
-        dynamicSwapFeePercentage = bound(dynamicSwapFeePercentage, 0, 1e18);
+        // Cannot be 100%, or it would do a zero amount swap, which is below the minimum.
+        dynamicSwapFeePercentage = bound(dynamicSwapFeePercentage, 0, 0.9999e18);
```
elshan-eth, line 218: I think we can put 0.9999e18 to a constant

elshan-eth (commented):
Cool!

> One consequence of this approach is we can no longer have a 100% fee (see changes to the hook tests). That would be a UX issue, as the devs would need to know that they can't go to 100% (and the exact upper limit would depend on the pool state). A solution to that would be to do swaps the same as liquidity operations: allow 0, and only enforce the minimum for 1 wei and above. This would "fix" all the hook tests and reduce the diff.

Can we check the minimum amount before debiting the commission?

pkg/vault/test/foundry/DynamicFeePoolTest.t.sol
EndymionJkb: Well, we're trying to avoid having to do this (per @danielmkm's comment), by allowing 0 swaps (as we do for liquidity operations). This permits this to be 1 again.

However, it's arguable whether this is an improvement, as there is still a "banned" range near 100%. It's just that exactly 100% is now allowed. The UX issue with not allowing 100 is you'd have to explain that, and calculate the top value (which varies by pool state).

However, if you now increased the swap fee towards 100%, it would fail at some point (for a range < 100 but > 99.9x%), then work again at 100%. We'd also have to document this.

So the question is how is this used? Are they always just going to set 100% for this LVR case, whereas other use cases will set much lower values? Then it might be ok. But if we have a use case where it gradually increases the fee until it reaches 100%, this is arguably worse than not allowing 0.

(Also, is allowing 0 swaps a security hole making this entire thing pointless?)

elshan-eth (approved):

pkg/vault/contracts/Vault.sol
```diff
@@ -661,17 +665,29 @@ contract Vault is IVaultMain, VaultCommon, Proxy {
...
-            // amountsInRaw are amounts actually entering the Pool, so we round up.
-            // Do not mutate in place yet, as we need them scaled for the `onAfterAddLiquidity` hook
-            uint256 amountInRaw = amountsInScaled18[i].toRawUndoRateRoundUp(
-                poolData.decimalScalingFactors[i],
-                poolData.tokenRates[i]
-            );
-            amountsInRaw[i] = amountInRaw;
+            {
+                uint256 amountInScaled18 = amountsInScaled18[i];
+                if (amountInScaled18 > 0) {
```
jubeira, line 678: Interesting.

We're already checking for this in `_ensureValidTradeAmount`, so this basically is an optimization for the math in `toRawUndoRateRoundUp`. What if we do it inside the library instead? The scope and the if inside a for makes the code a bit heavy to ready IMO. WDYT?

pkg/vault/contracts/Vault.sol
```diff
@@ -886,16 +902,28 @@ contract Vault is IVaultMain, VaultCommon, Proxy {
...
             // 1) Calculate raw amount out.
-            // amountsOut are amounts exiting the Pool, so we round down.
-            // Do not mutate in place yet, as we need them scaled for the `onAfterRemoveLiquidity` hook
-            uint256 amountOutRaw = amountsOutScaled18[i].toRawUndoRateRoundDown(
-                poolData.decimalScalingFactors[i],
-                poolData.tokenRates[i]
-            );
+            {
+                uint256 amountOutScaled18 = amountsOutScaled18[i];
+                if (amountOutScaled18 > 0) {
```
jubeira, line 915: (same as above)

pkg/vault/contracts/Vault.sol
```diff
@@ -1522,6 +1550,14 @@ contract Vault is IVaultMain, VaultCommon, Proxy {
         return amount + _MAX_CONVERT_ERROR;
     }
 
+    // Minimum swap amount (applied to scaled18 values), enforced as a security measure to block potential
+    // exploitation of rounding errors
+    function _ensureValidTradeAmount(uint256 tradeAmount) private pure {
+        if (tradeAmount < _MINIMUM_TRADE_AMOUNT && tradeAmount != 0) {
```
jubeira, line 1556: How about checking `!= 0` first? super minor, but that's going to be the case pretty often in unbalanced swaps.

jubeira (commented):
Great work @EndymionJkb ! Looks pretty reasonable.

I have a few suggestions (mostly minor, but maybe it improves the readability a bit?)

On the other hand: can we also change the minimum wrap amount to 1e3 as part of this PR? We're applying it to raw amounts, so 1e6 is way too high in this case for tokens that don't have 18 decimals.

pkg/vault/contracts/Vault.sol
EndymionJkb: I had it like that and switched it to this, thinking most trades would be swaps (and thus non-zero). How many straight swaps do we have vs. unbalanced add/remove liquidity operations?

pkg/vault/contracts/Vault.sol
EndymionJkb: Trying it. Have ScalingHelpers check for 0 for functions that do multiple operations (e.g., more than one multiply/divide). Not sure if the compiler already optimizes this?

pkg/vault/contracts/Vault.sol
jubeira: The difference is that in a successful swap you will still need to check both conditions, whereas in a successful `add` / `remove` you won't.

We don't care optimizing gas for cases that will revert anyways.

pkg/solidity-utils/contracts/helpers/ScalingHelpers.sol
```diff
@@ -82,7 +82,7 @@ library ScalingHelpers {
         uint256 tokenRate
     ) internal pure returns (uint256) {
         // Do division last, and round scalingFactor * tokenRate up to divide by a larger number.
-        return FixedPoint.divDown(amount, scalingFactor.mulUp(tokenRate));
+        return amount == 0 ? 0 : FixedPoint.divDown(amount, scalingFactor.mulUp(tokenRate));
```
jubeira, line 85: Just a triple check for `undo` functions:
- scaling factors depend on token decimals, and can never be 0.
- if token rate is 0, the previous code would revert with panic, but this one would not.

If the rate is 0, then the scaled balances would be 0 anyways. But in that case scaling back doesn't make sense; it's probably better to revert.

While tokens with rate 0 don't make sense per-se, we can't discard some weird intermediate state inside a tx making the rate 0 for tokens. That case should probably revert; continuing the execution might bring unintended consequences.

I'd remove this, and keep the code in `_addLiquidity` and `_removeLiquidity` as it is now without the extra `if`. If the amounts are 0, then let's just let the usual math do the job of multiplying 0 by the rate.

jubeira (commented):
Looks good! After some more thought, I think I'd remove that extra optimization in the libraries. The rest looks fine; I wouldn't add a special case when the amount is 0 in `add` and `remove`.

PR #751, opened 2024-07-09
Guardrails (Step 1a: minimum trade amount - allow 0)

EndymionJkb:
# Description

Variant of minimum Guardrails built on #748 , which allows 0-value swaps. This removes the special case we'd otherwise need for donations. It also allows 100% fees (but not *near 100%* fees that would result in small trades. So I'm not sure this is really better. It allows exactly 100%, but there is a "gap" of disallowed values that you'd have to calculate (see the corresponding test: `FixedPoint.ONE - MIN_TRADE_AMOUNT.divDown(defaultAmount)`).

## Type of change

- [ ] Bug fix 
- [ ] New feature 
- [ ] Breaking change 
- [ ] Dependency changes
- [X] Code refactor / cleanup
- [ ] Other

## Checklist:


pkg/vault/contracts/Vault.sol
```diff
@@ -1536,7 +1533,7 @@ contract Vault is IVaultMain, VaultCommon, Proxy {
     }
 
     function _ensureValidTradeAmount(uint256 tradeAmount) private pure {
-        if (tradeAmount < _MINIMUM_TRADE_AMOUNT) {
+        if (tradeAmount != 0 && tradeAmount < _MINIMUM_TRADE_AMOUNT) {
```
joaobrunoah, line 1536: It's much better than checking kind = `Donation`.

pkg/vault/contracts/Vault.sol
```diff
@@ -1536,7 +1533,7 @@ contract Vault is IVaultMain, VaultCommon, Proxy {
     }
 
     function _ensureValidTradeAmount(uint256 tradeAmount) private pure {
```
joaobrunoah, line 1535: This is probably not related to this PR, but maybe the comment over the _MINIMUM_TRADE_AMOUNT variable should be on top of this function as well, for clarity.

joaobrunoah (approved):
LGTM!

PR #760, opened 2024-07-10
Vault explorer

### LINKED ISSUE #750: Etherscan visibility
Bytecode constraints drove us to the Proxy pattern for the Vault: splitting it into three contracts that delegate call each other (thanks again, Igor!). A consequence of this is that looking at the [Vault on Etherscan](https://sepolia.etherscan.io/address/0xD5584b37D1845fFeD958C2d94bC675603DdCce68#readContract) doesn't show much, and trying to call functions on the extensions in Etherscan from an EOA will revert.

As Juani points out, "...you can build the tx in etherscan and then in the wallet target the vault". Another idea is to make a wrapper contract that presents a reasonable interface to users. Both involve indirection/non-ideal dev UX, but "unusable Etherscan" isn't great either.

EndymionJkb:
# Description

It was noted (see issue #750) that because of our use of the Proxy pattern for the Vault (necessitated by a bytecode far over the limit), Etherscan (and off-chain tools generally) would not be very useful, as nearly all the "interesting" functions on the Vault are either only accessible through unlocking, or are exposed on the Extension or Admin contracts, which must be called through the Vault.

This PR introduces the `VaultExplorer` (naming suggestions welcome; I didn't want to use "API", as that would be confusable with the off-chain API), which is a wrapper around all externally callable (and non-permissioned) "greater Vault" functions. These are mostly getters and read-only functions, but you can also call `registerPool` and `collectAggregateFees`.

We would deploy this after the Vault, and people who wanted to see the whole interface on-chain and be able to interact through Etherscan as they do today with V2, would have an easy way to do it.

This was a long-time coming, mainly because the tests were so tedious! They might be a bit overkill, but I did find things while doing them, so I think it was worth it.

## Type of change

- [ ] Bug fix 
- [X] New feature 
- [ ] Breaking change 
- [ ] Dependency changes
- [ ] Code refactor / cleanup
- [ ] Other

## Checklist:

## Issue Resolution

Closes #750


joaobrunoah (approved):
This is awesome! Just a small comment about a non-view function that I'd remove from VaultExplorer contract.

pkg/vault/contracts/VaultExplorer.sol
```diff
@@ -0,0 +1,333 @@
...
+    }
+
+    /*******************************************************************************
+                                        Fees
+    *******************************************************************************/
+
+    /// @inheritdoc IVaultExplorer
+    function collectAggregateFees(address pool) external {
+        return _vault.collectAggregateFees(pool);
+    }
```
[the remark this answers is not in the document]
EndymionJkb: You can also `registerPool`. That's why I was questioning the name. Probably people wouldn't be registering pools outside factories, and presumably collecting fees would be done programmatically somehow as well. "Explorer" does imply read-only. It could be "Facade" or "Interface" or "API" - but API is confusable, and interface is also confusable with the technical meaning.

I can remove those for now; we can always add them in later, since the tests, etc. are in source control.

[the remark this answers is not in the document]
EndymionJkb: Per discussion, we'll take out registerPool, which doesn't make a lot of sense (should be done through a factory - how did they deploy the pool they're registering?), but leave the fee collection, which doesn't hurt anything.

pkg/interfaces/contracts/vault/IVaultExplorer.sol
```diff
@@ -0,0 +1,403 @@
...
+    function getPoolRoleAccounts(address pool) external view returns (PoolRoleAccounts memory);
+
+    /**
+     * @notice Query the current dynamic swap fee of a pool, given a set of swap parameters.
+     * @param pool The pool
+     * @param swapParams The swap parameters used to compute the fee
+     * @return success True if the pool has a dynamic swap fee and it can be successfully computed
+     * @return dynamicSwapFee The dynamic swap fee percentage
+     */
+    function computeDynamicSwapFee(
```
jubeira, line 279: should this be dynamic fee percentage now?

pkg/vault/test/foundry/VaultExplorer.t.sol
```diff
@@ -0,0 +1,750 @@
...
+        assertEq(settlementAmount, defaultAmount, "Wrong settlement amount");
+        assertEq(vaultDelta, -settlementAmount.toInt256(), "Wrong Vault token delta");
+        assertEq(explorer.getTokenDelta(dai), vaultDelta, "getTokenDelta mismatch");
+    }
+
+    function testGetReservesOf() public {
+        dai.mint(address(vault), defaultAmount);
+
+        assertEq(vault.getReservesOf(dai), defaultAmount, "Wrong Vault reserves");
+        assertEq(explorer.getReservesOf(dai), defaultAmount, "Wrong Explorer reserves");
```
jubeira, line 160: reserves are actually 0 here :smile: 

you'd need to unlock and settle for the vault to acknowledge them

pkg/vault/test/foundry/VaultExplorer.t.sol
```diff
@@ -0,0 +1,750 @@
...
+                address(tokens[i]),
+                string.concat("Address of token ", Strings.toString(i), " does not match.")
+            );
+
+            assertEq(
+                uint8(poolData.tokenInfo[i].tokenType),
+                uint8(TokenType.WITH_RATE),
+                string.concat("Token type of token ", Strings.toString(i), " does not match")
+            );
+        }
```
jubeira, line 294: I think you could also get pool data from vault, then from explorer, and compare the hashes right?

pkg/vault/test/foundry/VaultExplorer.t.sol
```diff
@@ -0,0 +1,750 @@
...
+        PoolData memory poolData = vault.getPoolData(pool);
+
+        uint256 daiLiveBalance = poolData.balancesRaw[daiIdx].toScaled18ApplyRateRoundDown(
+            poolData.decimalScalingFactors[daiIdx],
+            poolData.tokenRates[daiIdx]
+        );
+        uint256 usdcLiveBalance = poolData.balancesRaw[usdcIdx].toScaled18ApplyRateRoundDown(
+            poolData.decimalScalingFactors[usdcIdx],
+            poolData.tokenRates[usdcIdx]
+        );
```
jubeira, line 349: Why not just get current live balances from the vault?

jubeira (approved):
Thanks @EndymionJkb, looks good and makes sense!

Just a few very minor comments regarding the tests, otherwise LGTM!

pkg/vault/test/foundry/VaultExplorer.t.sol
EndymionJkb: I guess, but this would tell you exactly what was wrong, and it's kind of self-documenting.

PR #762, opened 2024-07-10
Fix factories to expose `disableUnbalancedLiquidity` flag

### LINKED ISSUE #761: Expose LiquidityManagement on factories
The creation of Weighted Pools with the Fee Taking Hook example is failing because the user can't disable unbalanced joins/exits with the current factory:

https://dashboard.tenderly.co/explorer/fork/f94905ca-300c-4b54-82cc-e36b24e25b46/simulation/2085af54-6f11-4415-abe2-c6317e20f17c

We should expose the LiquidityManagement object in the factory's create function, so users have more flexibility to create the pool.

joaobrunoah:
# Description

When testing hooks with Weighted pools, @mendesfabio reported that pool creation was reverting. This happened because `disableUnbalancedLiquidity` flag was false, and then the hook is not able to enable `hookAdjustedAmounts`.

https://dashboard.tenderly.co/explorer/fork/f94905ca-300c-4b54-82cc-e36b24e25b46/simulation/2085af54-6f11-4415-abe2-c6317e20f17c?trace=0.538.0.25

This PR fixes weighted and stable pool factories.

## Type of change

- [x] Bug fix 
- [ ] New feature 
- [ ] Breaking change 
- [ ] Dependency changes
- [ ] Code refactor / cleanup
- [ ] Other

## Checklist:

## Issue Resolution

Closes #761


pkg/pool-hooks/test/foundry/DirectionalFeeHookExample.t.sol
```diff
@@ -75,6 +75,7 @@ contract DirectionalHookExampleTest is BaseVaultTest {
                 MIN_SWAP_FEE,
                 poolHooksContract,
                 false, // Does not allow donations
+                false, // keep support to unbalanced add/remove liquidity
```
EndymionJkb, line 78: I think this is a bit more clear/consistent.
```suggestion
                false, // Do not enable donations
                false, // Do not disable unbalanced add/remove liquidity
```

pkg/pool-stable/test/StablePool.test.ts
```diff
@@ -80,7 +80,8 @@ describe('StablePool', () => {
       { pauseManager: ZERO_ADDRESS, swapFeeManager: ZERO_ADDRESS, poolCreator: ZERO_ADDRESS },
       MIN_SWAP_FEE,
       ZERO_ADDRESS,
-      false,
+      false, // no donations
+      false, // keep support to unbalanced add/remove liquidity
```
EndymionJkb, line 84: Same as above.
```suggestion
      false, // Do not enable donations
      false, // Do not disable unbalanced add/remove liquidity
```

EndymionJkb: Here and below, where the same comments appear

EndymionJkb (commented):
Looks good, but ideally should have a test where you make a weighted or stable pool with such a hook, and show that it works.

We're on the edge of just passing in a LiquidityManagement struct, as we're manually setting two of the four fields. The other two are custom add/remove, which by definition would require a new pool type, so they would always be false. So I think it's ok as is.

joaobrunoah:
@EndymionJkb I just created the tests, and I agree with you about the LiquidityManagement. I didn't expose the whole struct because custom add and remove require a new pool type.

EndymionJkb (approved):
LGTM! Remember to deploy as version 2

PR #774, opened 2024-07-12
Alternate test for BaseHooks

EndymionJkb:
# Description

Based on #772, which ensures modifiers on hook examples, this provides a way to meaningfully test BaseHooks.

This PR:

1) Moves HookFlags from inside the IHooks interface to VaultTypes. I can see both sides here, but I don't think it's best practice generally to define structs inside interfaces (especially if widely used), and putting it next to HooksConfig makes it really clear that the HooksConfig = HookFlags + hook contract address, and we can visually inspect to ensure they don't start diverging.

2) Leaves BaseHooks abstract, but changes the visibility of the hook functions from external to public, so that derived contracts can call them.

3) Creates a BaseHooksMock concrete contract that overrides all the hook functions and calls the base class. The corresponding test ensures that BaseHooks functions all return false. We're checking here that if a misconfigured hook contract doesn't define a given hook, and the Vault calls it, it will fall back on BaseHooks and return false/revert.

It doesn't really make sense to test `onlyVault` constraints here, as real hook contracts will not be calling the base contract functions, so the modifiers would never be executed. `onlyVault` constraints have to be tested at the most derived level (e.g., in the tests for the example hook contracts).

## Type of change

- [ ] Bug fix 
- [ ] New feature 
- [ ] Breaking change 
- [ ] Dependency changes
- [ ] Code refactor / cleanup
- [ ] Other

## Checklist:


pkg/vault/contracts/BaseHooks.sol
```diff
@@ -84,7 +85,7 @@ abstract contract BaseHooks is IHooks, VaultGuard {
         uint256[] memory,
         uint256[] memory,
         bytes memory
-    ) external virtual onlyVault returns (bool) {
+    ) public virtual onlyVault returns (bool) {
```
joaobrunoah, line 88:
```suggestion
    ) public virtual returns (bool) {
```
Should we remove onlyVault from BaseHooks, since it's an abstract contract?

pkg/vault/contracts/BaseHooks.sol
```diff
@@ -98,18 +99,18 @@ abstract contract BaseHooks is IHooks, VaultGuard {
         uint256[] memory amountsOutRaw,
         uint256[] memory,
         bytes memory
-    ) external virtual onlyVault returns (bool, uint256[] memory) {
+    ) public virtual onlyVault returns (bool, uint256[] memory) {
```
joaobrunoah, line 102:
```suggestion
    ) public virtual returns (bool, uint256[] memory) {
```

pkg/vault/contracts/BaseHooks.sol
```diff
@@ -8,6 +8,7 @@ import { IBasePool } from "@balancer-labs/v3-interfaces/contracts/vault/IBasePoo
 import {
     AddLiquidityKind,
     HooksConfig,
+    HookFlags,
     LiquidityManagement,
     RemoveLiquidityKind,
     TokenConfig
```
joaobrunoah, line 14: I thought about removing the VaultGuard from BaseHooks since it won't use the `onlyVault` modifier, but maybe it's better to keep it so external developers have a more standard way to use the vault in hooks.

joaobrunoah (approved):
LGTM! Only small comments.

pkg/vault/contracts/BaseHooks.sol
EndymionJkb: Yes; must have missed that one.

pkg/vault/contracts/BaseHooks.sol
EndymionJkb: Yes; I also think it's better to leave in at this layer, even though it's not directly used, as otherwise the derived contracts would have to use it. I can add a comment about it, though.

PR #778, opened 2024-07-15
Revert potentially dangerous optimization

EndymionJkb:
# Description

As Juani pointed out, caching the `bufferWrappedSurplus` might be dangerous, given that there are external calls between when it's initialized and the last time it's used.

I tried a few variants that preserved some of it (e.g., just reloading as necessary), but as the point of this was to save bytecode and get the Vault to fit, simply reverting that change turned out to be best. (The variants were actually slightly more bytecode.)

Luckily, due to the other changes, it still fits with this change reverted.

## Type of change

- [X?] Bug fix 
- [ ] New feature 
- [ ] Breaking change 
- [ ] Dependency changes
- [ ] Code refactor / cleanup
- [ ] Other

## Checklist:


pkg/vault/contracts/Vault.sol
```diff
@@ -1179,17 +1176,16 @@ contract Vault is IVaultMain, VaultCommon, Proxy {
                 amountInUnderlying = vaultUnderlyingDelta - bufferUnderlyingSurplus;
                 // Since bufferUnderlyingSurplus was wrapped, the final amountOut needs to discount the wrapped amount
                 // that will stay in the buffer
-                amountOutWrapped = vaultWrappedDelta - bufferWrappedSurplus;
+                amountOutWrapped = vaultWrappedDelta - wrappedToken.convertToShares(bufferUnderlyingSurplus);
```
jubeira, line 1179: Maybe make `bufferWrappedSurplus` local
```suggestion
                uint256 bufferWrappedSurplus = wrappedToken.convertToShares(bufferUnderlyingSurplus)
                amountOutWrapped = vaultWrappedDelta - bufferWrappedSurplus;
```

pkg/vault/contracts/Vault.sol
```diff
@@ -1179,17 +1176,16 @@ contract Vault is IVaultMain, VaultCommon, Proxy {
...
                 // In a wrap operation, the underlying balance of the buffer will decrease and the wrapped balance will
                 // increase. To decrease underlying balance, we get the delta amount that was deposited
                 // (vaultUnderlyingDelta) and discounts the amount needed in the wrapping operation
                 // (amountInUnderlying). Same logic applies to wrapped balances.
                 // Note: bufferUnderlyingSurplus = vaultUnderlyingDelta - amountInUnderlying
-                //       bufferWrappedSurplus = vaultWrappedDelta - amountOutWrapped
                 bufferBalances = PackedTokenBalance.toPackedBalance(
                     bufferBalances.getBalanceRaw() - bufferUnderlyingSurplus,
-                    bufferBalances.getBalanceDerived() + bufferWrappedSurplus
+                    bufferBalances.getBalanceDerived() + (vaultWrappedDelta - amountOutWrapped)
```
jubeira, line 1188: And reuse here to save an addition:
```suggestion
                    bufferBalances.getBalanceDerived() + bufferWrappedSurplus
```

jubeira (commented):
Thanks @EndymionJkb !

I have one suggestion that may save some bytecode following the spirit of the original change, but without caching the results.

Note that we also need to apply these changes to `unwrap`. I'll do one more pass after that's done.

pkg/vault/contracts/Vault.sol
EndymionJkb: I swear I tried this, and the extra assignment canceled out the benefit, so I just left it. Now it seems to save 3 bytes, though. Well, it's clearer anyway.

jubeira (commented):
Looks good now!

One final ask: can we make the code symmetrical? Either option works (make the surplus variable as local as possible as `wrap`, or declare above and reuse as with `unwrap`).

Keeping the exact same structure helps readers follow through better. Same that happens with add / remove.

Sorry for being so annoying :D

EndymionJkb:
> Looks good now!
> 
> One final ask: can we make the code symmetrical? Either option works (make the surplus variable as local as possible as `wrap`, or declare above and reuse as with `unwrap`).
> 
> Keeping the exact same structure helps readers follow through better. Same that happens with add / remove.
> 
> Sorry for being so annoying :D

NP; will do

PR #781, opened 2024-07-15
Rename dynamic fee hook

EndymionJkb:
# Description

While reviewing the docs (including code comments), I kept stumbling over this. `onComputeDynamicSwapFee` does not compute a fee (as similarly named functions do). It actually computes a fee percentage, and even though the name gets long, I think it improves readability to rename it.

I did not rename the `shouldCallComputeDynamicSwapFee` and associated flags and functions, because the diff would go nuts, and I think that's clear enough. Not really wrong - it is ultimately computing the value of the fee - it's just that the hook function named that only computes a percentage.

I didn't want to put this under a general "polishing" PR; renaming a hook is a lot more consequential.

Since it is only a rename, the bytecode and gas should not have changed; probably main was out-of-date.

## Type of change

- [ ] Bug fix 
- [ ] New feature 
- [ ] Breaking change 
- [ ] Dependency changes
- [X] Code refactor / cleanup
- [ ] Other

## Checklist:


joaobrunoah (approved):
LGTM!

jubeira (approved):
LGTM, good point!

PR #783, opened 2024-07-16
Fix TransientEnumerableSet's remove

joaobrunoah:
# Description

The remove function worked properly when a removed element was never inserted again in the same transaction. However, sometimes this is not true, like when we use batch router to wrap/unwrap a token and call batch router again to swap with a boosted pool in the same transaction. This PR fixes it. (tests in TransientEnumerableSet are required, there's a ticket for that, #784)

## Type of change

- [x] Bug fix 
- [ ] New feature 
- [ ] Breaking change 
- [ ] Dependency changes
- [ ] Code refactor / cleanup
- [ ] Other

## Checklist:


jubeira (approved):
LGTM, great catch @joaobrunoah !

PR #785, opened 2024-07-16
Erc4626 fuzz tests

joaobrunoah:
# Description

## Type of change

- [ ] Bug fix 
- [ ] New feature 
- [ ] Breaking change 
- [ ] Dependency changes
- [ ] Code refactor / cleanup
- [ ] Other

## Checklist:


pkg/vault/.forge-snapshots/forkBoostedPoolEmptyBufferSwapExactIn.snap
```diff
@@ -1 +1 @@
-723492
\ No newline at end of file
+680048
```
joaobrunoah, line 1: The new aTokens are cheaper

pkg/vault/contracts/BatchRouter.sol
```diff
@@ -157,7 +157,6 @@ contract BatchRouter is IBatchRouter, BatchRouterStorage, RouterCommon, Reentran
             if (path.steps[0].isBuffer && EVMCallModeHelpers.isStaticCall() == false) {
                 // If first step is a buffer, take the token in advance. We need this to wrap/unwrap.
                 _takeTokenIn(params.sender, stepTokenIn, stepExactAmountIn, false);
-                _settledTokenAmounts().tAdd(address(stepTokenIn), stepExactAmountIn);
```
joaobrunoah, line 160: This was leaving some dust if batch router is used again in the same transaction, and the settled amount is not used.

jubeira: This line should also go away if I'm not mistaken: https://github.com/balancer/balancer-v3-monorepo/pull/785/files#diff-3bcb4ada7438e6f8075537f3930a0099ae12d4389c22e6cb86c973f84681b0e7R456

Settled token amounts is only for calculated amounts, not for given inputs.

pkg/vault/contracts/VaultExtension.sol
jubeira, line 1: hmm ok, so the changes here basically mimic everything that happens on the other side of the Vault for the actual operation; is that correct?

This is not wrong, and we really don't care about bytecode size nor gas here as long as it fits within the limit. But even if the fuzz tests pass, we basically need to get the code right twice: for the query, and the actual operation.

Is there a way in which we can execute (almost) the exact same code in both cases? What if we save a "query context" flag in the Vault, and we just execute the same except when we need to do deposit / redeem? If the flag is on, we execute preview; if it's off we execute the actual operation. Do you see any reason why that approach would fail?

pkg/vault/test/foundry/fork/BoostedPoolQueryVsActualOperation.t.sol
```diff
@@ -0,0 +1,491 @@
...
+        amountIn = bound(amountIn, (USDC_BUFFER_INIT_AMOUNT) / 10, USDC_BUFFER_INIT_AMOUNT);
+        IBatchRouter.SwapPathExactAmountIn[] memory paths = _buildExactInPaths(usdcFork, amountIn, 0);
+
+        uint256 snapshotId = vm.snapshot();
+        _prankStaticCall();
+        (uint256[] memory queryPathAmountsOut, , ) = batchRouter.querySwapExactIn(paths, bytes(""));
+        vm.revertTo(snapshotId);
+
+        vm.prank(lp);
+        (uint256[] memory actualPathAmountsOut, , ) = batchRouter.swapExactIn(paths, MAX_UINT256, false, bytes(""));
```
jubeira, line 71: (here and elsewhere) Now that we're at it, maybe it's also worth comparing the other two return values?

jubeira: Maybe it's an overkill; I'd like to hear your opinion about it.

pkg/vault/test/foundry/fork/BoostedPoolQueryVsActualOperation.t.sol
```diff
@@ -0,0 +1,491 @@
...
+        IERC20 tokenIn,
+        uint256 exactAmountIn,
+        uint256 minAmountOut
+    ) private view returns (IBatchRouter.SwapPathExactAmountIn[] memory paths) {
+        paths = new IBatchRouter.SwapPathExactAmountIn[](1);
+        paths[0] = IBatchRouter.SwapPathExactAmountIn({
+            tokenIn: tokenIn,
+            steps: _getSwapSteps(tokenIn),
+            exactAmountIn: exactAmountIn,
+            minAmountOut: minAmountOut // rebalance tests are a wei off
```
jubeira, line 307: stale comment?

pkg/vault/test/foundry/fork/BoostedPoolQueryVsActualOperation.t.sol
```diff
@@ -0,0 +1,491 @@
...
+    }
+
+    function testDaiInBufferUnbalancedExactOut__Fork__Fuzz(
+        uint256 amountOut,
+        uint256 unbalancedDai,
+        uint256 unbalancedUsdc
+    ) public {
+        unbalancedDai = bound(unbalancedDai, 0, DAI_BUFFER_INIT_AMOUNT);
+        unbalancedUsdc = bound(unbalancedUsdc, 0, USDC_BUFFER_INIT_AMOUNT);
+        _unbalanceBuffers(unbalancedDai, unbalancedUsdc);
```
jubeira, line 277: Question: in these cases we're just using the DAI buffer; do we need USDC? Is it just to have a single `_unbalanceBuffers` function?

pkg/vault/test/foundry/fork/BoostedPoolQueryVsActualOperation.t.sol
```diff
@@ -0,0 +1,491 @@
...
+        // Identify and label underlying tokens.
+        usdcFork = IERC20(wUSDC.asset());
+        vm.label(address(usdcFork), "USDC");
+        daiFork = IERC20(wDAI.asset());
+        vm.label(address(daiFork), "DAI");
+
+        (wDaiIdx, wUsdcIdx) = getSortedIndexes(address(wDAI), address(wUSDC));
+    }
+
+    function _setupLP() private {
```
jubeira, line 415: holy moly

pkg/vault/test/foundry/fork/BoostedPoolQueryVsActualOperation.t.sol
jubeira: maybe it's fine and a refactor would be too much work for little gain, but maybe just fuzz a single amount and zero out the one we're not using?

pkg/vault/test/foundry/fork/BoostedPoolWithInitializedBuffer.t.sol
```diff
@@ -172,59 +171,61 @@ contract BoostedPoolWithInitializedBufferTest is BaseVaultTest {
...
-        vars.expectedBufferDeltaDai = swapAmount;
+        vars.expectedBufferDeltaDai = int256(swapAmount);
         vars.expectedDeltaUsdc = swapAmount / USDC_FACTOR;
-        vars.expectedBufferDeltaUsdc = swapAmount / USDC_FACTOR;
+        vars.expectedBufferDeltaUsdc = -int256(swapAmount / USDC_FACTOR);
 
         IBatchRouter.SwapPathExactAmountIn[] memory paths = _buildExactInPaths(
             swapAmount,
-            swapAmount / USDC_FACTOR - 1
+            swapAmount / USDC_FACTOR - 2
```
jubeira, line 197: Can we add a comment here?

pkg/vault/test/foundry/fork/BoostedPoolWithInitializedBuffer.t.sol
```diff
@@ -251,38 +258,134 @@ contract BoostedPoolWithInitializedBufferTest is BaseVaultTest {
...
             tooLargeSwapAmount + DAI_TO_USDC_FACTOR,
             tooLargeSwapAmount / USDC_FACTOR
         );
 
-        snapStart("forkBoostedPoolSwapTooLarge-ExactOut");
+        // EXACT_IN DAI -> USDC does not introduce rounding issues, since the resulting amountOut is divided by 1e12
+        // to get the correct amount of USDC to return.
+        // However, EXACT_OUT DAI -> USDC has rounding issues, because amount out is given in 6 digits, and we want an
+        // 18 decimals amount in, which is not precisely calculated. The calculation below reproduces what happens in
+        // the vault to scale tokens in the swap operation of the boosted pool
```
jubeira, line 289: :100:

jubeira (commented):
Impressive work @joaobrunoah, really nice stuff!

First pass done. Overall looks good, and the tests seem to be very comprehensive. I'm feeling much better about it now.

While I'm not against the current query implementation, and the tests are already providing a good degree of confidence in it, I'd still want to consider whether there's a viable implementation where we reuse the actual operation code as much as possible. That's the approach we do everywhere else (i.e. the code for every other query is pretty much the same as the actual operation, with `removeLiquidity` being the notable exception). Perhaps we can use the same approach in this case?

joaobrunoah:
> This line should also go away if I'm not mistaken: https://github.com/balancer/balancer-v3-monorepo/pull/785/files#diff-3bcb4ada7438e6f8075537f3930a0099ae12d4389c22e6cb86c973f84681b0e7R456
> 
> Settled token amounts is only for calculated amounts, not for given inputs.

@jubeira , about this, when I remove this line the test breaks. In this case, tokenIn is calculated and user paid maxAmountsIn in advance. We need the settledTokenAmounts for tokenIn to return the difference between maxAmountsIn and amountsIn at the end of the operation.

pkg/vault/contracts/BatchRouter.sol
joaobrunoah: @jubeira , about this, when I remove this line the test breaks. In this case, tokenIn is calculated and user paid maxAmountsIn in advance. We need the settledTokenAmounts for tokenIn to return the difference between maxAmountsIn and amountsIn at the end of the operation.

pkg/vault/contracts/VaultExtension.sol
joaobrunoah: New PR opened with an alternative: #791

pkg/vault/test/foundry/fork/BoostedPoolQueryVsActualOperation.t.sol
joaobrunoah: Yes, makes sense! Done!

pkg/vault/test/foundry/fork/BoostedPoolQueryVsActualOperation.t.sol
joaobrunoah: Nope, we're using both. The batch swap has 3 steps: Buffer -> Boosted Pool -> Buffer

In this case, is `Buffer DAI -> Boosted Pool -> Buffer USDC` ("DAI-first" scenario).

So, the balance of both buffers alters the operation.

I tested the `Buffer USDC -> Boosted Pool -> Buffer DAI` ("USDC-first" scenario) scenario too, because DAI has 18 decimals and USDC has 6, which can create rounding issues not covered by the "DAI-first" scenario.

pkg/vault/test/foundry/fork/BoostedPoolWithInitializedBuffer.t.sol
joaobrunoah: I believe this comment was in there already, because of another test, but I agree it's a good comment, hahahaha

EndymionJkb:
We're trying not to say "Boosted" anywhere in the code, but ERC4626Buffer or just ERC4626. File names, variables, comments, etc.

joaobrunoah:
> We're trying not to say "Boosted" anywhere in the code, but ERC4626Buffer or just ERC4626. File names, variables, comments, etc.

Yep, I'm aware. I'm already changing the code, the new fork tests don't mention boosted, but `yield-bearing pools`.

pkg/solidity-utils/contracts/helpers/BufferHelpers.sol
```diff
@@ -0,0 +1,67 @@
...
+
+        uint256 wrappedBalanceAsUnderlying = 0;
+        if (bufferBalance.getBalanceDerived() > 0) {
+            wrappedBalanceAsUnderlying = wrappedToken.convertToAssets(bufferBalance.getBalanceDerived());
+        }
+
+        uint256 surplus = 0;
+        if (underlyingBalance > wrappedBalanceAsUnderlying) {
+            unchecked {
+                surplus = (underlyingBalance - wrappedBalanceAsUnderlying) / 2;
```
elshan-eth, line 34:
```
underlyingBalance = 2
wrappedBalanceAsUnderlying = 1
```
surplus may overflow in this case but maybe it isn't problem

pkg/solidity-utils/contracts/math/FixedPoint.sol
```diff
@@ -152,4 +152,12 @@ library FixedPoint {
             result := mul(lt(x, ONE), sub(ONE, x))
         }
     }
+
+    function getAbsoluteDifference(uint256 a, uint256 b) internal pure returns (uint256) {
+        int256 difference = int256(a) - int256(b);
```
elshan-eth, line 157: int256(a) is not a safe cast and maybe overflow can be here

pkg/vault/contracts/BatchRouter.sol
elshan-eth, line 160: Cool!

elshan-eth (approved):
Very cool!

pkg/solidity-utils/contracts/helpers/BufferHelpers.sol
jubeira: In that case surplus would be 0, which should be fine.

pkg/vault/contracts/Vault.sol
```diff
@@ -1143,47 +1141,65 @@ contract Vault is IVaultMain, VaultCommon, Proxy {
...
-            }
+            uint256 vaultUnderlyingDelta;
+            uint256 vaultWrappedDelta;
 
-            uint256 calculatedUnderlyingDelta;
-            uint256 calculatedWrappedDelta;
+            // When wrapping and in query mode, the vault needs to have enough reserves of underlying token to do the
+            // wrap operation. If the vault does not have enough tokens to do the actual wrap, use ERC4626 preview.
+            bool shouldPreview = _isQueryContext() &&
+                _reservesOf[underlyingToken] < amountInUnderlying + bufferUnderlyingSurplus;
```
jubeira, line 1153: as discussed off-thread, while this extra check improves accuracy whenever possible, the worst case scenario is still the same, and it's probably better to just make the result of the query independent from the vault balances (here and in `unwrap`)

pkg/vault/contracts/BatchRouter.sol
jubeira: Got it; makes sense.

pkg/vault/contracts/Vault.sol
joaobrunoah: Done, code and tests are adjusted

pkg/vault/test/foundry/fork/YieldBearingPoolSwapBase.t.sol
```diff
@@ -0,0 +1,1097 @@
...
+        vm.label(address(_token1Fork), "token1");
+        _token2Fork = IERC20(ybToken2.asset());
+        vm.label(address(_token2Fork), "token2");
+    }
+
+    function _setupLPAndVault() private {
+        vm.startPrank(donorToken1);
+        // Donate token1 (underlying) to LP.
+        _token1Fork.transfer(lp, 100 * _token1BufferInitAmount);
+        // Donate to vault, so it has enough tokens to wrap and do not preview.
```
jubeira, line 903: Can we document somewhere that this is not the worst-case scenario?

jubeira (commented):
Great job with the changes and the tests!

The changes in the vault make sense. I think this is a good tradeoff between complexity and reusing as much code as possible to make queries work.
The tests are great; extending to new wrappers is now pretty straightforward, so great work with the test architecture. I'll do one more pass over them, as they are fairly complex and require some more attention, but overall they look pretty solid. Thanks for pushing this, awesome job!

pkg/vault/test/foundry/fork/YieldBearingPoolSwapBase.t.sol
joaobrunoah: I'll remove this, it's not needed anymore.

pkg/vault/contracts/Vault.sol
```diff
@@ -1273,32 +1294,53 @@ contract Vault is IVaultMain, VaultCommon, Proxy {
...
-            if (bufferWrappedSurplus > 0) {
-                bufferUnderlyingSurplus = wrappedToken.convertToAssets(bufferWrappedSurplus);
-            }
+            uint256 vaultUnderlyingDelta;
+            uint256 vaultWrappedDelta;
+
+            // When unwrapping and in query mode, the vault needs to have enough reserves of wrapped token to do the
+            // unwrap operation. If the vault does not have enough tokens to do the actual unwrap, use ERC4626 preview.
+            bool shouldPreview = _isQueryContext() &&
+                _reservesOf[IERC20(address(wrappedToken))] < amountInWrapped + bufferWrappedSurplus;
```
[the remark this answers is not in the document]
joaobrunoah: Done!

pkg/vault/test/foundry/fork/YieldBearingPoolSwapBase.t.sol
```diff
@@ -0,0 +1,1097 @@
...
+        _testExactOut(paths, true, expectedBufferDeltaTokenIn, expectedBufferDeltaTokenOut);
+    }
+
+    function testToken1InToken2OutOutOfBufferExactIn__Fork__Fuzz(uint256 amountIn) public {
+        // Test from 2x to 4x of tokenIn's buffer to make sure it's out of both buffer ranges but yield-bearing pool
+        // has enough tokens to trade (5x buffer init amount).
+        amountIn = bound(amountIn, 2 * _token1BufferInitAmount, 4 * _token1BufferInitAmount);
+        IBatchRouter.SwapPathExactAmountIn[] memory paths = _buildExactInPaths(_token1Fork, amountIn, 0);
+
+        // Since operation is out of buffer range, buffer balances should not change (wrap/unwrap tokens directly).
```
jubeira, line 197: This happens because we start balanced, and we just wrap / unwrap right?

pkg/vault/test/foundry/fork/YieldBearingPoolSwapBase.t.sol
```diff
@@ -0,0 +1,1097 @@
...
+        unbalancedToken2 = bound(unbalancedToken2, 0, _token2BufferInitAmount);
+        _unbalanceBuffers(unbalancedToken1, unbalancedToken2);
+
+        amountIn = bound(amountIn, 2 * _token1BufferInitAmount, 4 * _token1BufferInitAmount);
+        IBatchRouter.SwapPathExactAmountIn[] memory paths = _buildExactInPaths(_token1Fork, amountIn, 0);
+
+        int256 expectedBufferDeltaTokenIn;
+        int256 expectedBufferDeltaTokenOut;
+
+        // If the difference is smaller than MIN_TRADE_AMOUNT, buffer is not unbalanced.
```
jubeira, line 236: Can we rephrase this a bit? This describes the `else` clause right?

I'd describe the `if` here (if the buffer is unbalanced beyond `MIN_TRADE_AMOUNT`, we expect a delta in the buffer.

pkg/vault/test/foundry/fork/YieldBearingPoolSwapBase.t.sol
```diff
@@ -0,0 +1,1097 @@
...
+        if (unbalancedToken1 > _token1BufferInitAmount / 2 + MIN_TRADE_AMOUNT) {
+            expectedBufferDeltaTokenIn = int256(_token1BufferInitAmount / 2) - int256(unbalancedToken1);
+        } else {
+            // The unbalance operation was to remove underlying tokens, which generated a surplus of wrapped tokens.
+            // Since the operation will wrap token1 and there's no surplus of underlying tokens, the buffer will not
+            // rebalance.
+            expectedBufferDeltaTokenIn = 0;
+        }
+
+        // If the difference is smaller than MIN_TRADE_AMOUNT, buffer is not unbalanced.
```
jubeira, line 246: Same as above. Or am I missing something?

pkg/vault/test/foundry/fork/YieldBearingPoolSwapBase.t.sol
```diff
@@ -0,0 +1,1097 @@
...
+
+        // The error is proportional to the amount of decimals of token in and token out. If tokenIn has 6 decimals
+        // and tokenOut has 18 decimals, an error of 1 wei in amountOut of the first buffer generates an error in the
+        // order of 1e12 (1e18/1e6) in amountOut of the last buffer.
+        // But, if it's the opposite case, 1e6/1e18 is rounded to 0, but the max error is actually 1 (the error in the
+        // tokenOut token itself), so the division is incremented by 1.
+        // Finally, the error can be slightly bigger than the division of the factors, since each token has a rate. The
+        // error is multiplied by 2 to give some space for rounding errors related to the rates.
+        uint256 tolerance = (2 *
+            (paths[0].tokenIn == _token1Fork ? _token2Factor / _token1Factor : _token1Factor / _token2Factor)) + 1;
```
jubeira, line 498: Alright, fair enough.

Can we also assert that the relative error is < 0.1%, or 0.01%? I'd keep this absolute tolerance as well, but in practice this will also depend on the pool type we're using I suppose.

pkg/vault/test/foundry/fork/YieldBearingPoolSwapBase.t.sol
```diff
@@ -0,0 +1,1097 @@
...
+        assertEq(paths.length, 1, "Incorrect output array length");
+
+        assertEq(paths.length, tokens.length, "Output array length mismatch");
+        assertEq(tokens.length, amounts.length, "Output array length mismatch");
+
+        // Check results
+        if (vars.kind == SwapKind.EXACT_IN) {
+            // Rounding issues occurs in favor of vault.
+            assertLe(paths[0], vars.expectedDeltaTokenOut, "paths AmountOut must be <= expected amountOut");
+            assertLe(paths[0], vars.expectedDeltaTokenOut, "amounts AmountOut must be <= expected amountOut");
```
jubeira, line 630:
```suggestion
            assertLe(amounts[0], vars.expectedDeltaTokenOut, "amounts AmountOut must be <= expected amountOut");
```

pkg/vault/test/foundry/fork/YieldBearingPoolSwapBase.t.sol
```diff
@@ -0,0 +1,1097 @@
...
+            assertApproxEqAbs(paths[0], vars.expectedDeltaTokenIn, 5, "Wrong path count");
+            assertApproxEqAbs(amounts[0], vars.expectedDeltaTokenIn, 5, "Wrong amounts count");
+            assertEq(tokens[0], address(vars.tokenIn), "Wrong token for SwapKind");
+        }
+
+        // Tokens were transferred
+        assertLe(
+            vars.tokenIn.balanceOf(lp),
+            vars.lpBeforeSwapTokenIn - vars.expectedDeltaTokenIn,
+            "LP balance tokenIn must be <= expected balance"
```
jubeira, line 651: Can we clarify that this `Le` is to test the rounding in favor of the Vault with a comment?

pkg/vault/test/foundry/fork/YieldBearingPoolSwapBase.t.sol
```diff
@@ -0,0 +1,1097 @@
...
+            steps[1] = IBatchRouter.SwapPathStep({
+                pool: address(yieldBearingPool),
+                tokenOut: IERC20(address(ybToken2)),
+                isBuffer: false
+            });
+            steps[2] = IBatchRouter.SwapPathStep({ pool: address(ybToken2), tokenOut: _token2Fork, isBuffer: true });
+        }
+    }
+
+    function _unbalanceBuffers(uint256 unbalancedToken1, uint256 unbalancedToken2) private {
```
jubeira, line 837: Can we add a very small `@dev` block explaining that this can unbalance the buffer to either side of the equilibrium point?

jubeira (approved):
Another pass done.

The current approach for the queries looks much simpler; now that we're not trying to pursue maximum accuracy in every possible case, this seems to make much more sense.

I only have minor comments left, mostly questions and small clarification suggestions. My only 'big' comment is regarding the refactor of the variables in the wrap / unwrap path: let's keep the variables as local as possible to improve readability, and let's make wrap / unwrap as symmetrical as possible. Other than that, I don't think I have anything else to add here.

Thanks again for pushing this! Amazing job :100:

forge-test.sh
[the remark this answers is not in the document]
joaobrunoah: So, when I use `yarn test:forge`, sometimes I use extra inline params (like --match-contract ...), and these extra params need to be propagated to all `forge test` calls. I've tried to do that with an inline function, but for some reason npm wasn't accepting the syntax, and felt simpler to create the bash script than trying to understand why the npm script was failing.

pkg/vault/test/foundry/fork/YieldBearingPoolSwapBase.t.sol
joaobrunoah: Yes, I adjusted the comment.

pkg/vault/test/foundry/fork/YieldBearingPoolSwapBase.t.sol
joaobrunoah: Hmm, good point.. I tried to explain the MIN_TRADE_AMOUNT variable, but you're right. I'll change this comment.

pkg/vault/test/foundry/fork/YieldBearingPoolSwapBase.t.sol
joaobrunoah: Nope, you're right. I was trying to explain the MIN_TRADE_AMOUNT variable, not the if clause. I changed the comment to explain the `if`, It's more intuitive.

pkg/vault/contracts/Vault.sol
```diff
@@ -1143,47 +1141,65 @@ contract Vault is IVaultMain, VaultCommon, Proxy {
...
             // Gets the amount of underlying to wrap in order to rebalance the buffer.
-            uint256 bufferUnderlyingSurplus = _getBufferUnderlyingSurplus(bufferBalances, wrappedToken);
+            uint256 bufferUnderlyingSurplus = bufferBalances.getBufferUnderlyingSurplus(wrappedToken);
             uint256 bufferWrappedSurplus;
 
-            if (bufferUnderlyingSurplus > 0) {
-                bufferWrappedSurplus = wrappedToken.convertToShares(bufferUnderlyingSurplus);
-            }
+            uint256 vaultUnderlyingDelta;
+            uint256 vaultWrappedDelta;
```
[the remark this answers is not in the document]
joaobrunoah: Done!

pkg/vault/contracts/VaultExtension.sol
```diff
@@ -36,6 +36,7 @@ import {
 } from "@balancer-labs/v3-solidity-utils/contracts/openzeppelin/ReentrancyGuardTransient.sol";
 import { FixedPoint } from "@balancer-labs/v3-solidity-utils/contracts/math/FixedPoint.sol";
 import { PackedTokenBalance } from "@balancer-labs/v3-solidity-utils/contracts/helpers/PackedTokenBalance.sol";
+import { BufferHelpers } from "@balancer-labs/v3-solidity-utils/contracts/helpers/BufferHelpers.sol";
```
EndymionJkb, line 39: Can remove BufferHelpers (here and below) in VaultExtension; looks like they're used in the main Vault now. (Unless something that uses it gets moved back.)

pkg/vault/test/foundry/fork/YieldBearingPoolSwapBase.t.sol
```diff
@@ -0,0 +1,1148 @@
...
+        assertLe(
+            vars.tokenIn.balanceOf(lp),
+            vars.lpBeforeSwapTokenIn - vars.expectedDeltaTokenIn,
+            "LP balance tokenIn must be <= expected balance"
+        );
+        // If there were rounding issues, make sure it's not a big one (less than 5 wei).
+        assertApproxEqAbs(
+            vars.tokenIn.balanceOf(lp),
+            vars.lpBeforeSwapTokenIn - vars.expectedDeltaTokenIn,
+            5,
```
EndymionJkb, line 702: Could make this "5" a ROUNDING_DELTA constant

pkg/vault/test/foundry/fork/YieldBearingPoolSwapBase.t.sol
```diff
@@ -0,0 +1,1148 @@
...
+            });
+            steps[2] = IBatchRouter.SwapPathStep({ pool: address(ybToken2), tokenOut: _token2Fork, isBuffer: true });
+        }
+    }
+
+    /**
+     * @notice Unbalance buffers of ybToken1 and ybToken2.
+     * @dev This function can unbalance buffers to both sides, up to 50% of buffer initial liquidity.
+     * If the amount to unbalance token is smaller than 50% of buffer initial liquidity, underlying are wrapped, else
+     * wrapped are transformed in underlying.
```
EndymionJkb, line 890:
```suggestion
     * wrapped are transformed to underlying.
```

EndymionJkb (commented):
I'm impressed by the number of yield-bearing protocol tokens we're able to test here. It makes me feel a lot better about it.

I don't feel that I understand the nuances enough to thoroughly review the substantive changes, though I get the gist of it. I think we need some graphics to illustrate all the issues/cases (would help devs and also auditors).

pkg/vault/contracts/test/VaultMock.sol
```diff
@@ -578,12 +580,16 @@ contract VaultMock is IVaultMainMock, Vault {
...
 
     function internalGetBufferWrappedSurplus(IERC4626 wrappedToken) external view returns (uint256) {
         bytes32 bufferBalance = _bufferTokenBalances[IERC20(address(wrappedToken))];
-        return _getBufferWrappedSurplus(bufferBalance, wrappedToken);
+        return bufferBalance.getBufferWrappedSurplus(wrappedToken);
+    }
+
+    function getBufferTokenBalances(IERC4626 wrappedToken) external view returns (bytes32) {
+        return _bufferTokenBalances[IERC20(address(wrappedToken))];
     }
```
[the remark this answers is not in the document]
joaobrunoah: We already have a getter in vault admin (getBufferBalances), but it's for both underlying and wrapped uint256, not for the bytes32 (which I need to calculate the surpluses using BufferHelper library). Anyway, I changed the function's name to make it clear it only returns the bytes.

PR #791, opened 2024-07-18
Alternative to ERC4626 queries (V2 - Vault primitive queries)

### LINKED ISSUE #776: Fuzz test ERC4626 fork tests

### LINKED ISSUE #777: Difference between query and swap amounts for ERC4626 buffers
SEPOLIA - Block Number 6288761

When trying to swap wUSDC for wDAI on Sepolia, the following difference between query and actual operation occurs: (Buffers initialized with 100 tokens each). 

* exactIn:
  * 98985649125345246328n query
  * 98985649209193025139n swap
* exactOut:
  * 101014646n query
  * 101014645n swap

It's out of buffer range, so this difference was a bit expected. However, for an operation within buffer range (50 tokens in/out) the same error was caught:

* exactIn:
  * 49497694044904535279n,
  * 49497694086789921819n
* exactOut:
  * 50502354n,
  * 50502353n

joaobrunoah:
# Description

In #785 we needed to reproduce the whole logic of the wrapUnwrap vault primitive in VaultExtension, so that the queries of ERC4626 buffers are as close as possible from the actual operation. So, this is an alternative to use only Vault and remove the code duplicity from VaultExtension.

## Type of change

- [ ] Bug fix 
- [ ] New feature 
- [ ] Breaking change 
- [ ] Dependency changes
- [x] Code refactor / cleanup
- [ ] Other

## Checklist:

## Issue Resolution

Closes #777
Closes #776


pkg/vault/contracts/Vault.sol
```diff
@@ -1148,44 +1146,55 @@ contract Vault is IVaultMain, VaultCommon, Proxy {
...
+                    underlyingToken,
+                    IERC20(wrappedToken)
                 );
-
-                // EXACT_OUT requires the exact amount of wrapped tokens to be returned, so mint is called.
-                wrappedToken.mint(calculatedWrappedDelta, address(this));
-
-                // Remove approval, in case mint consumed less tokens than we approved, due to convert error.
-                underlyingToken.forceApprove(address(wrappedToken), 0);
             }
```
jubeira, line 1196: If we are in a query context, can't we update the reserves anyways?

jubeira: ah, but there's no real operation in case we preview. Maybe I'd explain why in a query context we can't always do this, and why it doesn't matter?

pkg/vault/contracts/Vault.sol
```diff
@@ -1148,44 +1146,55 @@ contract Vault is IVaultMain, VaultCommon, Proxy {
...
+                    isQueryContext &&
+                    underlyingToken.balanceOf(address(this)) < amountInUnderlying + bufferUnderlyingSurplus
+                ) {
+                    vaultUnderlyingDelta = wrappedToken.previewMint(vaultWrappedDelta);
+                } else {
+                    // Add convert error because mint can consume a different amount of tokens than we anticipated.
+                    underlyingToken.forceApprove(
+                        address(wrappedToken),
+                        _addConvertError(amountInUnderlying + bufferUnderlyingSurplus)
+                    );
```
jubeira, line 1180: Not specific to this PR, but maybe this was related to the old wrappers we were using?

According to https://eips.ethereum.org/EIPS/eip-4626, `mint` must mint exact shares. Now that we're reviewing these details, we might save a few bytes here :)

jubeira (commented):
First pass done.

There's more red than green in this PR, which is a good indication that this might be the right direction :)

I'll think about this a bit more, but I'm leaning towards this one.

pkg/vault/contracts/Vault.sol
joaobrunoah: Hmm, mint is a confusing function:
* We ask to mint a certain amount of shares
* The wrapper gets underlying, and return the exact amount of shares.

Here we approve the underlying tokens that mint will consume, which is not an exact number. We use amountInUnderlying and bufferUnderlyingSurplus, but they were calculated by convertToAssets, so they have errors.

jubeira: got it, thanks for the explanation! Can we add that detail to the comments please?

joaobrunoah: Done!

pkg/vault/contracts/Vault.sol
joaobrunoah: Done!

PR #787, opened 2024-07-17
General docs polish (interfaces pkg)

EndymionJkb:
# Description

This is a continuation of the previous doc improvement efforts, doing a comprehensive pass through the whole codebase to check for missing or stale comments, inconsistently named variables or files, etc.

There are just two non-docs changes, though not "code" changes: moving structs out of interfaces, which are similar to those already in VaultTypes. (I could factor them out, like I did for the computeDynamicFee function, but that was a rename, while this is just moving a struct.)

Started with the interfaces package (which of course affected most of the others). Subsequent PRs will look at the other packages.

## Type of change

- [ ] Bug fix 
- [ ] New feature 
- [ ] Breaking change 
- [ ] Dependency changes
- [X] Code refactor / cleanup
- [ ] Optimization: [ ] gas / [ ] bytecode
- [X] Documentation or wording changes
- [ ] Other

## Checklist:

- [ ] The diff is legible and has no extraneous changes
- [ ] Complex code has been commented, including external interfaces
- [ ] Tests have 100% code coverage
- [ ] The base branch is either `main`, or there's a description of how to merge

PR #792, opened 2024-07-18
Add functions to BatchRouter to add/remove liquidity for boosted pools. 

elshan-eth:
# Description

Adding/removing liquidity from boosted pools using buffers through existing methods in BatchRouter is expensive because they take a fee from each token. This PR adds a cheaper way for the user to do this. 

## Type of change

- [ ] Bug fix 
- [X] New feature 
- [ ] Breaking change 
- [ ] Dependency changes
- [ ] Code refactor / cleanup
- [ ] Other

## Checklist:


pkg/vault/contracts/BatchRouter.sol
```diff
@@ -671,6 +671,317 @@ contract BatchRouter is IBatchRouter, BatchRouterStorage, RouterCommon, Reentran
...
+
+        // Wrap given underlying tokens for wrapped tokens
+        for (uint256 i = 0; i < boostedPoolTokens.length; ++i) {
+            // Boosted pool tokens are the wrappers
+            IERC4626 wrappedToken = IERC4626(address(boostedPoolTokens[i]));
+            IERC20 underlyingToken = IERC20(wrappedToken.asset());
+
+            if (isStaticCall == false) {
+                _takeTokenIn(params.sender, underlyingToken, amountsIn[i], params.wethIsEth);
+            }
```
joaobrunoah, line 965: Here isStaticCall is needed, because the sender may not have tokens.

pkg/vault/test/foundry/BatchRouterBoosted.t.sol
```diff
@@ -0,0 +1,416 @@
...
+        permit2.approve(address(boostedPool), address(router), type(uint160).max, type(uint48).max);
+        permit2.approve(address(boostedPool), address(batchRouter), type(uint160).max, type(uint48).max);
+
+        IERC20(address(boostedPool)).approve(address(router), type(uint256).max);
+        IERC20(address(boostedPool)).approve(address(batchRouter), type(uint256).max);
+
+        vm.stopPrank();
+    }
+
+    modifier checkBuffersWhenStaticCall(address sender) {
```
joaobrunoah, line 127: This is nice!!

joaobrunoah (commented):
Nice Elshan! I just have some comments/suggestions. But my main doubt is: Does it really need to be in the BatchRouter? I have a feeling there's no batch in the operations, so this should go in the Router, but not sure if it's required to be in the BatchRouter for some reason.

pkg/vault/contracts/BatchRouter.sol
```diff
@@ -671,6 +671,317 @@ contract BatchRouter is IBatchRouter, BatchRouterStorage, RouterCommon, Reentran
         (pathAmountsIn, tokensIn, amountsIn) = _swapExactOutHook(params);
     }
 
+    /*******************************************************************************
+                            Yield-bearing token buffers
+    *******************************************************************************/
```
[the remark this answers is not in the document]
jubeira: I think we don't have that much space left in the Router, unfortunately.

pkg/vault/contracts/BatchRouter.sol
```diff
@@ -671,6 +671,317 @@ contract BatchRouter is IBatchRouter, BatchRouterStorage, RouterCommon, Reentran
...
+                    wrappedToken: wrappedToken,
+                    amountGivenRaw: amountsOut[i],
+                    limitRaw: 0, // We don't apply the limit here
+                    userData: params.userData
+                })
+            );
+
+            if (isStaticCall == false) {
+                _sendTokenOut(params.sender, underlyingToken, underlingAmountsOut[i], params.wethIsEth);
+            }
```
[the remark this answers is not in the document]
elshan-eth: Unfortunately, we need this one because without it, Vault would not send tokens to BatchRoute but BatchRoute would try to send it to user

EndymionJkb (commented):
Getting there; too bad there's no room in the regular router. An alternative would be making a new router for this; did we consider that? Is that better or worse for the intended users of this? (Could also separate queries out into their own router, I suppose.)

Lots of naming conventions we're trying to enforce now (see the cleanup PRs, e.g., #787).

pkg/vault/contracts/BatchRouter.sol
```diff
@@ -671,6 +671,317 @@ contract BatchRouter is IBatchRouter, BatchRouterStorage, RouterCommon, Reentran
...
+    function addLiquidityBoostedPoolUnbalancedHook(
+        AddLiquidityHookParams calldata params
+    )
+        external
+        nonReentrant
+        onlyVault
+        returns (uint256[] memory amountsIn, uint256 bptAmountOut, bytes memory returnData)
+    {
+        IERC20[] memory boostedPoolTokens = _vault.getPoolTokens(params.pool);
+        uint256[] memory wrappedAmounts = _wrapToken(
```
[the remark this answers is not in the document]
elshan-eth: I didn't quite get this point. what argument do you mean?

pkg/vault/test/foundry/BatchRouterBoosted.t.sol
```diff
@@ -0,0 +1,416 @@
...
+
+        (waDaiIdx, waUsdcIdx) = getSortedIndexes(address(waDAI), address(waUSDC));
+
+        initializeBuffers();
+        initializeBoostedPool();
+    }
+
+    function initializeBuffers() private {
+        // Create and fund buffer pools
+        vm.startPrank(lp);
```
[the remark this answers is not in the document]
elshan-eth: Do you think it’s better to do this in this PR or another separately?

pkg/vault/contracts/BatchRouter.sol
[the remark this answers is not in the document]
EndymionJkb: You could drop the final `uint256 limitRaw` parameter from `_wrapToken`, and inside `_wrapToken`, say:

uint256 limitRaw = kind == SwapKind.EXACT_IN ? 0 : type(uint256).max;

pkg/vault/contracts/BatchRouter.sol
```diff
@@ -671,6 +671,317 @@ contract BatchRouter is IBatchRouter, BatchRouterStorage, RouterCommon, Reentran
...
+    }
+
+    function _wrapToken(
+        AddLiquidityHookParams calldata params,
+        IERC20[] memory erc4626PoolTokens,
+        uint256[] memory amountsIn,
+        SwapKind kind,
+        uint256 limitRaw
+    ) private returns (uint256[] memory wrappedAmounts) {
+        // Non-registered pools will fail here.
```
jubeira, line 952: stale comment? This would fail in `getPoolTokens` right?

pkg/vault/contracts/BatchRouter.sol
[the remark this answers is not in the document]
jubeira: yeah I agree; we can get the limit inside `_wrap`.

Also, there's a constant defined for the max limit, which is `uint128.max`. This is because the limit is scaled inside the vault, and for tokens with less than 18 decimals `uint256.max` will revert when doing so (check batch router for reference).

pkg/vault/contracts/BatchRouter.sol
```diff
@@ -671,6 +671,317 @@ contract BatchRouter is IBatchRouter, BatchRouterStorage, RouterCommon, Reentran
...
+                })
+            );
+
+            if (isStaticCall == false) {
+                _sendTokenOut(params.sender, underlyingToken, underlyingAmountsOut[i], params.wethIsEth);
+            }
+        }
+    }
+
+    function _wrapToken(
```
jubeira, line 945:
```suggestion
    function _wrapTokens(
```

nit; naming

pkg/vault/contracts/BatchRouter.sol
```diff
@@ -671,6 +671,317 @@ contract BatchRouter is IBatchRouter, BatchRouterStorage, RouterCommon, Reentran
...
+        );
+    }
+
+    function addLiquidityERC4626PoolUnbalancedHook(
+        AddLiquidityHookParams calldata params
+    )
+        external
+        nonReentrant
+        onlyVault
+        returns (uint256[] memory amountsIn, uint256 bptAmountOut, bytes memory returnData)
```
jubeira, line 841: Since this hook is specific for unbalanced, I think returning `bptAmountOut` should be enough.

pkg/vault/contracts/BatchRouter.sol
```diff
@@ -671,6 +671,317 @@ contract BatchRouter is IBatchRouter, BatchRouterStorage, RouterCommon, Reentran
...
+        );
+
+        // Add wrapped amounts to the boosted pool
+        (amountsIn, bptAmountOut, returnData) = _vault.addLiquidity(
+            AddLiquidityParams({
+                pool: params.pool,
+                to: params.sender,
+                maxAmountsIn: wrappedAmounts,
+                minBptAmountOut: params.minBptAmountOut,
+                kind: params.kind,
```
jubeira, line 859: might as well hardcode the kind

pkg/vault/contracts/BatchRouter.sol
```diff
@@ -671,6 +671,317 @@ contract BatchRouter is IBatchRouter, BatchRouterStorage, RouterCommon, Reentran
...
+        nonReentrant
+        onlyVault
+        returns (uint256[] memory amountsIn, uint256 bptAmountOut, bytes memory returnData)
+    {
+        IERC20[] memory erc4626PoolTokens = _vault.getPoolTokens(params.pool);
+        uint256[] memory maxWrappedAmountsIn = new uint256[](erc4626PoolTokens.length);
+
+        for (uint256 i = 0; i < erc4626PoolTokens.length; ++i) {
+            maxWrappedAmountsIn[i] = IERC4626(address(erc4626PoolTokens[i])).previewDeposit(params.maxAmountsIn[i]);
+        }
```
jubeira, line 878: not wrong, but probably expensive :)

let's try using `convertToShares` instead; the result should be similar, but cheaper to compute.
There might be a few wei difference between `convert` and the actual operation because of rounding. We might need to test this in a fuzz / fork test.

pkg/vault/contracts/BatchRouter.sol
```diff
@@ -671,6 +671,317 @@ contract BatchRouter is IBatchRouter, BatchRouterStorage, RouterCommon, Reentran
...
+
+        // Add wrapped amounts to the boosted pool
+        (amountsIn, bptAmountOut, returnData) = _vault.addLiquidity(
+            AddLiquidityParams({
+                pool: params.pool,
+                to: params.sender,
+                maxAmountsIn: maxWrappedAmountsIn,
+                minBptAmountOut: params.minBptAmountOut,
+                kind: params.kind,
```
jubeira, line 887: same, might as well hardcode the kind.

pkg/vault/contracts/BatchRouter.sol
```diff
@@ -671,6 +671,317 @@ contract BatchRouter is IBatchRouter, BatchRouterStorage, RouterCommon, Reentran
...
+        );
+    }
+
+    function addLiquidityERC4626PoolProportionalHook(
+        AddLiquidityHookParams calldata params
+    )
+        external
+        nonReentrant
+        onlyVault
+        returns (uint256[] memory amountsIn, uint256 bptAmountOut, bytes memory returnData)
```
jubeira, line 871: Same; since this is specific to proportional add, maybe returning `amountsIn` is enough.

pkg/vault/contracts/BatchRouter.sol
```diff
@@ -671,6 +671,317 @@ contract BatchRouter is IBatchRouter, BatchRouterStorage, RouterCommon, Reentran
...
+        onlyVault
+        returns (uint256 bptAmountIn, uint256[] memory amountsOut, bytes memory returnData)
+    {
+        IERC20[] memory erc4626PoolTokens = _vault.getPoolTokens(params.pool);
+        uint256[] memory minWrappedAmountsIn = new uint256[](erc4626PoolTokens.length);
+        uint256[] memory underlyingAmountsOut = new uint256[](erc4626PoolTokens.length);
+
+        for (uint256 i = 0; i < erc4626PoolTokens.length; ++i) {
+            minWrappedAmountsIn[i] = IERC4626(address(erc4626PoolTokens[i])).previewWithdraw(params.minAmountsOut[i]);
+        }
```
jubeira, line 909: Same as above; let's try to use `convert` instead of `preview` here.

pkg/vault/contracts/BatchRouter.sol
```diff
@@ -671,6 +671,317 @@ contract BatchRouter is IBatchRouter, BatchRouterStorage, RouterCommon, Reentran
...
+                    wrappedToken: wrappedToken,
+                    amountGivenRaw: amountsOut[i],
+                    limitRaw: 0, // We don't apply the limit here
+                    userData: params.userData
+                })
+            );
+
+            if (isStaticCall == false) {
+                _sendTokenOut(params.sender, underlyingToken, underlyingAmountsOut[i], params.wethIsEth);
+            }
```
jubeira, line 941: Another option would be to do this outside the hook, but maybe it's just fine like this.

pkg/vault/test/foundry/BatchRouterERC4626Pool.t.sol
```diff
@@ -0,0 +1,409 @@
...
+            assertEq(beforeDAIBalance - afterDAIBalance, operationAmount, "DAI balance should decrease");
+        }
+        {
+            (uint256 afterWaUSDCBufferBalanceUnderlying, uint256 afterWaUSDCBufferBalanceWrapped) = vault
+                .getBufferBalance(waUSDC);
+            assertEq(
+                beforeWaUSDCBufferBalanceWrapped - afterWaUSDCBufferBalanceWrapped,
+                operationAmount,
+                "waUSDC wrapped buffer balance should decrease"
+            );
```
jubeira, line 211:
```suggestion
            assertEq(
                afterWaUSDCBufferBalanceWrapped,
                beforeWaUSDCBufferBalanceWrapped - operationAmount,
                "waUSDC wrapped buffer balance should decrease"
            );
```

Super minor, but maybe writing it like this is easier to read with this error message.

pkg/vault/test/foundry/BatchRouterERC4626Pool.t.sol
```diff
@@ -0,0 +1,409 @@
...
+            assertEq(
+                afterWaUSDCBufferBalanceUnderlying - beforeWaUSDCBufferBalanceUnderlying,
+                operationAmount,
+                "waUSDC underlying buffer balance should increase"
+            );
```
jubeira, line 216:
```suggestion
            assertEq(
                afterWaUSDCBufferBalanceUnderlying,
                beforeWaUSDCBufferBalanceUnderlying + operationAmount,
                "waUSDC underlying buffer balance should increase"
            );
```

pkg/vault/test/foundry/BatchRouterERC4626Pool.t.sol
```diff
@@ -0,0 +1,409 @@
...
+            assertEq(
+                beforeWaDAIBufferBalanceWrapped - afterWaDAIBufferBalanceWrapped,
+                operationAmount,
+                "waDAI wrapped buffer balance should decrease"
+            );
+            assertEq(
+                afterWaDAIBufferBalanceUnderlying - beforeWaDAIBufferBalanceUnderlying,
+                operationAmount,
+                "waDAI underlying buffer balance should increase"
+            );
```
jubeira, line 230:
```suggestion
            assertEq(
                afterWaDAIBufferBalanceWrapped,
                beforeWaDAIBufferBalanceWrapped - operationAmount,
                "waDAI wrapped buffer balance should decrease"
            );
            assertEq(
                afterWaDAIBufferBalanceUnderlying,
                beforeWaDAIBufferBalanceUnderlying + operationAmount,
                "waDAI underlying buffer balance should increase"
            );
```

pkg/vault/test/foundry/BatchRouterERC4626Pool.t.sol
```diff
@@ -0,0 +1,409 @@
...
+        }
+
+        assertApproxEqAbs(bptOut, expectBPTOut, DELTA, "BPT operationAmount should match expected");
+        assertEq(IERC20(address(erc4626Pool)).balanceOf(alice), bptOut, "BPT balance should increase");
```
jubeira, line 234:
```suggestion
        assertEq(IERC20(address(erc4626Pool)).balanceOf(alice), bptOut, "Alice BPT balance should increase");
```

pkg/vault/test/foundry/BatchRouterERC4626Pool.t.sol
```diff
@@ -0,0 +1,409 @@
...
+            "waDAI wrapped buffer balance should be the same"
+        );
+        assertEq(
+            afterWaDAIBufferBalanceUnderlying - beforeWaDAIBufferBalanceUnderlying,
+            0,
+            "waDAI underlying buffer balance should be the same"
+        );
+    }
+
+    function testAddLiquidityUnbalancedToERC4626Pool() public {
```
jubeira, line 173: We're checking the buffers and the LP's BPT balance, but we should also check the pool balance for the boosted pool (amount of wrapped tokens that leave buffers should be added to the boosted pool).

pkg/vault/test/foundry/BatchRouterERC4626Pool.t.sol
```diff
@@ -0,0 +1,409 @@
...
+        }
+
+        assertApproxEqAbs(bptOut, expectBPTOut, DELTA, "BPT operationAmount should match expected");
+        assertEq(IERC20(address(erc4626Pool)).balanceOf(alice), bptOut, "BPT balance should increase");
+    }
+
+    function testAddLiquidityUnbalancedToERC4626PoolWhenStaticCall() public checkBuffersWhenStaticCall(alice) {
+        uint256[] memory exactUnderlyingAmountsIn = [operationAmount, operationAmount].toMemoryArray();
+
+        vm.prank(alice, address(0));
```
jubeira, line 240: _prankStaticCall()?

pkg/vault/test/foundry/BatchRouterERC4626Pool.t.sol
```diff
@@ -0,0 +1,409 @@
...
+        batchRouter.queryAddLiquidityUnbalancedToERC4626Pool(erc4626Pool, exactUnderlyingAmountsIn, new bytes(0));
+    }
+
+    function testAddLiquidityProportionalToERC4626Pool() public {
```
jubeira, line 244: same comments from above apply:
- assertions (try to better match `assertEq` with error messages)
- check boosted pool balances before and after

pkg/vault/test/foundry/BatchRouterERC4626Pool.t.sol
```diff
@@ -0,0 +1,409 @@
...
+        vm.prank(alice, address(0));
+        batchRouter.queryAddLiquidityProportionalToERC4626Pool(
+            erc4626Pool,
+            maxAmountsIn,
+            operationAmount,
+            new bytes(0)
+        );
+    }
+
+    function testRemoveLiquidityProportionalToERC4626Pool() public {
```
jubeira, line 329: (same as above)

jubeira (commented):
Great work @elshan-eth, looks almost cooked!

First pass done; I have a bunch of comments, mostly minor.
It might make sense to do some fuzz testing with real wrappers for this one, but let's leave that for a second iteration. For now I'll approve after the comments are addressed :)

Thanks for pushing this!

pkg/vault/test/foundry/BatchRouterBoosted.t.sol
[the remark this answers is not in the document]
elshan-eth: https://github.com/balancer/balancer-v3-monorepo/issues/808

pkg/vault/contracts/BatchRouter.sol
elshan-eth: yes, but we use this style everywhere in Router/BatchRouter. it seems it shouldn't have much gas saving

pkg/vault/test/foundry/BatchRouterERC4626Pool.t.sol
elshan-eth: `_prankStaticCall` also mocks the sender to zero address but we need the sender because vault checks tokens from the sender address

elshan-eth: Hm, maybe I'm mistaken

elshan-eth: yes, I'm right

pkg/interfaces/contracts/vault/IBatchRouter.sol
```diff
@@ -93,6 +93,57 @@ interface IBatchRouter {
         bytes calldata userData
     ) external payable returns (uint256[] memory pathAmountsIn, address[] memory tokensIn, uint256[] memory amountsIn);
 
+    /**
+     * @notice Add arbitrary amounts of underlying tokens to an ERC4626 pool through the buffer.
+     * @param pool Address of the liquidity pool
```
EndymionJkb, line 98: I think we need to define here (and for the other new functions) what an "ERC4626 pool" is. We've lost a bit of flexibility compared to the original "token-based" design, where you could mix ERC4626 and standard tokens in a pool. You can still do that, but these add/remove liquidity functions require that *all* pool tokens are ERC4626, or the wrap/unwrap functions would revert.

`@dev An "ERC4626 pool" is one in which all tokens conform to the IERC4626 yield-bearing token standard (e.g., waDAI).`

pkg/interfaces/contracts/vault/IBatchRouter.sol
```diff
@@ -93,6 +93,57 @@ interface IBatchRouter {
         bytes calldata userData
     ) external payable returns (uint256[] memory pathAmountsIn, address[] memory tokensIn, uint256[] memory amountsIn);
 
+    /**
```
EndymionJkb, line 96: Could add a header here for "ERC4626 Pools," and say generally that these functions facilitate liquidity operations using standard tokens on ERC4626 pools, which contain only wrapped tokens. So we are abstracting away the underlying wrap/unwrap interactions, allowing them to do in a single transparent operation what they would otherwise need to build into a batch operation.

```
    /***************************************************************************
                                   ERC4626 Pools

    // These functions allow interacting with ERC4626 Pools (which are composed of wrapped ERC4626 tokens) using only
    // underlying standard tokens. For instance, with `addLiquidityUnbalancedToERC4626Pool` it is possible to add
    // liquidity to an ERC4626 Pool with [waDAI, waUSDC], using only DAI, only USDC, or an arbitrary amount of both.
    // If the ERC4626 buffers in the Vault have liquidity, these will be used to avoid wrapping/unwrapping through
    // the wrapped token interface, saving gas.
    //
    // For instance, adding only DAI to the pool above (and assuming an aDAI buffer with enough liquidity), would
    // pull in the DAI from the user, swap it for waDAI in the internal Vault buffer, and deposit the waDAI into the
    // ERC4626 pool: 1) without having to do any expensive ERC4626 wrapping operations; and 2) without requiring the
    // user to construct a batch operation containing the buffer swap.
```

pkg/vault/contracts/BatchRouter.sol
```diff
@@ -677,6 +677,300 @@ contract BatchRouter is IBatchRouter, BatchRouterStorage, RouterCommon, Reentran
...
+        AddLiquidityHookParams calldata params
+    ) external nonReentrant onlyVault returns (uint256[] memory amountsIn) {
+        IERC20[] memory erc4626PoolTokens = _vault.getPoolTokens(params.pool);
+        uint256[] memory maxWrappedAmountsIn = new uint256[](erc4626PoolTokens.length);
+
+        for (uint256 i = 0; i < erc4626PoolTokens.length; ++i) {
+            maxWrappedAmountsIn[i] = IERC4626(address(erc4626PoolTokens[i])).convertToShares(params.maxAmountsIn[i]);
+        }
+
+        // Add wrapped amounts to the boosted pool
```
EndymionJkb, line 875:
```suggestion
        // Add wrapped amounts to the ERC4626 pool
```
Can't say "boosted" in code

pkg/vault/contracts/BatchRouter.sol
```diff
@@ -677,6 +677,300 @@ contract BatchRouter is IBatchRouter, BatchRouterStorage, RouterCommon, Reentran
...
+    ) external nonReentrant onlyVault returns (uint256 bptAmountOut) {
+        IERC20[] memory erc4626PoolTokens = _vault.getPoolTokens(params.pool);
+        uint256[] memory wrappedAmounts = _wrapTokens(
+            params,
+            erc4626PoolTokens,
+            params.maxAmountsIn,
+            SwapKind.EXACT_IN
+        );
+
+        // Add wrapped amounts to the boosted pool
```
EndymionJkb, line 852:
```suggestion
        // Add wrapped amounts to the ERC4626 pool
```

pkg/vault/contracts/BatchRouter.sol
```diff
@@ -677,6 +677,300 @@ contract BatchRouter is IBatchRouter, BatchRouterStorage, RouterCommon, Reentran
...
+        uint256[] memory amountsIn,
+        SwapKind kind
+    ) private returns (uint256[] memory wrappedAmounts) {
+        wrappedAmounts = new uint256[](erc4626PoolTokens.length);
+
+        bool isStaticCall = EVMCallModeHelpers.isStaticCall();
+
+        // Wrap given underlying tokens for wrapped tokens
+        for (uint256 i = 0; i < erc4626PoolTokens.length; ++i) {
+            // Boosted pool tokens are the wrappers
```
EndymionJkb, line 948:
```suggestion
            // ERC4626 pool tokens are all wrapped tokens.
```
And remember periods after non-inline comments (here and elsewhere).

```
// This comment needs a period.
uint256 pct = 1e18; // 100% doesn't need a period
```

EndymionJkb (commented):
Looks great! I'm glad we had the 6k extra space to accommodate this.

It would ideally be in the regular router, but unless we want multiple routers (e.g., a separate query router, or an ERC4626 router with only this in it), which I don't think we do, seems like it has to go here.

pkg/vault/contracts/BatchRouter.sol
```diff
@@ -677,6 +677,300 @@ contract BatchRouter is IBatchRouter, BatchRouterStorage, RouterCommon, Reentran
...
+
+    function addLiquidityERC4626PoolProportionalHook(
+        AddLiquidityHookParams calldata params
+    ) external nonReentrant onlyVault returns (uint256[] memory amountsIn) {
+        IERC20[] memory erc4626PoolTokens = _vault.getPoolTokens(params.pool);
+        uint256[] memory maxWrappedAmountsIn = new uint256[](erc4626PoolTokens.length);
+
+        for (uint256 i = 0; i < erc4626PoolTokens.length; ++i) {
+            maxWrappedAmountsIn[i] = IERC4626(address(erc4626PoolTokens[i])).convertToShares(params.maxAmountsIn[i]);
+        }
```
jubeira, line 873:
```suggestion
```
Same; let's apply limits on the second step.

pkg/vault/contracts/BatchRouter.sol
```diff
@@ -677,6 +677,300 @@ contract BatchRouter is IBatchRouter, BatchRouterStorage, RouterCommon, Reentran
...
+
+        // Add wrapped amounts to the boosted pool
+        (amountsIn, , ) = _vault.addLiquidity(
+            AddLiquidityParams({
+                pool: params.pool,
+                to: params.sender,
+                maxAmountsIn: maxWrappedAmountsIn,
```
jubeira, line 880: This should be an array with `_MAX_AMOUNT` on each element.

pkg/vault/contracts/BatchRouter.sol
```diff
@@ -677,6 +677,300 @@ contract BatchRouter is IBatchRouter, BatchRouterStorage, RouterCommon, Reentran
...
+                minBptAmountOut: params.minBptAmountOut,
+                kind: params.kind,
+                userData: params.userData
+            })
+        );
+
+        _wrapTokens(params, erc4626PoolTokens, amountsIn, SwapKind.EXACT_OUT);
```
jubeira, line 887: send the limits array to `_wrapTokens` and apply the limits there.

jubeira (commented):
Second pass done.
Sorry for the late comment; I think we need to apply the given limits in a slightly different manner. You can use the regular batch swap as a reference: limits should only apply to the last step in the chain, which determines what the user gets (or needs to send the vault).

The problem with `convert` is that you might have some accuracy issues, and limits should be exact to the last wei.

pkg/interfaces/contracts/vault/IBatchRouter.sol
elshan-eth: Thanks!

pkg/vault/test/foundry/BatchRouterERC4626Pool.t.sol
```diff
@@ -0,0 +1,504 @@
...
+        (uint256 beforeWaUSDCBufferBalanceUnderlying, uint256 beforeWaUSDCBufferBalanceWrapped) = vault
+            .getBufferBalance(waUSDC);
+        (uint256 beforeWaDAIBufferBalanceUnderlying, uint256 beforeWaDAIBufferBalanceWrapped) = vault.getBufferBalance(
+            waDAI
+        );
+
+        _;
+
+        uint256 afterUSDCBalance = usdc.balanceOf(sender);
+        assertEq(beforeUSDCBalance - afterUSDCBalance, 0, "USDC balance should be the same");
```
jubeira, line 138: nit (here and elsewhere): if it's just before == after (i.e. no extra operation between before and after), it reads a bit easier like this:
```suggestion
        assertEq(beforeUSDCBalance, afterUSDCBalance, "USDC balance should be the same");
```

jubeira (approved):
LGTM! Just some minor final comments; otherwise it's looking good. Great job!

On the other hand, perhaps it's also worth adding a few tests that hit the limits to test that aspect?

PR #797, opened 2024-07-22
Simplify base pool math

jubeira:
# Description

This PR saves some bytecode and gas by simplifying math optimizations to `BasePoolMath`.

There are a few instances where we're doing `a.mulDown(b).divDown(c)` (or up). In those instances, as long as we multiply first and divide later we don't need to use FP math, and it should also be more precise not to use FP math. Since FP operations require a multiplication and a division, this approach is more efficient even when inlining the math inside a for-loop, without saving intermediate results.

## Type of change

- [ ] Bug fix 
- [ ] New feature 
- [ ] Breaking change 
- [ ] Dependency changes
- [ ] Code refactor / cleanup
- [x] Optimization: [x] gas / [x] bytecode
- [ ] Other

## Checklist:

## Issue Resolution

N/A


pkg/solidity-utils/contracts/math/BasePoolMath.sol
```diff
@@ -206,9 +198,8 @@ library BasePoolMath {
         uint256 amountIn = newBalance - currentBalances[tokenInIndex];
 
         // Calculate the non-taxable amount, which is the new balance proportionate to the BPT minted.
-        // Round the `nonTaxableBalance` down to favor the protocol by increasing the taxable amount, which charges
-        // higher swap fees, ultimately increasing the amount of `tokenIn` that will be transferred from the caller.
-        uint256 nonTaxableBalance = newSupply.mulDown(currentBalances[tokenInIndex]).divDown(totalSupply);
+        // Since we multiply and divide we don't need to use FP math.
+        uint256 nonTaxableBalance = (newSupply * currentBalances[tokenInIndex]) / totalSupply;
```
joaobrunoah, line 202: I believe we're using BasePoolMath only for scaled18 values, but it's worth noting that `currentBalances[tokenInIndex]` must be an 18 decimals number.

pkg/solidity-utils/contracts/math/BasePoolMath.sol
```diff
@@ -263,12 +254,11 @@ library BasePoolMath {
...
-        // rounding in favor of the protocol. We round the first term up to subtract from more and get a higher
-        // `taxableAmount`, which charges higher swap fees, augmenting the amount of BPT that will be burned.
-        uint256 taxableAmount = computeInvariant(newBalances).divUp(currentInvariant).mulUp(
-            currentBalances[tokenOutIndex]
-        ) - newBalances[tokenOutIndex];
+        // Calculate the taxable amount by subtracting the new balance from the equivalent proportional balance.
+        // Since we multiply and divide we don't need to use FP math.
+        uint256 taxableAmount = (computeInvariant(newBalances) * currentBalances[tokenOutIndex]) /
+            currentInvariant -
+            newBalances[tokenOutIndex];
```
joaobrunoah, line 261: This one probably divided first and multiplied later to avoid overflow since an invariant may be a big number. If this is the case, maybe it's not worth changing this one.

joaobrunoah (commented):
Just a couple of comments, but looks good.

pkg/solidity-utils/contracts/math/BasePoolMath.sol
jubeira: As discussed off-thread, the maximum intermediate value might be larger with this approach, but I don't think that's going to be a problem.

If I'm not mistaken, the existing code in `main` is a result of an optimization after the first iteration of the code, where `invariantRatio` was an intermediate value in the stack: https://github.com/balancer/balancer-v3-monorepo/pull/145/files#diff-7732d3024d950d26c8460e3d18e8b260441c8527d497da76b921a66e6477b313R225-R227. I don't think there's a deeper reason behind the previous order.

On the other hand, the order using FP math is not _that_ important, but with raw math as with this PR we must multiply first to avoid losing precision.

pkg/solidity-utils/contracts/math/BasePoolMath.sol
jubeira: I think that was the case already right? FP math amplifies or divides by 1e18, and that only works if we're multiplying 18 decimal values.

pkg/solidity-utils/contracts/math/BasePoolMath.sol
EndymionJkb: We're losing the rounding direction though, right? It's going to round down instead of up. Do we need to say `(computeInvariant(newBalances) * currentBalances[tokenOutIndex] + 1) / currentInvariant`

pkg/solidity-utils/contracts/math/BasePoolMath.sol
```diff
@@ -39,15 +39,11 @@ library BasePoolMath {
...
-        // multiplication and division.
-
-        uint256 bptRatio = bptAmountOut.divUp(bptTotalSupply);
-
         // Create a new array to hold the amounts of each token to be deposited.
         amountsIn = new uint256[](balances.length);
         for (uint256 i = 0; i < balances.length; ++i) {
-            amountsIn[i] = balances[i].mulUp(bptRatio);
+            // Since we multiply and divide we don't need to use FP math.
+            amountsIn[i] = (balances[i] * bptAmountOut) / bptTotalSupply;
```
EndymionJkb, line 46: Do we need to add 1 to preserve the rounding direction? `(balances[i] * bptAmountOut + 1) / bptTotalSupply`

EndymionJkb (commented):
Not sure about the rounding here. If we use straight division, it will truncate (i.e., round down). Do we still need to preserve the original rounding directions?

elshan-eth:
Cool! This seems to not affect the accuracy of the calculations

elshan-eth (approved):
Cool! Do I understand correctly that there is no longer any logic with rounding up, we always round down?

jubeira:
Thanks for the feedback.

I've adjusted the code to preserve the rounding direction in the places I think it's worth it, adding comments about it.
Savings are not as good as they were, but it's still something.

pkg/solidity-utils/contracts/math/BasePoolMath.sol
jubeira: yeah in this case it might still make sense. Adjusted.

jubeira: > `(balances[i] * bptAmountOut + 1) / bptTotalSupply`

I think that's not the formula though. Probably better to do a raw multiplication and then divUp instead.

EndymionJkb (approved):
It lowers the improvement, but it seems a bit dangerous to just remove the rounding (though you did change removeLiquiditySingleToken, where it was rounding up and up again, and now it rounds down and then up). Maybe a question for the auditors.

elshan-eth (approved):
Cool!

jubeira:
> It lowers the improvement, but it seems a bit dangerous to just remove the rounding (though you did change removeLiquiditySingleToken, where it was rounding up and up again, and now it rounds down and then up). Maybe a question for the auditors.

This should be fine; the result of the multiplication is a fixed point with 36 decimals which doesn't lose precision; the division is the one that can't be represented perfectly up to the very last wei and needs to be rounded either up or down.

PR #801, opened 2024-07-23
E2E swap test (Do/Undo ExactIn and ExactOut)

### LINKED ISSUE #288: Test do / undo operation end to end
Similar to the work in #270, fuzz operations that should revert the effects of each other (e.g. swap in one direction, then the opposite) and verify that the caller doesn't extract value out of it.

Choose ranges carefully, and consider fuzzing individual variables.

joaobrunoah:
# Description

This PR creates E2E tests (Do/Undo) of swaps for Stable, Weighted and PoolMock (Linear) pool types.

## Type of change

- [ ] Bug fix 
- [ ] New feature 
- [ ] Breaking change 
- [ ] Dependency changes
- [ ] Code refactor / cleanup
- [x] Tests
- [ ] Other

## Checklist:

## Issue Resolution

Closes #288.


pkg/pool-weighted/test/foundry/E2eSwap.t.sol
```diff
@@ -0,0 +1,273 @@
...
+
+contract E2eSwapWeightedTest is E2eSwapTest {
+    using ArrayHelpers for *;
+    using FixedPoint for uint256;
+
+    uint256 internal constant DEFAULT_SWAP_FEE = 1e16; // 1%
+
+    uint256 internal poolCreationNonce;
+    uint256 private _daiIdx;
+    uint256 private _usdcIdx;
```
EndymionJkb, line 32: Other tests just say `uint256 internal daiIdx`

pkg/vault/contracts/test/ProtocolFeeControllerMock.sol
```diff
@@ -18,4 +18,8 @@ contract ProtocolFeeControllerMock is ProtocolFeeController {
     ) external view returns (address poolCreator, uint256 creatorSwapFeePercentage, uint256 creatorYieldFeePercentage) {
         return (_poolCreators[pool], _poolCreatorSwapFeePercentages[pool], _poolCreatorYieldFeePercentages[pool]);
     }
+
+    function manualSetPoolCreator(address pool, address poolCreator) external {
```
EndymionJkb, line 22: Would add a comment here that Standard Balancer Pools specifically disallow pool creators to be passed in through PoolRoleAccounts; otherwise, this wouldn't be necessary.

EndymionJkb (commented):
Wasn't sure why this used different conventions for token naming, etc. See #827. (That one fork test is failing again, too.)

pkg/vault/test/foundry/E2eSwap.t.sol
```diff
@@ -0,0 +1,563 @@
...
+        );
+        assertLe(
+            balancesAfter.userTokens[_idxToken2],
+            balancesBefore.userTokens[_idxToken2],
+            "Wrong sender token2 balance"
+        );
+    }
+
+    function testDoExactInUndoExactInLiquidity__Fuzz(uint256 liquidityToken1, uint256 liquidityToken2) public {
+        liquidityToken1 = bound(liquidityToken1, poolInitAmount / 10, 10 * poolInitAmount);
```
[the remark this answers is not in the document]
joaobrunoah: I thought this E2eSwap file could be used in the future to run fork fuzz tests, that's why I left it more generic, but you're right, I'll rename them

pkg/vault/test/foundry/E2eSwap.t.sol
```diff
@@ -0,0 +1,563 @@
...
+import { PoolMock } from "../../contracts/test/PoolMock.sol";
+import { BaseVaultTest } from "./utils/BaseVaultTest.sol";
+
+contract E2eSwapTest is BaseVaultTest {
+    using ArrayHelpers for *;
+
+    IERC20 internal token1;
+    IERC20 internal token2;
+    uint256 private _idxToken1;
+    uint256 private _idxToken2;
```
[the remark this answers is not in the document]
joaobrunoah: As I said, I think it's nice if we can support in this file tests with other tokens, maybe fork, maybe tokens with a different decimal configuration, so I'd leave it as generic as possible.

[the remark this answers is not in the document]
jubeira: This is not a unit test, but an e2e test that verifies certain properties indirectly; whatever happens in the middle is a black box.

In this case we don't need a concrete result; we are just verifying that the system does not leak value when it's bombarded with requests under multiple conditions. 

In that context, having tokens with different decimals is another valid knob to adjust.

pkg/pool-weighted/contracts/test/WeightedPoolMock.sol
```diff
@@ -0,0 +1,62 @@
...
+    WeightedPoolDynamicData,
+    WeightedPoolImmutableData
+} from "@balancer-labs/v3-interfaces/contracts/pool-weighted/IWeightedPool.sol";
+import { ISwapFeePercentageBounds } from "@balancer-labs/v3-interfaces/contracts/vault/ISwapFeePercentageBounds.sol";
+import { IVault } from "@balancer-labs/v3-interfaces/contracts/vault/IVault.sol";
+import { IVaultErrors } from "@balancer-labs/v3-interfaces/contracts/vault/IVaultErrors.sol";
+import { SwapKind, PoolSwapParams } from "@balancer-labs/v3-interfaces/contracts/vault/VaultTypes.sol";
+import { IBasePool } from "@balancer-labs/v3-interfaces/contracts/vault/IBasePool.sol";
+
+import { BalancerPoolToken } from "@balancer-labs/v3-vault/contracts/BalancerPoolToken.sol";
```
elshan-eth, line 16: There are several imports that are not used

pkg/pool-stable/test/foundry/E2eSwap.t.sol
```diff
@@ -0,0 +1,86 @@
+// SPDX-License-Identifier: GPL-3.0-or-later
+
+pragma solidity ^0.8.24;
+
+import "forge-std/Test.sol";
+
+import { IAuthentication } from "@balancer-labs/v3-interfaces/contracts/solidity-utils/helpers/IAuthentication.sol";
+import { IProtocolFeeController } from "@balancer-labs/v3-interfaces/contracts/vault/IProtocolFeeController.sol";
```
elshan-eth, line 8: There are several imports that are not used

pkg/vault/test/foundry/E2eSwap.t.sol
```diff
@@ -0,0 +1,516 @@
...
+import { IERC20 } from "@openzeppelin/contracts/token/ERC20/IERC20.sol";
+
+import { IAuthentication } from "@balancer-labs/v3-interfaces/contracts/solidity-utils/helpers/IAuthentication.sol";
+import { IProtocolFeeController } from "@balancer-labs/v3-interfaces/contracts/vault/IProtocolFeeController.sol";
+
+import { ArrayHelpers } from "@balancer-labs/v3-solidity-utils/contracts/helpers/ArrayHelpers.sol";
+import { ERC20TestToken } from "@balancer-labs/v3-solidity-utils/contracts/test/ERC20TestToken.sol";
+import { FixedPoint } from "@balancer-labs/v3-solidity-utils/contracts/math/FixedPoint.sol";
+
+import { PoolMock } from "../../contracts/test/PoolMock.sol";
```
elshan-eth, line 16: unused import

pkg/vault/test/foundry/E2eSwap.t.sol
```diff
@@ -0,0 +1,516 @@
...
+        authorizer.grantRole(
+            feeControllerAuth.getActionId(IProtocolFeeController.setGlobalProtocolSwapFeePercentage.selector),
+            admin
+        );
+
+        _setUpVariables();
+
+        // Set protocol and creator fees to 50%, so we can measure the charged fees.
+        vm.prank(admin);
+        feeController.setGlobalProtocolSwapFeePercentage(50e16);
```
elshan-eth, line 53: Maybe it would be good if the value will be a constant. WDYT?

pkg/vault/test/foundry/E2eSwap.t.sol
```diff
@@ -0,0 +1,516 @@
...
+
+        vm.prank(poolCreator);
+        // Set pool creator fee to 100%, so protocol + creator fees equals the total charged fees.
+        feeController.setPoolCreatorSwapFeePercentage(pool, FixedPoint.ONE);
+
+        (tokenAIdx, tokenBIdx) = _getTokenIndexes();
+
+        // Donate tokens to vault, so liquidity tests are possible.
+        tokenA.mint(address(vault), 100 * poolInitAmount);
```
elshan-eth, line 62: At `100 * poolInitAmount` the same

elshan-eth (approved):
Cool! Good job!

pkg/vault/test/foundry/E2eSwap.t.sol
```diff
@@ -0,0 +1,516 @@
...
+    }
+
+    function _checkUserBalancesAndPoolInvariant(
+        BaseVaultTest.Balances memory balancesBefore,
+        BaseVaultTest.Balances memory balancesAfter,
+        uint256 feesTokenA,
+        uint256 feesTokenB
+    ) private view {
+        // Pool invariant cannot decrease after the swaps. All fees should be paid by the user.
+        assertGe(balancesAfter.poolInvariant, balancesBefore.poolInvariant, "Pool invariant is smaller than before");
```
elshan-eth, line 488: Can we also check here the vault balances?

elshan-eth (approved):

pkg/vault/test/foundry/E2eSwap.t.sol
joaobrunoah: I think, since the test is prepared to deal with different decimals now, creating a constant for 100 * poolInitAmount doesn't make much sense anymore (because poolInitAmount is different, depending on the decimals)

pkg/pool-weighted/test/foundry/E2eSwap.t.sol
```diff
@@ -0,0 +1,311 @@
...
+            MAX_UINT128,
+            false,
+            bytes("")
+        );
+        uint256 feesTokenA = vault.getAggregateSwapFeeAmount(pool, tokenA);
+        vm.stopPrank();
+
+        BaseVaultTest.Balances memory balancesAfter = getBalances(sender);
+        // getBalances measures the poolInvariant of `pool` variable. In this test another pool variable is used, so
+        // poolInvariant should be overriden.
```
EndymionJkb, line 121:
```suggestion
        // `BaseVaultTest.getBalances` checks the poolInvariant of `pool`. In this test, a different pool is used, so
        // poolInvariant should be overwritten.
```
Here and the other places this text appears.

pkg/pool-weighted/test/foundry/E2eSwap.t.sol
```diff
@@ -0,0 +1,311 @@
...
+            MAX_UINT128,
+            false,
+            bytes("")
+        );
+
+        uint256 feesTokenA = vault.getAggregateSwapFeeAmount(pool, tokenA);
+
+        // In the first swap, the trade was exactAmountInDo => exactAmountOut (tokenB) + feesTokenA (tokenA). So, if
+        // there were no fees, trading `exactAmountInDo - feesTokenA` would get exactAmountOut. Therefore, a swap
+        // with exact_out `exactAmountInDo - feesTokenA` is comparable with `exactAmountOut`, given that the fees are
```
EndymionJkb, line 167:
```suggestion
        // with exact_out `exactAmountInDo - feesTokenA` is comparable to `exactAmountOut`, given that the fees are
```

pkg/pool-weighted/test/foundry/E2eSwap.t.sol
```diff
@@ -0,0 +1,311 @@
...
+        PoolHooksMock(poolHooksContract).allowFactory(address(factory));
+
+        WeightedPool newPool = WeightedPool(
+            factory.create(
+                "50/50 Weighted Pool",
+                "50_50WP",
+                vault.buildTokenConfig(tokens.asIERC20()),
+                [uint256(50e16), uint256(50e16)].toMemoryArray(),
+                roleAccounts,
+                DEFAULT_SWAP_FEE, // 1% swap fee, but test will override it.
```
EndymionJkb, line 212:
```suggestion
                DEFAULT_SWAP_FEE, // 1% swap fee, but test will override it
```
Might be other places too; inline comments don't have periods

pkg/pool-weighted/test/foundry/E2eSwap.t.sol
```diff
@@ -0,0 +1,311 @@
...
+                poolHooksContract,
+                false, // Do not enable donations
+                false, // Do not disable unbalanced add/remove liquidity
+                // NOTE: sends a unique salt
```
EndymionJkb, line 216:
```suggestion
                // NOTE: sends a unique salt.
```

pkg/pool-weighted/test/foundry/E2eSwap.t.sol
```diff
@@ -0,0 +1,311 @@
...
+                bytes32(poolCreationNonce++)
+            )
+        );
+        vm.label(address(newPool), label);
+
+        // Cannot set pool creator directly with stable pool factory.
```
EndymionJkb, line 222:
```suggestion
        // Cannot set the pool creator directly on a standard Balancer stable pool factory.
```
Here and elsewhere.

pkg/pool-weighted/test/foundry/E2eSwapBigDecimalsToSmall.t.sol
```diff
@@ -0,0 +1,30 @@
...
+
+import "forge-std/Test.sol";
+
+import { ERC20TestToken } from "@balancer-labs/v3-solidity-utils/contracts/test/ERC20TestToken.sol";
+
+import { E2eSwapWeightedTest } from "./E2eSwap.t.sol";
+
+contract E2eSwapWeightedBigDecimalsToSmallTest is E2eSwapWeightedTest {
+    function setUpTokens() internal override {
+        tokenA = new ERC20TestToken("tokenA", "tokenA", 12);
```
EndymionJkb, line 13: Are the decimal choices deliberate? Here it is 6 & 12; elsewhere we have 8 & 14.

Should it work with the full range of decimals? We'd talked about putting in the docs that although we technically support even 0-decimal tokens (and those silly 2-decimal stable coins), those aren't going to work well in an AMM.

The lowest decimal token in common use would be USDC and friends, with 6, as you have it here. (And then WBTC & Co. with 8 decimals.) We can say we don't recommend using < 6 decimals; do so at your own risk.

Should this test from 6 to 18 then? (Could also try < 6, just to see if/where it breaks, even if we leave it at 6 in the end.)

Could we have a test that loops through 6 to 17, testing them all against 18? (Probably overkill to do all the permutations.) You'd have to refactor a little to call the token setup again

pkg/vault/test/foundry/E2eSwap.t.sol
```diff
@@ -0,0 +1,705 @@
...
+        vm.prank(admin);
+        feeController.setGlobalProtocolSwapFeePercentage(FIFTY_PERCENT);
+
+        vm.prank(poolCreator);
+        // Set pool creator fee to 100%, so protocol + creator fees equals the total charged fees.
+        feeController.setPoolCreatorSwapFeePercentage(pool, FixedPoint.ONE);
+    }
+
+    /**
+     * @notice Override pool created by BaseVaultTest
```
EndymionJkb, line 81:
```suggestion
     * @notice Override pool created by BaseVaultTest.
```

pkg/vault/test/foundry/E2eSwap.t.sol
```diff
@@ -0,0 +1,705 @@
...
+     * @dev For this test to be generic and support tokens with different decimals, tokenA and tokenB must be set by
+     * _setUpVariables function. If this function runs before `BaseVaultTest.setUp()`, in the `setUp()` function, tokens
```
EndymionJkb, line 83:
```suggestion
     * `_setUpVariables`. If this function runs before `BaseVaultTest.setUp()`, in the `setUp()` function, tokens
```

pkg/vault/test/foundry/E2eSwap.t.sol
```diff
@@ -0,0 +1,705 @@
...
+        tokenB.mint(address(vault), 100 * poolInitAmountTokenB);
+        // Override vault liquidity, to make sure the extra liquidity is registered.
+        vault.manualSetReservesOf(tokenA, 100 * poolInitAmountTokenA);
+        vault.manualSetReservesOf(tokenB, 100 * poolInitAmountTokenB);
+    }
+
+    function testDoExactInUndoExactInNoFees__Fuzz(uint256 exactAmountIn) public {
+        exactAmountIn = bound(exactAmountIn, minSwapAmountTokenA, maxSwapAmountTokenA);
+
+        // Set swap fees to 0 (do not check pool fee percentage limits, some pool types do not accept 0 fees).
```
EndymionJkb, line 143:
```suggestion
        // Set swap fees to 0 (do not check pool fee percentage limits; some pool types do not accept 0 fees).
```

pkg/vault/test/foundry/E2eSwap.t.sol
```diff
@@ -0,0 +1,705 @@
...
+        _checkUserBalancesAndPoolInvariant(balancesBefore, balancesAfter, 0, 0);
+    }
+
+    function testDoExactInUndoExactInLiquidity__Fuzz(uint256 liquidityTokenA, uint256 liquidityTokenB) public {
+        liquidityTokenA = bound(liquidityTokenA, poolInitAmountTokenA / 10, 10 * poolInitAmountTokenA);
+        liquidityTokenB = bound(liquidityTokenB, poolInitAmountTokenB / 10, 10 * poolInitAmountTokenB);
+
+        uint256 exactAmountIn = _setPoolBalancesAndGetAmountIn(liquidityTokenA, liquidityTokenB);
+
+        // Set swap fees to 0 (do not check pool fee percentage limits, some pool types do not accept 0 fees).
```
EndymionJkb, line 184:
```suggestion
        // Set swap fees to 0 (do not check pool fee percentage limits; some pool types do not accept 0 fees).
```

pkg/vault/test/foundry/E2eSwap.t.sol
```diff
@@ -0,0 +1,705 @@
...
+            MAX_UINT128,
+            false,
+            bytes("")
+        );
+
+        uint256 feesTokenB = vault.getAggregateSwapFeeAmount(pool, tokenB);
+
+        // In the first swap, the trade was exactAmountIn => exactAmountOutDo + feesTokenB. So, if
+        // there were no fees, trading `exactAmountOutDo + feesTokenB` would get exactAmountIn. Therefore, a swap
+        // with exact_in `exactAmountOutDo + feesTokenB` is comparable with `exactAmountIn`, given that the fees are
```
EndymionJkb, line 243:
```suggestion
        // with exact_in `exactAmountOutDo + feesTokenB` is comparable to `exactAmountIn`, given that the fees are
```
This is in a couple more places.

pkg/vault/test/foundry/E2eSwap.t.sol
```diff
@@ -0,0 +1,705 @@
...
+            MAX_UINT128,
+            MAX_UINT128,
+            false,
+            bytes("")
+        );
+        vm.stopPrank();
+
+        if (decimalsTokenA != decimalsTokenB) {
+            // If tokens have different decimals, an error is introduced in the computeBalance in the order of the
+            // difference of the decimals.
```
EndymionJkb, line 536:
```suggestion
            // If tokens have different decimals, `computeBalance` will be off by an order of magnitude proportional to the difference in decimals.
```

pkg/vault/test/foundry/E2eSwap.t.sol
```diff
@@ -0,0 +1,705 @@
...
+            }
+            assertApproxEqAbs(
+                exactAmountInSwap,
+                exactAmountIn,
+                tolerance,
+                "ExactOut and ExactIn amountsIn should match"
+            );
+        } else {
+            // Accepts an error of 0.0000001% between amountIn from ExactOut and ExactIn swaps. This error is caused by
+            // differences in computeInGivenExactOut and computeOutGivenExactIn functions of the pool math.
```
EndymionJkb, line 551:
```suggestion
            // differences in the computeInGivenExactOut and computeOutGivenExactIn functions of the pool math.
```
Text also appears elsewhere.

pkg/vault/test/foundry/E2eSwap.t.sol
```diff
@@ -0,0 +1,705 @@
...
+        vm.stopPrank();
+
+        console.log("exactAmountIn    ", exactAmountIn);
+        console.log("exactAmountInSwap", exactAmountInSwap);
+        console.log("feesTokenA       ", feesTokenA);
+        console.log("feesTokenB       ", feesTokenB);
+        console.log("exactAmountOut   ", exactAmountOut);
+        console.log("AmountOut fee    ", exactAmountOut + feesTokenB);
+        console.log("AmountIn + fee   ", exactAmountInSwap + feesTokenA);
+        console.log("AmountIn - fee   ", exactAmountInSwap - feesTokenA);
```
EndymionJkb, line 598:
```suggestion
```
Leftover debug code

pkg/vault/test/foundry/E2eSwap.t.sol
```diff
@@ -0,0 +1,705 @@
...
+            balancesBefore.vaultTokens[tokenBIdx],
+            "Wrong vault tokenB balance"
+        );
+    }
+
+    function _setPoolBalancesAndGetAmountIn(
+        uint256 liquidityTokenA,
+        uint256 liquidityTokenB
+    ) private returns (uint256 amountIn) {
+        // Set liquidity of pool.
```
EndymionJkb, line 667:
```suggestion
        // Set pool liquidity.
```

pkg/vault/test/foundry/E2eSwapSmallDecimalsToBig.t.sol
```diff
@@ -0,0 +1,47 @@
...
+contract E2eSwapSmallDecimalsToBigTest is E2eSwapTest {
+    using ArrayHelpers for *;
+    using FixedPoint for uint256;
+
+    function setUp() public virtual override {
+        E2eSwapTest.setUp();
+    }
+
+    function setUpTokens() internal override {
+        tokenA = new ERC20TestToken("tokenA", "tokenA", 8);
```
EndymionJkb, line 30: Same comment as above about token decimal choices.

EndymionJkb (commented):
Much needed! Have a lot of minor comments; the substantive question I have is about the particular decimals chosen. It might be hard to generalize, but I think comments on the chosen ranges would help in any case. Also just curious about what it does with extreme decimals.

pkg/vault/test/foundry/E2eSwap.t.sol
```diff
@@ -0,0 +1,696 @@
...
+    /// @notice Donate tokens to vault, so liquidity tests are possible.
+    function _donateToVault() internal virtual {
+        tokenA.mint(address(vault), 100 * poolInitAmountTokenA);
+        tokenB.mint(address(vault), 100 * poolInitAmountTokenB);
+        // Override vault liquidity, to make sure the extra liquidity is registered.
+        vault.manualSetReservesOf(tokenA, 100 * poolInitAmountTokenA);
+        vault.manualSetReservesOf(tokenB, 100 * poolInitAmountTokenB);
+    }
+
+    function testDoUndoExactInNoFees__Fuzz(uint256 exactAmountIn) public {
```
jubeira, line 140: I'm wondering whether we should test this case or not.

If a pool fails because the fee is 0 but it has a guardrail it should still be ok for the most part (unless LP fees are bypassed with aggregate swap fees completely). What if a pool type fails in this case?

pkg/vault/test/foundry/E2eSwap.t.sol
```diff
@@ -0,0 +1,696 @@
...
+
+        decimalsTokenA = IERC20Metadata(address(tokenA)).decimals();
+        decimalsTokenB = IERC20Metadata(address(tokenB)).decimals();
+        poolInitAmountTokenA = poolInitAmount.mulDown(10 ** decimalsTokenA);
+        poolInitAmountTokenB = poolInitAmount.mulDown(10 ** decimalsTokenB);
+
+        setUpVariables();
+        createAndInitCustomPool();
+
+        _donateToVault();
```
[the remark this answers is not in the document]
jubeira: ah we're setting the balances manually. 

For the purposes of this test I think this is OK; it shouldn't affect the outcome.
Was this just a shortcut, or is it that the invariant ratio for add liquidity unbalanced makes setting the initial conditions harder?

pkg/vault/test/foundry/E2eSwap.t.sol
```diff
@@ -0,0 +1,696 @@
...
+        vm.stopPrank();
+
+        BaseVaultTest.Balances memory balancesAfter = getBalances(sender);
+
+        assertLe(exactAmountOutUndo, exactAmountIn, "Amount out undo should be <= exactAmountIn");
+
+        _checkUserBalancesAndPoolInvariant(balancesBefore, balancesAfter, 0, 0);
+    }
+
+    function testDoUndoExactInLiquidity__Fuzz(uint256 liquidityTokenA, uint256 liquidityTokenB) public {
```
jubeira, line 178: I'd suggest a modification over this one: 
- Have an internal function that receives initial liquidity and swap sizes
- Fuzz initial liquidity, keep swap size the same (i.e. what this test does now)
- Keep initial liquidity fixed, fuzz swap size
- Fuzz all

(this can be done in a follow-up PR; you might need to add a few more bounds here and there).

pkg/pool-weighted/test/foundry/E2eSwap.t.sol
```diff
@@ -0,0 +1,309 @@
...
+        );
+        uint256 feesTokenA = vault.getAggregateSwapFeeAmount(pool, tokenA);
+        vm.stopPrank();
+
+        BaseVaultTest.Balances memory balancesAfter = getBalances(sender);
+        // `BaseVaultTest.getBalances` checks the poolInvariant of `pool`. In this test, a different pool is used, so
+        // poolInvariant should be overwritten.
+        balancesAfter.poolInvariant = _calculatePoolInvariant(address(poolWithMutableWeights));
+
+        assertLe(exactAmountOutUndo, exactAmountIn - feesTokenA, "Amount out undo should be <= exactAmountIn");
```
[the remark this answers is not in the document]
jubeira: Also, this is a bit stricter than what I had in mind: from the user's perspective, to check if you're getting any value out of the operation checking `exactAmountIn` against `exactAmountOutUndo` should be enough. But stricter is better :+1:

pkg/vault/test/foundry/E2eSwap.t.sol
```diff
@@ -0,0 +1,696 @@
...
+
+        _checkUserBalancesAndPoolInvariant(balancesBefore, balancesAfter, feesTokenA, feesTokenB);
+    }
+
+    function testDoUndoExactInVariableFeesAmountInAndLiquidity__Fuzz(
+        uint256 exactAmountIn,
+        uint256 poolSwapFeePercentage,
+        uint256 liquidityTokenA,
+        uint256 liquidityTokenB
+    ) public {
```
jubeira, line 270: Ah nice, so this one is fuzzing all at once!

I wonder if we could have a single internal function like this, and then have multiple external ones that fuzz piece by piece (can be done in another PR too)

pkg/vault/test/foundry/E2eSwap.t.sol
```diff
@@ -0,0 +1,696 @@
...
+                exactAmountInSwap,
+                exactAmountIn,
+                tolerance,
+                "ExactOut and ExactIn amountsIn should match"
+            );
+        } else {
+            // Accepts an error of 0.0000001% between amountIn from ExactOut and ExactIn swaps. This error is caused by
+            // differences in the computeInGivenExactOut and computeOutGivenExactIn functions of the pool math.
+            assertApproxEqRel(exactAmountInSwap, exactAmountIn, 1e9, "ExactOut and ExactIn amountsIn should match");
+        }
```
jubeira, line 553: :hushed: oohh nice

pkg/vault/test/foundry/E2eSwap.t.sol
```diff
@@ -0,0 +1,696 @@
...
+                exactAmountIn,
+                1e9,
+                "ExactOut and ExactIn amountsIn should match"
+            );
+        }
+    }
+
+    function _checkUserBalancesAndPoolInvariant(
+        BaseVaultTest.Balances memory balancesBefore,
+        BaseVaultTest.Balances memory balancesAfter,
```
jubeira, line 620: One more to add could be live balances.
WDYT? Might be an overkill to test that here though. Not a strong opinion.

jubeira: actually, might make sense to run the existing test against tokens that are yield bearing. But that's another PR.

jubeira (approved):
Awesome job really :100: 

I've just got a few minor comments and thoughts, but overall this looks great. This also gave me some more ideas to keep testing, but most of those can be addressed in follow-up PRs. We can sync about those potential ideas if needed :)

pkg/vault/test/foundry/E2eSwap.t.sol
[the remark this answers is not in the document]
joaobrunoah: It was a shortcut, just a way to change pool balances without initializing the pool again. I just change the pool balances since the vault already have reserves for the token.

pkg/vault/test/foundry/E2eSwap.t.sol
joaobrunoah: Removed it!

pkg/vault/test/foundry/E2eSwap.t.sol
joaobrunoah: I think, with the latest version, this was addressed.

pkg/vault/test/foundry/E2eSwap.t.sol
joaobrunoah: Done!

pkg/vault/test/foundry/E2eSwap.t.sol
joaobrunoah: I'll create a test with Yield-Bearing tokens in another PR.

PR #803, opened 2024-07-23
Minimize storage reads in WeightedPool::computeBalance

gerrrg:
# Description
Currently, calling `computeBalance()` can perform (at most) 8 storage reads since it's calling `_getNormalizedWeights()` and then taking the `i`th index. Instead, we can directly query `_getNormalizedWeight(i)` and just perform a single storage read for the weight we want.

## Type of change

- [ ] Bug fix 
- [ ] New feature 
- [ ] Breaking change 
- [ ] Dependency changes
- [ ] Code refactor / cleanup
- [x] Optimization: [x] gas / [ ] bytecode
- [ ] Other

## Checklist:


EndymionJkb:
Hmmm! Good catch; certainly seems like this should work. I don't believe we ever call this, which is one reason we never noticed. Should have a test. (We would have found that in the coverage tests I suppose.)

jubeira (approved):
Good catch! Thanks @gerrrg :)

This should positively impact unbalanced liquidity operations (although weights are immutables, so it's not as expensive as it'd be if we were reading them from storage).

gerrrg:
Ah, yeah good point @jubeira about immutable reads vs normal storage reads. Still seems like a good thing to add in, just not as significant as I initially thoughts 🙂

PR #818, opened 2024-07-28
Missing token indices

EndymionJkb:
# Description

There are still some tests that work accidentally, without specifying token order. Sometimes it's ok - there are tests that set parallel token arrays for tokens and rate providers, where 0 = dai and 1 = usdc, but those arrays are then passed into functions that sort them.

There are also plenty of cases that are going to revert anyway, so using numeric indices is ok. Or both entries have the same value (or token type, etc.), so it likewise doesn't matter what order they're in. Those are fine and unchanged.

Even where it's not strictly necessary, it's good documentation to use indices. (And often it is necessary.)

## Type of change

- [ ] Bug fix 
- [ ] New feature 
- [ ] Breaking change 
- [ ] Dependency changes
- [X] Code refactor / cleanup
- [ ] Other

## Checklist:


joaobrunoah (approved):
LGTM

PR #820, opened 2024-07-29
Remove ERC4626 wrap/unwrap verification from critical path

joaobrunoah:
# Description

Verifying if an ERC4626 wrapper returned the correct amount of tokens is an expensive task in terms of gas and bytecode, and is only effective in the scenario in which the buffer has liquidity, but not enough to do the wrap/unwrap operation (situation in which we need to compare the convert and actual operations). This PR removes this verification, which removes 700 bytes from the Vault code.

A fuzz test of ERC4626 wrappers will compensate for the lack of this verification.

## Type of change

- [ ] Bug fix 
- [ ] New feature 
- [ ] Breaking change 
- [ ] Dependency changes
- [x] Code refactor / cleanup
- [ ] Other

## Checklist:


pkg/vault/contracts/Vault.sol
```diff
@@ -1172,11 +1172,8 @@ contract Vault is IVaultMain, VaultCommon, Proxy {
                 // The mint operation returns exactly `vaultWrappedDelta` shares. To do so, it withdraws underlying
                 // from the vault and returns the shares. So, the vault needs to approve the transfer of underlying
                 // tokens to the wrapper.
-                // Add convert error because mint can consume a different amount of tokens than we anticipated.
-                underlyingToken.forceApprove(
-                    address(wrappedToken),
-                    _addConvertError(amountInUnderlying + bufferUnderlyingSurplus)
-                );
+                // Add 2 because mint can consume up to 2 wei more tokens than we anticipated.
+                underlyingToken.forceApprove(address(wrappedToken), amountInUnderlying + bufferUnderlyingSurplus + 2);
```
jubeira, line 1176: I've been thinking about this. I think we'll need to keep `forceApprove`, which prevents 2 issues:
- ABI incompatibility (return value)
- Token requiring spender allowance to be 0 when a value different than 0 is set

The second condition shouldn't have an issue with regular `approve`, since we're zeroing out the allowance below. But unfortunately the code will revert because of the first issue if we don't use this.

This way of doing it (approve 'exact' quantity needed + use allowance + zero out allowance) seems to be _very_ correct. I thought about a few optimizations:
- Avoid the sum (`amountInUnderlying + bufferUnderlyingSurplus + 2`); just give it a huge allowance since you're going to zero it out anyways afterwards. The problem with this is that the wrapper could take more underlying than what we were calculating, but that should revert since we measure how much it took, and if outside the requested limit the operation will revert right?
  - alt: make sum unchecked :face_with_spiral_eyes: this will very rarely overflow, and if it does we'll be granting less allowance than required, which will revert anyways
- Just leave the leftover allowance; in any case it should be small right? The problem here is that if the wrapper takes less tokens during the operation and the leftover allowance is significant, the wrapper can then drain those tokens without any sort of accounting.
- Minor, but here it goes: each `forceApprove` actually does 2 calls to the token: one with 0 allowance, and one with the desired allowance. Then, these two `forceApprove` actually are calling:
```
token.approve(wrapper, 0);
token.approve(wrapper,  amountInUnderlying + bufferUnderlyingSurplus + 2);
// mint
token.approve(wrapper, 0);
token.approve(wrapper,  0);
```

If we make a small modification to `SafeERC20` we can avoid the last call; we don't need to zero out the allowance twice.


Let's just think about this for now. No action needed in this PR.

pkg/vault/test/foundry/fork/ERC4626MainnetAaveDai.t.sol
jubeira, line 1: Can we have the dreaded USDT too? :innocent:

pkg/vault/test/foundry/fork/ERC4626WrapperBase.t.sol
```diff
@@ -0,0 +1,213 @@
...
+    IERC20 internal underlyingToken;
+    uint256 internal underlyingToWrappedFactor;
+
+    address internal user;
+    uint256 internal userInitialUnderlying;
+    uint256 internal userInitialShares;
+
+    uint256 internal constant MIN_DEPOSIT = 100;
+    // Tolerance of 1 wei difference between convert/preview and actual operation.
+    uint256 internal constant TOLERANCE = 1;
```
jubeira, line 28: Question: Why is the tolerance 1 wei here, but we use 2 wei in the mint path?

jubeira (approved):
LGTM, great job @joaobrunoah !

Just a few minor comments / questions. Let's leave potential gas optimizations around approval for another PR anyways. 

This is a good call; the savings are very significant, and the new tests make sense. Thanks for pushing this!

pkg/vault/test/foundry/fork/ERC4626WrapperBase.t.sol
joaobrunoah: Just because the mint path is in the vault, so I was afraid of being too restrictive, but 1 would work for all wrappers that we've tested so far.

pkg/vault/test/foundry/fork/ERC4626MainnetAaveDai.t.sol
joaobrunoah: Done! I also created a fuzz test for an yield-bearing pool with USDT.

EndymionJkb (approved):
Love all that red, and glad to be below the limit again! The deleted code is a good reference for exactly how all this works. I'd say we should document this internally - at least a paragraph in the "decision log" about the conclusion, and a reference  to the PR

pkg/vault/test/foundry/fork/ERC4626MainnetAaveDai.t.sol
EndymionJkb: We should consider modifying ERC20TestToken so that it behaves like USDT - then we'd be testing everything against it, just in case there's somewhere else that would break.

pkg/vault/contracts/Vault.sol
joaobrunoah: I tried replacing the sum by a big int value, but the code increased by 7 bytes, and there were no changes in terms of gas, so I'm leaving it as is for now.

joaobrunoah:
@EndymionJkb https://www.notion.so/Decision-Log-7257a5ae01e44e4cb8c48f65937f2a37?pvs=4#568014ea070d4eaf9c8b444263303229

PR #824, opened 2024-07-30
Aggregate fee tests

### LINKED ISSUE #823: Adjust `VaultLiquidityFees` test
It's still using protocol / creator fees, which doesn't apply anymore.
Use single aggregate fee percentage.

EndymionJkb:
# Description

We had a legacy helper in `BaseVaultTest` (and a duplicate in one of the test files) that computed aggregate percentages from their components. It had (documented) precision issues, which were resolved in the latest `ProtocolFeeController` code.

This makes minimal changes to remove these helpers and use the `ProtocolFeeController` itself for aggregate computation, so that it's consistent with what the system does. For the fuzz tests, there were still precision issues trying to set each to its full range, so in that case I had to fuzz the aggregate fee directly (bounding it, then manually adjusting for precision).

A more extensive change would be to always use the aggregate percentage natively, and not try to compute it.

## Type of change

- [ ] Bug fix 
- [ ] New feature 
- [ ] Breaking change 
- [ ] Dependency changes
- [X] Code refactor / cleanup
- [ ] Other

## Checklist:

## Issue Resolution

Resolves #823 .

### DIFF
--- a/pkg/vault/test/foundry/unit/VaultUnitSwap.t.sol
+++ b/pkg/vault/test/foundry/unit/VaultUnitSwap.t.sol
@@ -64,7 +64,7 @@ contract VaultUnitSwapTest is BaseTest {
         uint256 limitRaw = 1000e18;
         uint256 swapFeePercentage = 1e16;
         uint256 protocolFeePercentage = 20e16;
-        uint256 poolCreatorFeePercentage = 5e17;
+        uint256 poolCreatorFeePercentage = 50e16;
 
         (, SwapState memory state, PoolData memory poolData) = _makeParams(
             SwapKind.EXACT_IN,
--- a/pkg/vault/test/foundry/unit/VaultUnitSwap.t.sol
+++ b/pkg/vault/test/foundry/unit/VaultUnitSwap.t.sol
@@ -6,6 +6,7 @@ import "forge-std/Test.sol";
 
 import { IERC20 } from "@openzeppelin/contracts/token/ERC20/IERC20.sol";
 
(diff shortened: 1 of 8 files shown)


pkg/vault/test/foundry/unit/VaultUnitSwap.t.sol
```diff
@@ -43,9 +44,12 @@ contract VaultUnitSwapTest is BaseTest {
     uint256[] decimalScalingFactors = [uint256(1e18), 1e18];
     uint256[] tokenRates = [uint256(1e18), 2e18];
 
+    IProtocolFeeController feeController;
+
     function setUp() public virtual override {
         BaseTest.setUp();
         vault = IVaultMock(address(VaultMockDeployer.deploy()));
+        feeController = vault.getProtocolFeeController();
```
jubeira, line 52: Maybe we can put this in the base layer, and make `_getAggregateFeePercentage` use it?

jubeira: (or just keep the code as is now, but store `feeController` reference in the base test to reduce setup in every test)

jubeira (approved):
LGTM, thanks @EndymionJkb !

How about storing the `feeController` in the base test? That would reduce a few setup lines here and there; it could be useful to have.

pkg/vault/test/foundry/unit/VaultUnitSwap.t.sol
EndymionJkb: Moved it up to the BaseVaultTest

PR #826, opened 2024-07-30
Use forceApprove for arbitrary token fee collection

EndymionJkb:
# Description

Inspired to check our own code to ensure we were always using `forceApprove` where necessary, and it seems like we missed one. (Probably redundant in this case, as the approval would almost certainly be 0 to start with, but doesn't hurt.)

Good thing this is the VaultAdmin - seems to have added 500 bytes?!

## Type of change

- [X] Bug fix 
- [ ] New feature 
- [ ] Breaking change 
- [ ] Dependency changes
- [ ] Code refactor / cleanup
- [ ] Other

## Checklist:


jubeira (approved):
LGTM, it is what is is.

PR #827, opened 2024-07-30
Use standard dai/usdc for e2e

EndymionJkb:
# Description

This uses token1/2 and has a custom sort. I wondered if there was a reason to do this vs. just using dai/usdc and the common sort like other tests, and there doesn't seem to be.

Merge into #801 if I'm not missing something.

## Type of change

- [ ] Bug fix 
- [ ] New feature 
- [ ] Breaking change 
- [ ] Dependency changes
- [X] Code refactor / cleanup
- [ ] Other

## Checklist:


joaobrunoah:
Hey @EndymionJkb ! Since the idea is that this is a base test for other tests, and could be extended to fork tests too with small adjustments, I thought it'd be better to don't use dai/usdc directly here, and use variables for the tokens instead. But I'll use tokenA/tokenB instead of the numbers, as you suggested in the other PR.

EndymionJkb:
Primary PR will be modified as described above.

PR #829, opened 2024-07-30
Generalizing erc4626 buffer tests

### LINKED ISSUE #808: Create generalizing tests from ERC4626 buffer tests
https://github.com/balancer/balancer-v3-monorepo/pull/792#discussion_r1686929189

elshan-eth:
# Description

The initialization of erc4626 buffer tests was repeated in different tests. This PR adds a generalizing erc4626 contract for using common initialization. 

## Type of change

- [ ] Bug fix 
- [ ] New feature 
- [ ] Breaking change 
- [ ] Dependency changes
- [ ] Code refactor / cleanup
- [X] Other

## Checklist:

## Issue Resolution

Closes #808


joaobrunoah (approved):
Nice, looks good to me.
I'm still confused if we're gonna call the pool that holds yield-bearing tokens a `yield-bearing pools`, `boosted pool` or `erc4626 pool`. I think we should decide on that to keep the standard. @EndymionJkb @jubeira , which one should we adopt (or have we adopted `yield-bearing pool` already)?

EndymionJkb:
> Nice, looks good to me. I'm still confused if we're gonna call the pool that holds yield-bearing tokens a `yield-bearing pools`, `boosted pool` or `erc4626 pool`. I think we should decide on that to keep the standard. @EndymionJkb @jubeira , which one should we adopt (or have we adopted `yield-bearing pool` already)?

We don't want "marketing" terms in the code, so definitely not "boosted" pool. We have been using "erc4626" for the tokens (as this is explicit in the code), and "yield-bearing" for the pool, as ERC4626 tokens are yield-bearing by definition, so I'd stick with that.

pkg/vault/test/foundry/QueryERC4626Buffer.t.sol
```diff
@@ -7,100 +7,77 @@ import "forge-std/Test.sol";
...
-
-        // "USDC" is deliberately 18 decimals to test one thing at a time.
-        waUSDC = new ERC4626TestToken(usdc, "Wrapped aUSDC", "waUSDC", 18);
-        vm.label(address(waUSDC), "waUSDC");
-
-        (waDaiIdx, waUsdcIdx) = getSortedIndexes(address(waDAI), address(waUSDC));
-
-        _initializeBuffers();
-        _initializeYieldBearingPool();
+        bufferInitialAmount = erc4626PoolInitialAmount / 100;
```
jubeira, line 26: We're using `bufferInitialAmount` before this line to set the `swapAmount`. Shouldn't we be updating `swapAmount` after?

pkg/vault/test/foundry/YieldBearingPoolBufferAsVaultPrimitive.t.sol
```diff
@@ -17,189 +16,125 @@ import { IRateProvider } from "@balancer-labs/v3-interfaces/contracts/vault/IRat
...
-
-        // "USDC" is deliberately 18 decimals to test one thing at a time.
-        waUSDC = new ERC4626TestToken(usdc, "Wrapped aUSDC", "waUSDC", 18);
-        vm.label(address(waUSDC), "waUSDC");
-
-        (waDaiIdx, waUsdcIdx) = getSortedIndexes(address(waDAI), address(waUSDC));
-
-        initializeBuffers();
-        initializeYieldBearingPool();
+        bufferInitialAmount = erc4626PoolInitialAmount / 100;
```
jubeira, line 33: Same; we're already using this to set `swapAmount` and `unbalanceDelta`. Is that correct?

jubeira (commented):
Great job @elshan-eth, looks much cleaner! I just have one doubt regarding one of the initial variables, otherwise LGTM. Could you please take a look?

pkg/vault/test/foundry/QueryERC4626Buffer.t.sol
elshan-eth: Yes, you are right

PR #831, opened 2024-07-31
Add VaultExplorer support for `getAggregateFeePercentages`

EndymionJkb:
# Description

We added a new function to `IPoolInfo`, so not technically "the Vault," but it's a permissionless getter for pool info, so I think it should be supported here as well. Also works for every pool, even custom ones that don't implement `IPoolInfo`, so it's an equivalent/safer path for offchain processes who don't want to keep track of which pools support `IPoolInfo`.

## Type of change

- [ ] Bug fix 
- [X] New feature 
- [ ] Breaking change 
- [ ] Dependency changes
- [ ] Code refactor / cleanup
- [ ] Other

## Checklist:

### DIFF
--- a/pkg/interfaces/contracts/vault/IVaultExplorer.sol
+++ b/pkg/interfaces/contracts/vault/IVaultExplorer.sol
@@ -358,6 +358,23 @@ interface IVaultExplorer {
     *******************************************************************************/
 
     /**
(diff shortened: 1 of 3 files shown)


jubeira (approved):
LGTM, makes sense!

PR #833, opened 2024-08-01
Static analysis update

EndymionJkb:
# Description

Running slither over the current codebase - we didn't find much!

First, there was apparently an issue with the current version (0.10.1) and transient storage: it doesn't run at all. Upgrading to 0.10.3, it still shows the same errors, but at least doesn't crash.

There are hundreds of "findings" that are ok, such as: 1) it sees reentrancy in functions that are actually protected by nonReentrant; 2) it says we're "ignoring" return values when we’re returning them to an outer function (e.g., all of VaultExplorer); 3) it doesn’t understand the Proxy pattern, and says a lot of stuff isn't used when it is; 4) it complains about low-level calls, using assembly, 0.8.20 in OZ vs 0.8.24 in our codebase; 5) various things in imported OZ libraries and other places we can't/won't change; 6) such as LogExpMath; 7) it complains about cyclomatic complexity in the BatchRouter (not much we can do without breaking up the path functions; we made that trade-off for readability), and so on...

One interesting finding was that we never use the return value of the Vault's settle (except in tests). Leaving that in for now. We also don't use the 3-argument `_canPerform` in BasePoolAuthentication: but something; that's a reusable base contract.

There were still a few unused imports, and it noted that BalancerPoolToken was really an `IRateProvider` (I think there were nested inheritance issues at one time, but no longer).

## Type of change

- [ ] Bug fix 
- [ ] New feature 
- [ ] Breaking change 
- [ ] Dependency changes
- [X] Code refactor / cleanup
- [ ] Other

## Checklist:


jubeira (approved):
LGTM! (trusting omissions in the DB)

EndymionJkb:
> LGTM! (trusting omissions in the DB)

They're noted in the description at the top. It also doesn't recognize @inheritdoc, so there were some "unused imports" that were used for docs. I've updated it again after all the recent merges.

PR #835, opened 2024-08-02
Improve precision of StableMath's computeBalance

joaobrunoah:
# Description

When testing ExactIn and ExactOut swaps, the tests got different results for equivalent operations. That was because computeBalance was introducing rounding issues to the inverse invariant calculation.

This PR improves the rounding of computeBalance.

## Type of change

- [ ] Bug fix 
- [ ] New feature 
- [ ] Breaking change 
- [ ] Dependency changes
- [x] Code refactor / cleanup
- [ ] Other

## Checklist:


pkg/pool-hooks/foundry.toml
```diff
@@ -1 +0,0 @@
-../../foundry.toml
```
joaobrunoah, line 1: Changing this because CI is failing

pkg/solidity-utils/test/foundry/StableMath.t.sol
```diff
@@ -19,7 +19,8 @@ contract StableMathTest is Test {
     uint256 constant NUM_TOKENS = 4;
 
     uint256 constant MIN_BALANCE_BASE = 1e18;
-    uint256 constant MAX_BALANCE_BASE = 1000e18;
+    // Max balance of each token of a stable pool to don't overflow the stable math.
```
EndymionJkb, line 22:
```suggestion
    // Max balance of a stable pool token that won't cause an overflow in stable math.
```

EndymionJkb (approved):
I have StableMath PTSD from v2 (found a precision issue, due to computation order, that reduced the range of invariant convergence)... but this does look ok. It *increases* precision, and the 1e3 error observed makes perfect sense (matching the AMP_PRECISION).

Also, we've got much better verification tools now, and that range certainly looks sufficient. I would be ~$1.5 trillion of Shiba Inu :)

PR #840, opened 2024-08-04
Polish: storage docs

EndymionJkb:
# Description

Guess I missed this during the Vault polish. Some comments were out-of-date, and things were fairly jumbled around. This separates things and adds block comments to make everything more readable.

The diff isn't very legible. Best to just view the current file; there are no code changes.

Interestingly, just moving things around seemed to increase bytecode and decrease gas. This makes some sense, because it does lay the storage out in order. Maybe there is an optimal order for variables to maximize gas, even though it would be a small effect.

If it's large enough to worry about, we could make this the VaultStorageDoc interface for readability, but not use it. It would have a comment pointing to the optimized VaultStorage, which would just declare everything in optimal order with no comments.

## Type of change

- [ ] Bug fix 
- [ ] New feature 
- [ ] Breaking change 
- [ ] Dependency changes
- [ ] Code refactor / cleanup
- [ ] Other

## Checklist:


pkg/vault/test/.contract-sizes/Vault
```diff
@@ -1,2 +1,2 @@
-Bytecode	23.816
-InitCode	25.092
\ No newline at end of file
+Bytecode	23.943
```
jubeira, line 1: ooof ~120 bytes for 100 gas? Are we sure this is from this PR?

I was hoping to get some bytecode buffer to play a bit with the compiler settings; feels like we can get better gas improvements out of that.

We also have some external getters in the extension that would be better off in the main vault for the batch router.

jubeira (commented):
LGTM; I'm a bit concerned about the bytecode size increase. Are we sure it's from this PR?

pkg/vault/test/.contract-sizes/Vault
EndymionJkb: It seems to be from this PR, as I can influence it by changing the order. Experimenting now with shifting things around.

EndymionJkb: It really doesn't like moving the contract addresses

jubeira (approved):
LGTM, thanks for investigating this! :)

PR #844, opened 2024-08-05
Fix query add liquidity proportional parameters

jubeira:
# Description

Max amounts in is not needed for the query.

## Type of change

- [x] Bug fix 
- [ ] New feature 
- [ ] Breaking change 
- [ ] Dependency changes
- [ ] Code refactor / cleanup
- [ ] Other

## Checklist:

## Issue Resolution

N/A

### DIFF
--- a/pkg/interfaces/contracts/vault/IRouter.sol
+++ b/pkg/interfaces/contracts/vault/IRouter.sol
@@ -390,14 +390,12 @@ interface IRouter {
     /**
      * @notice Queries an `addLiquidityProportional` operation without actually executing it.
      * @param pool Address of the liquidity pool
-     * @param maxAmountsIn Maximum amounts of tokens to be added, sorted in token registration order
      * @param exactBptAmountOut Exact amount of pool tokens to be received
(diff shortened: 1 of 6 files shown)


EndymionJkb (commented):
Didn't we decide to keep these parallel, even though the limits parameters aren't actually used? (Could keep the argument, but not use it, I guess.)

If we're removing arguments, what about the same parameter in `queryAddLiquidityCustom`? And `maxBptAmountIn` for `queryRemoveLiquidityCustom`?

jubeira:
> Didn't we decide to keep these parallel, even though the limits parameters aren't actually used? (Could keep the argument, but not use it, I guess.)

All the other queries don't have the limit on the calculated amount alrady.

> If we're removing arguments, what about the same parameter in queryAddLiquidityCustom? And maxBptAmountIn for queryRemoveLiquidityCustom?

In `custom`, you don't really know which end (in / out) is the limit, so you need to keep both.

EndymionJkb (approved):
Thanks for the explanation! Adds a few bytes, but luckily still fits!

PR #846, opened 2024-08-06
Extreme do/undo amounts tests

### LINKED ISSUE #359: Test: extreme swap with infinite BPT in composable pool
Suppose you have a version of `bb-a-usd` (pool A: boosted dai / usdc / usdt), and pair that against `wstETH` (pool B: bb-a-usd / wstETH).

Technically infinite `bb-a-usd` can be minted by adding liquidity, and be used to trade against `wstETH` in pool B. We need to make sure that this doesn't break the weighted pool B.

elshan-eth:
# Description

This test checks that the logic is not broken when `do/undo` tests with super large amounts.

## Type of change

- [ ] Bug fix 
- [ ] New feature 
- [ ] Breaking change 
- [ ] Dependency changes
- [ ] Code refactor / cleanup
- [X] Other

## Checklist:

## Issue Resolution

Closes #359


pkg/vault/test/foundry/ExtremeSwapWithInfiniteBPT.t.sol
```diff
@@ -0,0 +1,163 @@
...
+            [poolInitialAmount, poolInitialAmount].toMemoryArray(),
+            poolInitialAmount * 2 - MIN_BPT
+        );
+
+        vm.stopPrank();
+
+        vm.startPrank(alice);
+        IERC20(address(poolA)).approve(address(router), type(uint256).max);
+        IERC20(address(poolA)).approve(address(permit2), type(uint256).max);
+        permit2.approve(address(poolA), address(router), type(uint160).max, type(uint48).max);
```
jubeira, line 65: I'm thinking, this could potentially be a helper in the base vault test right? It'a very common pattern in most tests that involve pools with custom tokens.

pkg/vault/test/foundry/ExtremeSwapWithInfiniteBPT.t.sol
```diff
@@ -0,0 +1,163 @@
...
+
+    function testAddLiquidityAndSwapSingleTokenExactIn() public {
+        vm.startPrank(alice);
+
+        (, , uint256[] memory beforeBalancesRaw, ) = vault.getPoolTokenInfo(address(poolB));
+
+        uint256 exactBptAmount = poolInitialAmount;
+        uint256[] memory maxAmountsIn = [MAX_UINT128, MAX_UINT128].toMemoryArray();
+
+        router.addLiquidityProportional(address(poolA), maxAmountsIn, exactBptAmount, false, bytes(""));
```
jubeira, line 81: Note that `addLiquidityProportional` is exact out, so you'll be adding liquidity "up to max uint128", but only until you min the specified exact BPT out, which in this case is pretty low.

To actually mint more, you'd need to specify a huge number in `exactBptOut`.

jubeira (commented):
Good start; thanks for pushing this @elshan-eth !

I'd suggest some changes though. This test is more about pushing things to the edge attempting to break something in the process.

It is true that in testing environments we can just mint more initial funds, but in practice you can also trigger huge swaps by adding liquidity, using BPT in a swap, and then try to revert the trade. The steps would be something like this:
- Add huge liquidity amounts to pool A, get BPT A
- Swap BPT A for another token in composable pool B (in this test, that'd be WETH)
- Then, try to undo: swap WETH for BPT out, and exit

The attack vector here would be trying to exploit uncapped swap sizes back and forth in pool B. The concern is that you don't even need real reserves or an external flashloan, since calling `addLiquidity` is enough to get you BPT even without paying your debts. But to do this, you need to call the vault directly; that's not something you can achieve with the standard router.

The expectation is that either:
- pools will do their job, and doing / undoing big trades will just not be profitable. More or less what #801 attempts to prove, but that PR is not focused on particularly large trades.
- pools will revert if they can't handle big trades

Pool B in this test can also be weighted or stable (or any other potential pool type). Ideally we'd test this against real pool types, not just mocks.

pkg/vault/test/foundry/LinearBaseExtremeAmounts.t.sol
```diff
@@ -0,0 +1,28 @@
...
+
+    function setUp() public virtual override {
+        BaseExtremeAmountsTest.setUp();
+    }
+
+    function _createPool(address[] memory tokens, string memory label) internal override returns (address) {
+        address newPool = address(new PoolMock(IVault(address(vault)), "ERC20 Pool - DAI/USDC", "ERC20_POOL_DAI_USDC"));
+        vm.label(newPool, label);
+
+        factoryMock.registerTestPool(newPool, vault.buildTokenConfig(tokens.asIERC20()), poolHooksContract, lp);
```
EndymionJkb, line 24:
```suggestion
        factoryMock.registerTestPool(newPool, vault.buildTokenConfig(tokens.asIERC20()), address(0), lp);
```
None of the pools seem to use hooks.

pkg/vault/test/foundry/utils/BaseExtremeAmountsTest.sol
```diff
@@ -0,0 +1,391 @@
...
+            })
+        );
+
+        assertEq(amountInReturn, amountOut, "amountInReturn should be equal to amountOut");
+        assertLe(amountOutReturn, amountIn, "amountOutReturn should be less or equal to amountIn");
+    }
+    //#endregion
+
+    //#region Internal functions
+    function _removeSelectorFromErrorReason(
```
EndymionJkb, line 377:
```suggestion
    function _extractSelectorFromErrorReason(
```
Consider this name; you're parsing out and returning pieces

EndymionJkb (commented):
First pass, as it's still a draft (has debug code, etc.) I'm getting a test failure (see below). Numerical examples would help with understanding the intent.

`[FAIL. Reason: Total amount out should be less than amount in: 153127065114322308282125714582882500795 >= 153127065114322308075208245550570587690; counterexample: calldata=0x25735054000000000000000000000000000000000000000000000000000000000000583a00000000000000000000000000000000000000000000000000000000000044d4 args=[[22586 [2.258e4], 17620 [1.762e4]]]] testAddUnbalancedAndRemoveLiquidityProportional_Fuzz(uint256[2]) (runs: 3, μ: 424959, ~: 425202)`

pkg/vault/test/foundry/utils/BaseExtremeAmountsTest.sol
```diff
@@ -0,0 +1,546 @@
...
+        uint256[] memory maxAmountsIn = new uint256[](2);
+        for (uint256 i = 0; i < maxAmountsIn.length; i++) {
+            uint256 maxAmount = initBalances[i] * 100;
+
+            // if invariant ratio is really high, then we don't need to receive an exception
+            unchecked {
+                maxAmountsIn[i] = ((initBalances[i] * maxInvariantRatio) / 1e18);
+            }
+
+            if (maxAmountsIn[i] > maxAmount || maxAmountsIn[i] < initBalances[i]) {
```
jubeira, line 194: Can we add a small comment explaining why we need both conditions?

pkg/vault/test/foundry/utils/BaseExtremeAmountsTest.sol
```diff
@@ -0,0 +1,546 @@
...
+                userData: bytes("")
+            })
+        );
+        assertEq(bptAmountIn, bptAmountOut, "bptAmountIn should be equal to bptAmountOut");
+
+        // if amountsOut[0] more then amountsIn[0], then amountsOut[1] should be less than amountsIn[1] otherwise user gets free tokens
+        // if amountsOut[0] equal to amountsIn[0], then amountsOut[1] should be less or equal to amountsIn[1] otherwise user gets free tokens
+        // if amountsOut[0] less then amountsIn[0], we can't check anything because user doesn't get free tokens
+        if (amountsOut[0] > amountsIn[0]) {
+            assertLt(amountsOut[1], amountsIn[1], "amountsOut[1] should be less than amountsIn[1]");
```
jubeira, line 236: This case doesn't mean that no value was extracted right?

In other words, if the assertion fails for sure there's something wrong, but if it doesn't fail there's still some room for something to be wrong.

I think this is fine for a first iteration. Let's leave a comment; we can check the invariants afterwards.

pkg/vault/test/foundry/utils/BaseExtremeAmountsTest.sol
```diff
@@ -0,0 +1,546 @@
...
+                to: address(this),
+                maxAmountsIn: [maxAdditionalAmountsIn[0], maxAdditionalAmountsIn[1]].toMemoryArray(),
+                minBptAmountOut: exactBPTAmount,
+                kind: AddLiquidityKind.PROPORTIONAL,
+                userData: bytes("")
+            })
+        );
+        assertEq(bptAmountOut, exactBPTAmount, "bptAmountOut should be equal to exactBPTAmount");
+
+        uint removePerOperation = exactBPTAmount / 2;
```
jubeira, line 285:
```suggestion
        uint256 removePerOperation = exactBPTAmount / 2;
```
nit

pkg/vault/test/foundry/utils/BaseExtremeAmountsTest.sol
```diff
@@ -0,0 +1,546 @@
...
+
+        // first single token exact in removal
```
jubeira, line 287: let's just clarify that we can't remove everything with a single token because there's not enough balance to do so (is that right?)

pkg/vault/test/foundry/utils/BaseExtremeAmountsTest.sol
```diff
@@ -0,0 +1,546 @@
...
+            assertEq(amountsOut[1], amountsIn[1], "amountsOut[1] should be equal to amountsIn[1]");
+        } catch (bytes memory reason) {
+            (bytes4 selector, bytes memory reasonWithoutSelector) = _extractSelectorFromErrorReason(reason);
+            (, uint256 balance, uint256 needed) = abi.decode(reasonWithoutSelector, (address, uint256, uint256));
+
+            assertEq(
+                selector,
+                IERC20Errors.ERC20InsufficientBalance.selector,
+                "Selector should be ERC20InsufficientBalance"
+            );
```
jubeira, line 450: I understand the problem here is that you're trying to burn more BPT than what you have. If I'm reading this correctly, you can exit with the exact same amounts you've entered, or the `remove` step reverts when you try to burn more BPT than what you have.

Instead of doing this, I'd use `bptAmountOut` as `maxBptAmountIn` in the remove step.

Maybe you can first run a query. If the query says that you will need more BPT in than what you have, run the actual operation using `expectRevert` and `bptAmountOut` as the limit; that will give you a known revert reason. If the query says that it's exactly the same, run the actual operation with the assertions inside the `try` block: 

```
            assertEq(bptAmountIn, bptAmountOut, "bptAmountIn should be equal to bptAmountOut");
            assertEq(amountsOut[0], amountsIn[0], "amountsOut[0] should be equal to amountsIn[0]");
            assertEq(amountsOut[1], amountsIn[1], "amountsOut[1] should be equal to amountsIn[1]");
```

That way you should avoid try / catch. In general "tests that might revert but it's not trivial to know when" are tricky, but I'd rather avoid try / catch logic here.

pkg/vault/test/foundry/utils/BaseVaultTest.sol
```diff
@@ -131,7 +131,13 @@ abstract contract BaseVaultTest is VaultStorage, BaseTest, Permit2Helpers {
 
     function setUp() public virtual override {
         BaseTest.setUp();
+        _setUpBaseVaultTest();
+        // Add initial liquidity
+        initPool();
+        vaultConvertFactor = vault.getConvertFactor();
+    }
 
+    function _setUpBaseVaultTest() internal {
```
jubeira, line 140: Is there any particular reason for this refactor?

jubeira (commented):
Thanks for pushing this @elshan-eth ! First pass done.

Overall it looks good. I have two general comments:
- Later on it'd be good to have a test that checks all of this against different pool types with different configurations. The base vault test architecture creates a single pool and runs the test, but that's not very representative. We can do that in a later stage.
- I think the assertion for `_testAddUnbalancedAndRemoveLiquidityProportional` is not strong enough to ensure that everything is OK. Let's leave a comment for now; we can check invariants in a follow-up.

Other than that, just minor suggestions and comment requests. Great job!

pkg/vault/test/foundry/utils/BaseExtremeAmountsTest.sol
```diff
@@ -0,0 +1,546 @@
...
+        _testAddAndRemoveLiquidityProportional(_calculateMaxBPTAmountForProportionalOperations());
+    }
+
+    function testAddAndRemoveLiquidityProportional_FuzzBPTAmount(uint256 exactBPTAmount) public {
+        _setUpVariables([poolInitAmount, poolInitAmount].toMemoryArray(), maxSwapFee / 2);
+
+        _testAddAndRemoveLiquidityProportional(_boundExactBPTAmount(exactBPTAmount));
+    }
+
+    function _calculateMaxBPTAmountForProportionalOperations() private view returns (uint256) {
```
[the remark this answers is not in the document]
elshan-eth: do you mean why is the `_calculateMaxBPTAmountForProportionalOperations` function needed?

pkg/vault/test/foundry/utils/BaseExtremeAmountsTest.sol
elshan-eth: I don't quite understand what's wrong here

pkg/vault/test/foundry/utils/BaseExtremeAmountsTest.sol
elshan-eth: the idea of ​​this test is to remove tokenA first and then tokenB and check what the difference is with proportional removal. so we take the BPT amount from the proportional addition and divide by two

pkg/vault/test/foundry/utils/BaseVaultTest.sol
elshan-eth: > Later on it'd be good to have a test that checks all of this against different pool types with different configurations. The base vault test architecture creates a single pool and runs the test, but that's not very representative. We can do that in a later stage.

This may help to implement this point

pkg/vault/test/foundry/utils/BaseExtremeAmountsTest.sol
jubeira: `amountsOut[0] > amountsIn[0]` means that the user got some `token0` out.

If the assertion fails it means that it also got `token1` out, which means the user got free money for sure.

But let's suppose that `amountsOut[0] >>> amountsIn[0]` (user got a lot of token0 out), and then `amountsOut[1] == amountsIn[1]` (or maybe just a bit larger, but not much). That difference of `token0` that the user got might be more valuable than what they paid for it in `token1`. If I'm not mistaken, this assertion is not catching that case.

elshan-eth: > If the assertion fails it means that it also got token1 out, which means the user got free money for sure. So if the assertion fails, we know for sure that something is wrong.

It means that he got token0 less and token1 more than before operations. This case is legal.

> But let's suppose that amountsOut[0] >>> amountsIn[0] (user got a lot of token0 out), and then amountsOut[1] == amountsIn[1] (or maybe just a bit larger, but not much). That difference of token0 that the user got might be more valuable than what they paid for it in token1. If I'm not mistaken, this assertion is not catching that case.

you mean if the user gets token0 more and token1 less, then the first token could be much more valuable, right?

pkg/vault/test/foundry/utils/BaseExtremeAmountsTest.sol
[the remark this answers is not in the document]
jubeira: yes, exactly

pkg/vault/test/foundry/utils/BaseExtremeAmountsTest.sol
jubeira: yes, exactly.
I think the only way to address this is computing the invariants before and after. We can do that in these tests, but there's always the issue with invariant error. The solution to that might be computing the invariant in Python with more precision, but that's out of scope for this PR.

jubeira (approved):
Let's solve merge conflicts and proceed with this one; this is a good base that I'd like to extend a bit in a few cases.

PR #849, opened 2024-08-07
Fix swap fee percentage

joaobrunoah:
# Description

ExactIn and ExactOut need to be as symmetrical as possible, and the fee% was applied differently. This PR makes it more symmetrical.

## Type of change

- [x] Bug fix 
- [ ] New feature 
- [ ] Breaking change 
- [ ] Dependency changes
- [ ] Code refactor / cleanup
- [ ] Other

## Checklist:


pkg/vault/contracts/Vault.sol
```diff
@@ -376,6 +376,12 @@ contract Vault is IVaultMain, VaultCommon, Proxy {
 
         // (1) and (2): get raw amounts and check limits.
         if (params.kind == SwapKind.EXACT_IN) {
+            // Set locals.swapFeeAmountScaled18 based on the amountCalculated.
+            if (state.swapFeePercentage > 0) {
```
jubeira, line 372: I can't comment outside the diff, but L370-375 should be eliminated as we're replacing that with this specific code for each swap kind.

pkg/vault/contracts/Vault.sol
```diff
@@ -391,6 +397,16 @@ contract Vault is IVaultMain, VaultCommon, Proxy {
...
             }
         } else {
+            // Set locals.swapFeeAmountScaled18 based on the amountCalculated.
+            if (state.swapFeePercentage > 0) {
+                // To ensure symmetry with EXACT_IN, the swap fee used by ExactOut is
+                // `fee%/(100% - fee%) * amountCalculated`. Add it to the calculated amountIn. Round up to avoid losses
+                // during precision loss.
+                locals.swapFeeAmountScaled18 =
+                    (amountCalculatedScaled18 * state.swapFeePercentage) /
+                    (FixedPoint.ONE - state.swapFeePercentage);
```
jubeira, line 399:
```suggestion
                locals.swapFeeAmountScaled18 = amountCalculatedScaled18.mulDivUp(state.swapFeePercentage, state.swapFeePercentage.complement());
```

As the comment indicates, we should probably round up.
`mulDivUp` does just that: raw multiplication, then div up.

If `swapFeePercentage` is ONE or greater, the complement will be 0 and `mulDivUp` will revert. We shouldn't need that protection, but it should be ok.

pkg/vault/contracts/Vault.sol
```diff
@@ -391,6 +397,16 @@ contract Vault is IVaultMain, VaultCommon, Proxy {
                 revert SwapLimit(amountOutRaw, params.limitRaw);
             }
         } else {
+            // Set locals.swapFeeAmountScaled18 based on the amountCalculated.
+            if (state.swapFeePercentage > 0) {
+                // To ensure symmetry with EXACT_IN, the swap fee used by ExactOut is
+                // `fee%/(100% - fee%) * amountCalculated`. Add it to the calculated amountIn. Round up to avoid losses
```
jubeira, line 395:
```suggestion
                // `fee% / (100% - fee%) * amountCalculated`. Add it to the calculated amountIn. Round up to avoid losses
```

jubeira (commented):
Thanks for catching this, great job!

Just a few comments; otherwise LGTM.

pkg/vault/contracts/Vault.sol
jubeira: already addressed while I was reviewing :)

pkg/vault/contracts/Vault.sol
EndymionJkb: It fits with just this change!

EndymionJkb: It fits with room to spare and all tests pass if you this and remove the if's. (swapFeeAmounts initialize to 0 anyway, so if it's ever 0, it will just multiply 0 by 0.)

```
locals.swapFeeAmountScaled18 =
    (amountCalculatedScaled18 * state.swapFeePercentage) /
    state.swapFeePercentage.complement();
```
The equivalent `mulDivUp` is failing tests, though?

jubeira:
```
(amountCalculatedScaled18 * state.swapFeePercentage) / (FixedPoint.ONE - state.swapFeePercentage);
```

This is equivalent to rounding down, not up.

EndymionJkb: Tests also round down, which is why they're failing, then.

joaobrunoah: Ok, fixed the code and tests.

EndymionJkb (approved):
Looks correct now; even saves a little gas. We might still add more tests later, but this should be sufficient to validate it.

PR #855, opened 2024-08-08
Fix bug that allowed vault to be drained with a malicious ERC4626 wrapper

### LINKED ISSUE #854: `deposit` might not be exact in for malicious wrappers
We need to reset the approvals after calling `deposit` to prevent malicious wrappers from draining the vault.

joaobrunoah:
# Description (by Trail of Bits)

Here's an example exploit path, assuming 1M USDC total vault balance.

1. Create a malicious wrapper  with USDC as the underlying and add 0 liquidity to it (only done to set `_bufferAssets`). The malicious wrapper `deposit` is a noop, does nothing and does not revert.
2. Call `erc4626BufferWrapOrUnwrap` with `EXACT_IN` and 1M tokens as `amountGiven`, triggering `_wrapWithBuffer`
3. Since the buffer doesn't have any liquidity, this [else](https://github.com/balancer/balancer-v3-monorepo/blob/a24ebf0141e9350a42639d8593c1436241deae59/pkg/vault/contracts/Vault.sol#L1156) clause is triggered. Since it's `EXACT_IN`, this [if](https://github.com/balancer/balancer-v3-monorepo/blob/a24ebf0141e9350a42639d8593c1436241deae59/pkg/vault/contracts/Vault.sol#L1165-L1173) clause is triggered. This approves the wrapper for 1M USDC but the `deposit` does nothing.
4. Since we don't have a `bufferUnderlyingSurplus` and the [deltas](https://github.com/balancer/balancer-v3-monorepo/blob/a24ebf0141e9350a42639d8593c1436241deae59/pkg/vault/contracts/Vault.sol#L1193-L1196) are 0, this [else](https://github.com/balancer/balancer-v3-monorepo/blob/a24ebf0141e9350a42639d8593c1436241deae59/pkg/vault/contracts/Vault.sol#L1230-L1233) clause is triggered overwriting the `amountInUnderlying` and `amountOutWrapped` with 0.
5. The user incurs no debt or credit allowing them to finish the transaction. But, they keep the approvals.
6. The wrapper uses `transferFrom(vault, exploiter, 1M USDC)` to steal the funds

## Type of change

- [x] Bug fix 
- [ ] New feature 
- [ ] Breaking change 
- [ ] Dependency changes
- [ ] Code refactor / cleanup
- [ ] Other

## Checklist:

Fixes #854


pkg/vault/test/foundry/utils/ERC4626TokenBrokenRateMock.sol
```diff
@@ -1,31 +0,0 @@
-// SPDX-License-Identifier: GPL-3.0-or-later
```
joaobrunoah, line 1: Unused test files

pkg/vault/test/foundry/BufferVaultPrimitive.t.sol
```diff
@@ -139,6 +147,34 @@ contract BufferVaultPrimitiveTest is BaseVaultTest {
...
+                    userData: bytes("")
+                })
+            )
+        );
+
+        vm.expectRevert(
+            abi.encodeWithSelector(IERC20Errors.ERC20InsufficientAllowance.selector, address(waDAI), 0, vaultBalance)
+        );
+        vm.prank(address(waDAI));
+        dai.transferFrom(address(vault), address(waDAI), vaultBalance);
```
jubeira, line 175: I'd (also) check that the allowance granted by the vault is 0 (maybe a good check to add in other wrapper tests as well?)

jubeira: ah nvm; it's encoded in the error.

As an alternative, maybe just check that the allowance is 0 instead of calling `transferFrom` and expecting revert?

pkg/vault/test/foundry/BufferVaultPrimitive.t.sol
```diff
@@ -481,4 +551,18 @@ contract BufferVaultPrimitiveTest is BaseVaultTest {
             exactAmountOut: exactAmountOut
         });
     }
+
+    /// @notice Hook used to create a vault approval using a malicious erc4626 and drain the vault.
+    function erc4626MaliciousHook(BufferWrapOrUnwrapParams memory params) external {
+        (, uint256 amountIn, uint256 amountOut) = vault.erc4626BufferWrapOrUnwrap(params);
+        if (params.kind == SwapKind.EXACT_OUT) {
+            // When the wrap is EXACT_OUT, a minimum amount of tokens must be wrapped. so, balances need to be settled
+            // at the end to don't revert the transaction and keep an approval to remove underlying tokens from the
```
EndymionJkb, line 560:
```suggestion
            // at the end to not revert the transaction and keep an approval to remove underlying tokens from the
```

EndymionJkb (approved):
Still added bytes even though we just moved a line; luckily it still fits.

What was the original intent of the unused stuff we're deleting (part of the tests for well-behaved wrappers)? Are we testing that elsewhere?

joaobrunoah:
> Still added bytes even though we just moved a line; luckily it still fits.
> 
> What was the original intent of the unused stuff we're deleting (part of the tests for well-behaved wrappers)? Are we testing that elsewhere?

So, regarding the part that mocks assets and shares used during wrap/unwrap, we decided not to check that in the code because it exclusively affected the buffer, so we don't have a test to revert the operation when convert is far from the wrap/unwrap operation result. However, the code to mock that behavior was in there, even without the tests, and that's what this code removes. We don't intend to test that, but we need to make sure the buffer balance is isolated from the vault.

pkg/vault/test/foundry/BufferVaultPrimitive.t.sol
joaobrunoah: Done!

PR #857, opened 2024-08-08
Fix fee collection (1: using credit)

### LINKED ISSUE #853: Update token reserves when collecting fees

jubeira:
# Description

In the current `main` we're giving the `ProtocolFeeController` a special privilege to pull tokens out of the vault, since we're approving it to pull tokens out in `collectAggregateFees`. The problem here is that the reserves need to be updated as well when doing this.

The mechanism to send tokens out of the vault properly is being used elsewhere extensively, and it involves supplying credit and then settling. This PR applies the same pattern to `collectAggregateFees`.

The result is a code that is a bit more complex on the fee controller side. But it eliminates alternate ways of pulling tokens out of the vault, and associated errors that may arise from them.

Tests WIP.

## Type of change

- [x] Bug fix 
- [ ] New feature 
- [ ] Breaking change 
- [ ] Dependency changes
- [ ] Code refactor / cleanup
- [ ] Other

## Checklist:

## Issue Resolution

Closes #853

### DIFF
--- a/pkg/interfaces/contracts/vault/IVaultAdmin.sol
+++ b/pkg/interfaces/contracts/vault/IVaultAdmin.sol
@@ -119,7 +119,9 @@ interface IVaultAdmin {
      * @dev Fees are sent to the ProtocolFeeController address.
      * @param pool The pool on which all aggregate fees should be collected
      */
-    function collectAggregateFees(address pool) external;
+    function collectAggregateFees(
+        address pool
+    ) external returns (uint256[] memory totalSwapFees, uint256[] memory totalYieldFees);
 
     /**
      * @notice Update an aggregate swap fee percentage.
--- a/pkg/interfaces/contracts/vault/IVaultAdmin.sol
+++ b/pkg/interfaces/contracts/vault/IVaultAdmin.sol
@@ -118,10 +118,12 @@ interface IVaultAdmin {
      * @notice Collects accumulated aggregate swap and yield fees for the specified pool.
      * @dev Fees are sent to the ProtocolFeeController address.
      * @param pool The pool on which all aggregate fees should be collected
+     * @return swapFeeAmounts An array with the total swap fees collected, sorted in token registration order
+     * @return yieldFeeAmounts An array with the total yield fees collected, sorted in token registration order
      */
     function collectAggregateFees(
         address pool
-    ) external returns (uint256[] memory totalSwapFees, uint256[] memory totalYieldFees);
+    ) external returns (uint256[] memory swapFeeAmounts, uint256[] memory yieldFeeAmounts);
 
     /**
      * @notice Update an aggregate swap fee percentage.
--- a/pkg/vault/contracts/ProtocolFeeController.sol
+++ b/pkg/vault/contracts/ProtocolFeeController.sol
@@ -134,9 +134,16 @@ contract ProtocolFeeController is
         _;
     }
 
+    modifier withValidPoolCreatorFee(uint256 newPoolCreatorFeePercentage) {
+        if (newPoolCreatorFeePercentage > FixedPoint.ONE) {
+            revert PoolCreatorFeePercentageTooHigh();
+        }
+        _;
+    }
+
     // Force collection and disaggregation (e.g., before changing protocol fee percentages).
     modifier withLatestFees(address pool) {
-        getVault().collectAggregateFees(pool);
+        collectAggregateFees(pool);
         _;
     }
 
@@ -150,7 +157,7 @@ contract ProtocolFeeController is
     }
 
     /// @inheritdoc IProtocolFeeController
-    function collectAggregateFees(address pool) external {
+    function collectAggregateFees(address pool) public {
         getVault().unlock(abi.encodeWithSelector(ProtocolFeeController.collectAggregateFeesHook.selector, pool));
     }
 
@@ -211,13 +218,7 @@ contract ProtocolFeeController is
     }
 
     /// @inheritdoc IProtocolFeeController
-    function updateProtocolSwapFeePercentage(address pool) external {
-        getVault().unlock(
-            abi.encodeWithSelector(ProtocolFeeController.updateProtocolSwapFeePercentageHook.selector, pool)
-        );
-    }
-
-    function updateProtocolSwapFeePercentageHook(address pool) external onlyVault withLatestFees(pool) {
+    function updateProtocolSwapFeePercentage(address pool) external withLatestFees(pool) {
         PoolFeeConfig memory feeConfig = _poolProtocolSwapFeePercentages[pool];
         uint256 globalProtocolSwapFee = _globalProtocolSwapFeePercentage;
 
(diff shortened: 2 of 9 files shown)


pkg/vault/contracts/VaultExplorer.sol
```diff
@@ -292,7 +292,7 @@ contract VaultExplorer is IVaultExplorer {
 
     /// @inheritdoc IVaultExplorer
     function collectAggregateFees(address pool) external {
-        return _vault.collectAggregateFees(pool);
```
EndymionJkb, line 295: Do we still want to expose this on the Vault Explorer, as it's not strictly a "Vault" operation anymore? Then the VaultExplorer would be just read only.

pkg/vault/contracts/VaultAdmin.sol
```diff
@@ -270,28 +270,32 @@ contract VaultAdmin is IVaultAdmin, VaultCommon, Authentication {
     }
 
     /// @inheritdoc IVaultAdmin
-    function collectAggregateFees(address pool) public onlyVaultDelegateCall nonReentrant withRegisteredPool(pool) {
+    function collectAggregateFees(
```
EndymionJkb, line 273: I was considering whether we should rename this, as it doesn't by itself collect the fees now - just calculates them. But it is zeroing out the fee balances, so that's a form of collection. I guess the name is fine.

pkg/vault/contracts/ProtocolFeeController.sol
```diff
@@ -134,9 +134,16 @@ contract ProtocolFeeController is
         _;
     }
 
+    modifier withValidPoolCreatorFee(uint256 newPoolCreatorFeePercentage) {
```
EndymionJkb, line 137: The check was already being done, but the others have a modifier, so adding this here makes it more consistent.

EndymionJkb (commented):
Much prefer this to #2. There should be only one way to remove tokens. I guess there will be more tests?

joaobrunoah (approved):
Nice that extracting fees to FeeController uses the unlock mechanism, it really gives some peace of mind. Besides, now to collect fees we need to call ProtocolFeeController, which makes more sense than calling the vault directly.

pkg/vault/contracts/VaultAdmin.sol
jubeira: well, `swap` doesn't actually do the swap either :)

pkg/vault/contracts/VaultAdmin.sol
```diff
@@ -270,28 +270,33 @@ contract VaultAdmin is IVaultAdmin, VaultCommon, Authentication {
...
+        address pool
+    )
+        public
+        onlyVaultDelegateCall
+        onlyWhenUnlocked
+        onlyProtocolFeeController
+        withRegisteredPool(pool)
```
jubeira, line 280: Note for the reviewers: modifiers have changed.

- There are no external calls here, so `nonReentrant` was removed. I've also removed `receiveAggregateFees` from this function, so it could be re-added if you think there's a reason I might be missing.
- `onlyWhenUnlocked` is now a must because this function supplies credit. Not having this modifier is a critical issue; there's a test for that too.
- `onlyProtocolFeeController` is now a must because otherwise anyone can claim the credit granted by calling this function, and then steal the fees. There's also a test for that.

pkg/vault/contracts/VaultExplorer.sol
EndymionJkb: Ok; I guess it makes it easier to test fees.

pkg/interfaces/contracts/vault/IVaultAdmin.sol
```diff
@@ -119,7 +119,9 @@ interface IVaultAdmin {
      * @dev Fees are sent to the ProtocolFeeController address.
      * @param pool The pool on which all aggregate fees should be collected
      */
-    function collectAggregateFees(address pool) external;
+    function collectAggregateFees(
+        address pool
+    ) external returns (uint256[] memory totalSwapFees, uint256[] memory totalYieldFees);
```
EndymionJkb, line 124:
```suggestion
    ) external returns (uint256[] memory swapFeeAmounts, uint256[] memory yieldFeeAmounts);
```
See above

pkg/vault/contracts/ProtocolFeeController.sol
```diff
@@ -149,6 +156,16 @@ contract ProtocolFeeController is
         return getVault();
     }
 
+    /// @inheritdoc IProtocolFeeController
+    function collectAggregateFees(address pool) public {
+        getVault().unlock(abi.encodeWithSelector(ProtocolFeeController.collectAggregateFeesHook.selector, pool));
+    }
+
+    function collectAggregateFeesHook(address pool) external onlyVault {
```
EndymionJkb, line 164:
```suggestion
    /**
     * @dev Copy and zero out the `aggregateFeeAmounts` collected in the Vault accounting, supplying credit
     * for each token. Then have the Vault transfer tokens to this contract, debiting each token for the amount
     * transferred so that the transaction settles when the hook returns.
     */
    function collectAggregateFeesHook(address pool) external onlyVault {
```

pkg/vault/contracts/ProtocolFeeController.sol
```diff
@@ -304,12 +321,21 @@ contract ProtocolFeeController is
...
-    function receiveAggregateFees(
+    /**
+     * @notice Settle fee credits from the vault.
+     * @dev This must be called after calling `collectAggregateFees` in the Vault. Note that since charging protocol
+     * fees (i.e., distributing tokens between pool and fee balances) occurs in the Vault, but fee collection
+     * happens in the ProtocolFeeController, the swap fees reported here may encompass multiple operations.
+     *
+     * @param pool The address of the pool on which the swap fees were charged
+     * @param swapFeeAmounts An array parallel to the pool tokens, with the swap fees collected in each token
+     * @param yieldFeeAmounts An array parallel to the pool tokens, with the yield fees collected in each token
```
EndymionJkb, line 332:
```suggestion
     * @param swapFeeAmounts An array with the total swap fees collected, sorted in token registration order
     * @param yieldFeeAmounts An array with the total yield fees collected, sorted in token registration order
```
Parameter consistency

pkg/vault/contracts/ProtocolFeeController.sol
```diff
@@ -149,6 +156,16 @@ contract ProtocolFeeController is
...
+    /// @inheritdoc IProtocolFeeController
+    function collectAggregateFees(address pool) public {
+        getVault().unlock(abi.encodeWithSelector(ProtocolFeeController.collectAggregateFeesHook.selector, pool));
+    }
+
+    function collectAggregateFeesHook(address pool) external onlyVault {
+        (uint256[] memory totalSwapFees, uint256[] memory totalYieldFees) = getVault().collectAggregateFees(pool);
+        _receiveAggregateFees(pool, totalSwapFees, totalYieldFees);
+    }
+
```
EndymionJkb, line 168: I suggest moving the `_receiveAggregateFees` functions out of the permissioned section and up to here, right under the hook, since they're no longer permissioned (i.e., called from here, vs. from the Vault). The enum should probably be at the top with the rest of the type declarations, too.

pkg/vault/contracts/VaultAdmin.sol
EndymionJkb: Was there a reason we used `withRegisteredPool` instead of `withInitializedPool` here? There wouldn't be fees on a pool until it's initialized, right? (The tests would have to be updated to initialize the pool before calling this.)

EndymionJkb (approved):
A few doc/organizational suggestions (and maybe a modifier change), but I much prefer this approach.

pkg/vault/contracts/VaultAdmin.sol
jubeira: I think it doesn't make much of a difference; a pool that is not initialized will have no fees and a pool that was just initialized pool won't have fees either.

elshan-eth (approved):
LGTM!

PR #859, opened 2024-08-08
Add rest tests for HooksConfigLib and PoolConfigLib (Mutation tests part #2)

elshan-eth:
# Description

Add rest unit tests for HooksConfigLib and PoolConfigLib.

## Type of change

- [ ] Bug fix 
- [ ] New feature 
- [ ] Breaking change 
- [ ] Dependency changes
- [ ] Code refactor / cleanup
- [X] Other

## Checklist:


EndymionJkb (commented):
First pass - too many to verify everything in one sitting. Have some initial suggestions.

pkg/vault/test/foundry/unit/HooksConfigLibHooksHelpers.t.sol
```diff
@@ -0,0 +1,1072 @@
...
+                amountCalculatedRaw,
+                router,
+                params,
+                state,
+                poolData,
+                IHooks(hooksContract)
+            );
+    }
+
+    function _callAfterSwapHookAndExpectRevert(
```
jubeira, line 338: This one expects revert when applying limits.

There's one more case missing: the after swap hook returning false.

pkg/vault/test/foundry/unit/HooksConfigLibHooksHelpers.t.sol
```diff
@@ -0,0 +1,1072 @@
...
+            amountCalculatedRaw,
+            hookAdjustedAmountCalculatedRaw,
+            params,
+            state,
+            poolData
+        );
+
+        assertEq(value, hookAdjustedAmountCalculatedRaw, "return value mismatch");
+    }
+
```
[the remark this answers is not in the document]
jubeira: hmm I think that's already covered in the first two standard cases right?

jubeira (commented):
Awesome job @elshan-eth! This looks pretty good and complete.

I only have a few minor comments and one missing test (after swap on failure), adding to those from @EndymionJkb. Other than that I think this is good to go. 

Thanks for pushing this!

pkg/vault/test/foundry/unit/HooksConfigLibHooksHelpers.t.sol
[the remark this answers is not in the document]
elshan-eth: Yes, it tested in `testCallAfterSwapHookExactOut`

elshan-eth:
> hmm I think that's already covered in the first two standard cases right?

no, here we tested it for EXACT_OUT with the adjusted amount but the previous test EXACT_IN

EndymionJkb (approved):
Looks good now - I updated to current main to run the coverage tests, and it looks like 100% (there is one "invalid enum" line that can't be tested, at least outside a fork test). Note #920 on top of this one.

pkg/vault/test/foundry/unit/HooksConfigLibHelpers.t.sol
```diff
@@ -0,0 +1,1116 @@
...
+    address router = address(0x00);
+    address pool = address(0x11);
+    address hooksContract = address(0x1234567890123456789012345678901234567890);
+    HooksConfigLibMock hooksConfigLibMock;
+
+    function setUp() public {
+        hooksConfigLibMock = new HooksConfigLibMock();
+    }
+
+    //#region callComputeDynamicSwapFeeHook
```
joaobrunoah, line 30: About these 'region' comments, I think we removed them a while ago in other test files, not sure we should keep them here.

joaobrunoah (approved):
LGTM!

pkg/vault/test/foundry/unit/HooksConfigLibHelpers.t.sol
EndymionJkb: I'd considered a cleanup PR to remove those... they're kind of an "Elshan marker" :) I believe Elshan uses some setting in his IDE that makes this look better for him, but it's not standard, and nobody else sees it. Regular VSCode lets you collapse sections without any kind of marker.

Would lean toward removing all these.

PR #865, opened 2024-08-11
Prevent multiple buffer initialization

### LINKED ISSUE #864: Prevent multiple ERC4626 buffer initialization
Unlike with pools, there is no explicit initialization step with ERC4626 wrapped token buffers. We use the original v2 approach of a single `addLiquidity` function that checks whether it's been initialized by checking the `bufferAssets` mapping. This mapping stores the underlying token of the wrapper; if it is zero, the buffer hasn't been initialized.

Unless... the underlying token is zero. This is probably not a very useful wrapper, but it would cause the initialization code to be run more than once, which is not intended.

Introducing a separate initialization step is probably overkill in this case. A buffer is a unique combination of the wrapper + underlying address. We are already checking for the case where an upgradeable wrapper changes its asset, which will also revert. Here, we also ensure that the underlying token is valid (i.e., non-zero).

EndymionJkb:
# Description

Unlike with pools, there is no explicit initialization step with buffers. It just checks for a non-zero underlying asset in the `bufferAssets` mapping.

Which means that if a malicious wrapper set its underlying asset address to zero, it would cause the initialization code in `addLiquidityToBuffer` to be run multiple times. That wouldn't be a very useful wrapper - not sure if there's an exploit there - but in any case that is unintended behavior, which this PR prevents.

A buffer is a unique combination of the wrapper contract address (which of course can't be zero, and would revert if you passed 0 anyway, since the first line calls `asset()` on it), and the underlying asset, which we're now enforcing also cannot be zero.

## Type of change

- [X] Bug fix 
- [ ] New feature 
- [ ] Breaking change 
- [ ] Dependency changes
- [ ] Code refactor / cleanup
- [ ] Other

## Checklist:

## Issue Resolution

Resolves #864


pkg/vault/contracts/VaultAdmin.sol
```diff
@@ -432,6 +432,12 @@ contract VaultAdmin is IVaultAdmin, VaultCommon, Authentication {
     {
         address underlyingToken = wrappedToken.asset();
 
+        if (underlyingToken == address(0)) {
+            // Should never happen, but a malicious wrapper could return the zero address and cause the buffer
+            // initialization code to run more than once.
+            revert InvalidUnderlyingTokenAsset();
+        }
```
jubeira, line 439: This can only happen when `_bufferAssets[wrappedToken]` is not initialized, but I think it's fine to have it here. Even more clear that `address(0)` asset is not accepted.

jubeira (approved):
LGTM! Thanks @EndymionJkb

joaobrunoah (approved):
LGTM!

PR #867, opened 2024-08-12
Remove infinite BPT approval for vault as spender

jubeira:
# Description

This UX optimization stopped making sense when we removed `transferFrom` calls from the Vault (namely when `takeFrom` / `retrieve` was removed). In V2 it made more sense in the context of BPTs being separate contracts from the Vault, with preminted BPT, trusted relayers, etc.

The V3 Vault never uses `transferFrom` currently, as we rely on `settle` and plain transfers from senders to pay debts. Removing this also saves some bytecode and gas. There's no strong reason to keep this.

## Type of change

- [ ] Bug fix 
- [ ] New feature 
- [ ] Breaking change 
- [ ] Dependency changes
- [x] Code refactor / cleanup
- [x] Optimization: [x] gas / [x] bytecode
- [ ] Other

## Checklist:

## Issue Resolution

N/A


EndymionJkb (approved):
Every 20 bytes and 100 gas helps!

This is another case where we theoretically have two mechanisms to do something (or a way to bypass the "usual" method: once again, approval vs. using our credit/debit mechanism). Even though it's the Vault itself this time, so it's not dangerous.

PR #873, opened 2024-08-13
Use storage vs. function call in ProtocolFeeController

EndymionJkb:
# Description

Slight optimization noted looked at ProtocolFeeController tests. It's a SingletonAuthentication (because it needs `authenticate`), and a VaultGuard (because it needs `onlyVault`). VaultGuard already makes an immutable _vault storage variable accessible, so we can use that rather than a function call. (Probably optimizes to the same thing and makes no difference, but more consistent with other contracts that have the vault address embedded.)

## Type of change

- [ ] Bug fix 
- [ ] New feature 
- [ ] Breaking change 
- [ ] Dependency changes
- [X] Code refactor / cleanup
- [ ] Other

## Checklist:


joaobrunoah (approved):
LGTM!

PR #874, opened 2024-08-13
Use sender for buffer add liquidity + variable name refactor

jubeira:
# Description

We need to use `msg.sender` when adding liquidity for buffers, otherwise any caller can add liquidity in the name of someone else. The actual LP doesn't lose funds, but the caller can use the approvals granted to lock the tokens into buffers, which is not correct. This is the same pattern we use for regular `addLiquidity`.

Also a minor refactor so that the vault admin code matches the interface.

## Type of change

- [x] Bug fix 
- [ ] New feature 
- [ ] Breaking change 
- [ ] Dependency changes
- [x] Code refactor / cleanup
- [ ] Other

## Checklist:

## Issue Resolution

N/A

### DIFF
--- a/pkg/interfaces/contracts/vault/IRouter.sol
+++ b/pkg/interfaces/contracts/vault/IRouter.sol
@@ -317,16 +317,13 @@ interface IRouter {
      * @param wrappedToken Address of the wrapped token that implements IERC4626
      * @param amountUnderlyingRaw Amount of underlying tokens that will be deposited into the buffer
      * @param amountWrappedRaw Amount of wrapped tokens that will be deposited into the buffer
-     * @param sharesOwner Address that will own the liquidity. Only this contract will be able to remove liquidity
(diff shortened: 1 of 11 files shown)


elshan-eth (approved):
Cool! LGTM!

EndymionJkb (approved):
Why did we do this in the first place? Was this going to be permissioned at one time (i.e., `sharesOwner` was an "add" equivalent of withdrawing to a `receiver`)?

Anyway looks better now.

jubeira:
> Was this going to be permissioned at one time

`remove` must be permissioned because the vault can't know who the real sender is, so it needs a router that reports the sender correctly (otherwise funds can be stolen).

For `add` that's not the case since the address from where the funds are pulled gets the buffer shares anyways, but that's still not good enough to prevent others using your funds in the first place.

PR #875, opened 2024-08-13
Simplify authentication by role

EndymionJkb:
# Description

The authenticate-by-role design - originally intended to save duplication of modifiers - started ballooning a bit when we added the onlyOwner concept for swap fees, and is unnecessarily complicated. It uses an assignment function during registration that stores essentially duplicate information. We need to know the roles, which are already stored in _poolRoleAccounts, and whether the function is onlyOwner. Given that there are only two functions this is used with, that information can be hard-coded into the respective functions.

This eliminates the extra storage and processing on registration, and replaces the `authenticateByRole` modifier with a function that implements the logic, with the two parameters (permissioned account and onlyOwner flag), passed in.

It saves a fair bit of bytecode, albeit only in the extensions.

## Type of change

- [ ] Bug fix 
- [ ] New feature 
- [ ] Breaking change 
- [ ] Dependency changes
- [X] Code refactor / cleanup
- [ ] Other

## Checklist:


pkg/vault/contracts/VaultAdmin.sol
```diff
@@ -550,6 +520,29 @@ contract VaultAdmin is IVaultAdmin, VaultCommon, Authentication {
         emit AuthorizerChanged(newAuthorizer);
     }
 
+    function _ensureAuthenticatedByRole(address pool, address roleAddress, bool onlyOwner) private view {
```
jubeira, line 523: I like the change, but the core logic is still the same.
This is not a strong opinion, but I'd like to hear your thoughts.

How about splitting this one in 2 functions, removing `onlyOwner`?
- For functions that are `onlyOwner`, same name applies; just check whether role address matches `msg.sender` in case there's a role address set.
- For functions that are not only owner, check both role address and call `_canPerform` (`_ensureAuthenticatedByRoleOrAuthorized`? too long?)

You might need a second lower level helper equivalent to `_canPerform` that returns a boolean and doesn't revert for roles.

jubeira (commented):
Thanks @EndymionJkb, I like mostly-red PRs!

Just one suggestion: maybe splitting the function into 2 might make things even more explicit. From the caller perspective, `onlyOwner` boolean doesn't sound crystal clear. We have plenty of space in the admin, so maybe 2 functions might be better in this case.

pkg/vault/contracts/VaultAdmin.sol
EndymionJkb: Ok; refactored again to simplify it further.

joaobrunoah (approved):
LGTM!

jubeira (approved):
LGTM, I like it!

PR #879, opened 2024-08-14
Protect PoolData load/write to storage paths

### LINKED ISSUE #878: Consider `_loadPoolDataUpdatingBalancesAndYieldFees` being `nonReentrant`
`_loadPoolDataUpdatingBalancesAndYieldFees` only does `view` external calls, and updates the vault storage afterwards (no external calls).

Probably unnecessary to make it `nonReentrant`.

EndymionJkb:
# Description

Before the introduction of the PoolDataLib and clear separation of read/write operations on pools, it was less clear when functions needed to be protected against reentrancy. We tended to be conservative, and protected all storage update functions, even if there were no external non-view calls. We're still being conservative, but more consistent.

This makes all calls to `_loadPoolData` that write to storage protected by the nonReentrant modifier.

## Type of change

- [ ] Bug fix 
- [ ] New feature 
- [ ] Breaking change 
- [ ] Dependency changes
- [X] Code refactor / cleanup
- [X] Optimization: [X] gas / [ ] bytecode
- [ ] Other

## Checklist:

## Issue Resolution

Resolves #878


pkg/vault/contracts/VaultCommon.sol
```diff
@@ -300,22 +309,14 @@ abstract contract VaultCommon is IVaultEvents, IVaultErrors, VaultStorage, Reent
...
     /**
      * @dev Fill in PoolData, including paying protocol yield fees and computing final raw and live balances.
-     * This function modifies protocol fees and balance storage. Since it modifies storage and makes external
-     * calls, it must be nonReentrant.
      * Side effects: updates `_aggregateFeeAmounts` and `_poolTokenBalances` in storage.
      */
     function _loadPoolDataUpdatingBalancesAndYieldFees(
         address pool,
         Rounding roundingDirection
-    ) internal nonReentrant returns (PoolData memory poolData) {
```
joaobrunoah, line 310: In `poolData.load`, we update rates with `getTokenRate()`, which calls `getRate()` from the RateProvider. Although getRate is a view function, someone can manipulate the rate somehow. For example, swap() calls this function, and when getting the rate, the attacker executes a different vault function (like addLiquidity) with the manipulated rate and then returns another rate in the swap. Not sure if this is entirely possible, but it'd be a Read-only reentrancy.

Swap, Add and Remove liquidity external calls are reentrant, so should we remove the nonReentrant modifier from here? I mean, the attack above would only happen with a malicious rate provider, but with permissionless rate providers, maybe the nonReentrant modifier here is a guardrail.

joaobrunoah (commented):
Maybe I'm being too paranoid or my example don't make any sense, so if it doesn't just ignore it, but I think removing the nonReentrant modifier allows a read-only reentrancy exploit.

pkg/vault/contracts/VaultCommon.sol
EndymionJkb: The auditors brought this up. If it's actually unsafe, then instead of removing it, we'd move it to `_loadPoolData`, to protect both.

joaobrunoah: Got it.

joaobrunoah (dismissed):
LGTM!

pkg/vault/contracts/VaultCommon.sol
EndymionJkb: Would like to hear what @jubeira / the auditors think before merging.

jubeira: you can only call `view` functions anyways.

If the rate provider happens to call the vault to get some data, that could be a problem. But I don't think that's a valid use-case in principle.

> If it's actually unsafe, then instead of removing it, we'd move it to _loadPoolData, to protect both.

I'd wait a bit longer before merging; this is true.

jubeira:
I would wait a bit before merging.

We can't make `_loadPoolData`, but we can make the functions that call it and modify the state non-reentrant (`initialize` and `_setPoolRecoveryMode`). 

In principle I don't think this protection is really necessary; I believe we've put in place just out of abundance of caution, but that stance still holds.

pkg/vault/contracts/VaultCommon.sol
```diff
@@ -300,22 +309,16 @@ abstract contract VaultCommon is IVaultEvents, IVaultErrors, VaultStorage, Reent
...
     ) internal nonReentrant returns (PoolData memory poolData) {
         // Initialize poolData with base information for subsequent calculations.
-        poolData.load(
-            _poolTokenBalances[pool],
-            _poolConfigBits[pool],
-            _poolTokenInfo[pool],
-            _poolTokens[pool],
-            roundingDirection
-        );
+        poolData = _loadPoolData(pool, roundingDirection);
```
joaobrunoah, line 321: Looks like this change increased the vault bytecode by 23 bytes. Is it really needed?

joaobrunoah (approved):
LGTM! Only a minor comment about vault contract size.

pkg/vault/contracts/VaultCommon.sol
EndymionJkb: I'll check if that's really what did it. Seems counter-intuitive that calling a function vs. duplicating that function inline *increases* the bytecode - it was intended to 1) remove duplication (and ensure what should be identical code doesn't diverge); and 2) decrease bytecode by sharing code. But the optimizer is voodoo, so who knows. If I can save 23 bytes, I'll put it back.

PR #880, opened 2024-08-14
Disable recovery mode fees

### LINKED ISSUE #876: Disable recovery mode: don't charge fees loading pool data
We should probably first load pool data without charging fees (as live balances are invalid at the time) and only then update the pool config.

EndymionJkb:
# Description

We intend to forfeit yield fees on re-enabling Recovery Mode. (They cannot be charged during Recovery Mode, as updating the live balances would involve external calls, which might fail and block withdrawals.) To ensure yield fees aren't erroneously subtracted from balances, keep the Recovery Mode bit set during the accounting during a re-enable.

I added a test for this - ensuring the balances are unchanged after an operation that would verify fees. (If you restore the original order of operations in `_setPoolRecoveryMode`, this test should fail.)

## Type of change

- [X] Bug fix 
- [ ] New feature 
- [ ] Breaking change 
- [ ] Dependency changes
- [ ] Code refactor / cleanup
- [ ] Other

## Checklist:

## Issue Resolution

Resolves #876


jubeira (approved):
LGTM, thanks @EndymionJkb !

Can we also check the slot that stores the fees in `_assertBalanceChangeRemoveLiquidityRecovery` as well? that one should stay the same before and after (both before and after removing, and before and after disabling recovery mode). It's a bit of an overkill, but truth is that the system should also have that property.

PR #881, opened 2024-08-14
Buffers explicit initialization

### LINKED ISSUE #868: Explicit buffer init step
For security reasons, it might be better to prevent buffer wraps / unwraps until an init step (only executable once) freezes the wrapper assets and mints the initial shares.

This will make the setup a bit more involved, but should prevent issues arising from bad wrappers that can change assets after init or after the first wrap / unwrap.

jubeira:
# Description

Given the security concerns around buffers, having an explicit init step to fix the underlying token forever and burning some shares before everything else can occur seems to be a good idea.

With this change:
- Buffers cannot operate at all before calling initialize (`wrap-unwrap` / `add` / `remove` will all be rejected)
- Initialize does not accept `address(0)` as a valid underlying asset (as a consequence, a buffer is consider initialized when the underlying asset is not 0).
- `Add` and `wrap / unwrap` will revert if the asset changes with respect to the initialized one. Remove won't revert in this case to prevent external calls in the `remove` workflow.
- The events changed a bit, since now events related to shares are separated from events related to assets going into buffer balances. There was a small mistake here, as burnt shares were not being considered in the initial event before; now it should be correct.

This is ready for a first pass while I continue with the tests.

## Type of change

- [x] Bug fix 
- [ ] New feature 
- [ ] Breaking change 
- [ ] Dependency changes
- [ ] Code refactor / cleanup
- [ ] Other

## Checklist:

## Issue Resolution

Closes #868


EndymionJkb (commented):
We had that outage just as I submitted this... don't remember exactly what I wrote for this first pass. Definitely cleaner; I tried to consolidate some repeated code (see comments), but everything I did actually made the bytecode worse. Compiler optimization is voodoo.

pkg/interfaces/contracts/vault/IVaultErrors.sol
```diff
@@ -284,6 +285,43 @@ interface IVaultErrors {
      */
     error PoolPauseWindowExpired(address pool);
 
+    /*******************************************************************************
+                            Yield-bearing token buffers
```
joaobrunoah, line 289:
```suggestion
                            ERC4626 token buffers
```
Sometimes we call yield-bearing, and other times we call ERC4626. As a suggestion, maybe we could adopt ERC4626 everywhere, since yield-bearing is a long name (even for test functions, the names are huge when we adopt yieldBearingPool, we could improve code readability with erc4626Pool)

pkg/interfaces/contracts/vault/IVaultErrors.sol
```diff
@@ -284,6 +285,43 @@ interface IVaultErrors {
...
+    error BufferNotInitialized(IERC4626 wrappedToken);
+
+    /// @dev The user is trying to remove more than their allocated shares from the buffer.
+    error NotEnoughBufferShares();
+
+    /// @dev The wrapped token asset does not match the underlying token.
+    error WrongWrappedTokenAsset(IERC4626 wrappedToken, address underlyingToken);
+
+    /// @dev A wrapped token reported the zero address as its underlying token asset.
+    error InvalidUnderlyingTokenAsset();
```
joaobrunoah, line 305: I guess it's not related to this PR, but underlying token and asset is the same thing. Maybe this error should be `InvalidUnderlyingToken` or `InvalidAsset` (or even `InvalidWrapperAsset`, to reference the function `wrapper.asset()`).

pkg/interfaces/contracts/vault/IVaultErrors.sol
```diff
@@ -284,6 +285,43 @@ interface IVaultErrors {
...
+
+    /// @dev The wrappedToken wrap/unwrap function did not deposit/return the expected amount of underlying tokens.
+    error WrongUnderlyingAmount(address wrappedToken);
+
+    /// @dev The wrappedToken wrap/unwrap function did not burn/mint the expected amount of wrapped tokens.
+    error WrongWrappedAmount(address wrappedToken);
```
joaobrunoah, line 311: I believe we're not using these errors anymore, I forgot to remove them.

pkg/vault/contracts/VaultAdmin.sol
```diff
@@ -438,41 +442,92 @@ contract VaultAdmin is IVaultAdmin, VaultCommon, Authentication {
...
+        bytes32 bufferBalances = _bufferTokenBalances[wrappedToken];
+        bufferBalances = PackedTokenBalance.toPackedBalance(
+            bufferBalances.getBalanceRaw() + amountUnderlyingRaw,
+            bufferBalances.getBalanceDerived() + amountWrappedRaw
+        );
+
+        _bufferTokenBalances[wrappedToken] = bufferBalances;
+
         // Amount of shares to issue is the total underlying token that the user is depositing.
         issuedShares = wrappedToken.convertToAssets(amountWrappedRaw) + amountUnderlyingRaw;
```
joaobrunoah, line 462: Just to mention as auditors got it, `convertToAssets` can be manipulated in a way that returns a big number for any input, including 0. This would make the initializeBuffer() to mint shares for an empty buffer and consider the buffer to be initialized.

Not sure if we want to prevent that, because we'd need to initialize a malicious wrapper in the first place, but maybe we can consider a different "invariant" formula to calculate buffer shares.

pkg/vault/contracts/VaultAdmin.sol
```diff
@@ -438,41 +442,92 @@ contract VaultAdmin is IVaultAdmin, VaultCommon, Authentication {
...
+        onlyVaultDelegateCall
+        onlyWhenUnlocked
+        whenVaultBuffersAreNotPaused
+        withInitializedBuffer(wrappedToken)
+        nonReentrant
+        returns (uint256 issuedShares)
+    {
+        // Check wrapped token asset correctness.
+        address underlyingToken = wrappedToken.asset();
+        _ensureCorrectBufferAsset(wrappedToken, underlyingToken);
```
joaobrunoah, line 490: Maybe these lines could be inside the "withInitializedBuffer" modifier, since they'll repeat in each function that requires an initialized buffer.

pkg/vault/contracts/VaultAdmin.sol
```diff
@@ -438,41 +442,92 @@ contract VaultAdmin is IVaultAdmin, VaultCommon, Authentication {
...
             bufferBalances.getBalanceRaw() + amountUnderlyingRaw,
             bufferBalances.getBalanceDerived() + amountWrappedRaw
         );
-
         _bufferTokenBalances[wrappedToken] = bufferBalances;
 
-        _takeDebt(IERC20(underlyingToken), amountUnderlyingRaw);
-        _takeDebt(wrappedToken, amountWrappedRaw);
+        // Amount of shares to issue is the total underlying token that the user is depositing.
+        issuedShares = wrappedToken.convertToAssets(amountWrappedRaw) + amountUnderlyingRaw;
```
joaobrunoah, line 505: This line probably requires a function since it now repeats for buffer initialization and buffer addLiquidity.

pkg/vault/contracts/VaultAdmin.sol
joaobrunoah: Maybe we can consider that the initialization is proportional, and then use the proportion of initialization in the addLiquidity. Therefore, we wouldn't need to call an external function to calculate the issuedShares, we'd just use the proportion of underlying and wrapped tokens to calculate the internal rate of the buffer.

joaobrunoah (approved):
Minor comments, but LGTM!

pkg/interfaces/contracts/vault/IVaultErrors.sol
EndymionJkb: InvalidERC4626Asset sounds good, but we use underlyingToken all over the place in the API, so maybe InvalidUnderlyingToken is best to avoid a massive rename

pkg/vault/contracts/VaultAdmin.sol
EndymionJkb: I tried all that (see my comment above). Everything increases the bytecode size.

pkg/interfaces/contracts/vault/IVaultErrors.sol
jubeira: yeah I thought about this one. I think the confusion arises from getting `underlyingToken` when calling `wraped.asset()`.
Let's go for `InvalidUnderlyingToken` I guess.

pkg/vault/contracts/VaultAdmin.sol
jubeira: The main problem that this prevents is a malicious wrapper changing the asset, which causes very bad problems down the line.

If the wrapper is malicious and you're initializing / adding liquidity, you're getting shares for a buffer that doesn't make real sense. We can't prevent that, but all of that should be worthless anyways. As I understand it, the main point is not to mix good wrappers and assets with bad wrappers that can somehow access good assets.

pkg/vault/contracts/VaultAdmin.sol
jubeira: the problem is that we need the underlying token besides the check. I'd evaluate other options after the first pass; I think this part is clear enough like this for now. It's also the most straightforward way considering the starting point anyways.

pkg/vault/contracts/VaultAdmin.sol
jubeira: That would help 'reusability', but it introduces one more hop that sounds a bit unnecessary given that it's a single line. If it was more than one line I'd be more inclined to put it into a function, but if it's just for this I think it's more readable like this.

pkg/vault/contracts/VaultAdmin.sol
jubeira: That's not exactly the case.

`remove` needs initialized buffer (asset != 0), but shouldn't need to check whether the current wrapper asset matches. That way it works more similarly to a remove liquidity recovery, where no external calls are made.

For good wrappers it should always be the case, and checking whether the asset changed in the same step as checking whether the asset is not address(0) would make sense too.

pkg/interfaces/contracts/vault/IRouter.sol
```diff
@@ -312,8 +312,26 @@ interface IRouter {
                             Yield-bearing token buffers
     *******************************************************************************/
 
+    /**
+     * @notice Adds liquidity for the first time to a yield-bearing token buffer (internal ERC4626 buffer in the Vault).
+     * @dev Calling this method binds the wrapped token to its underlying asset internally; the asset in the wrapper
+     * cannot change afterwards, or every other operation on that wrapper (add / remove / wrap / unwrap) will fail.
+
```
EndymionJkb, line 319:
```suggestion
     *
```
Still need the "*"

pkg/interfaces/contracts/vault/IRouter.sol
```diff
@@ -329,6 +347,8 @@ interface IRouter {
     /**
      * @notice Removes liquidity from a yield-bearing token buffer (internal ERC4626 buffer in the Vault).
      * @dev Only proportional withdrawals are supported, and removing liquidity is permissioned.
+     * Requires the buffer to be initialized beforehand.
+
```
EndymionJkb, line 351:
```suggestion
     *
```

pkg/interfaces/contracts/vault/IVaultErrors.sol
```diff
@@ -284,6 +285,40 @@ interface IVaultErrors {
...
+
+    /// @dev The wrapped token asset does not match the underlying token.
+    error WrongUnderlyingToken(IERC4626 wrappedToken, address underlyingToken);
+
+    /// @dev A wrapped token reported the zero address as its underlying token asset.
+    error InvalidUnderlyingToken();
+
+    /// @dev The wrappedToken wrap/unwrap function did not deposit/return the expected amount of underlying tokens.
+    error WrongUnderlyingAmount(address wrappedToken);
+
```
EndymionJkb, line 309:
```suggestion
```
Unused

pkg/interfaces/contracts/vault/IVaultMain.sol
```diff
@@ -130,6 +130,7 @@ interface IVaultMain {
      * @notice Wraps/unwraps tokens based on provided parameters, using the buffer of the wrapped token when it has
      * enough liquidity to avoid external calls.
      * @dev All parameters are given in raw token decimal encoding.
+     * Requires the buffer to be initialized beforehand.
```
EndymionJkb, line 133:
```suggestion
     * @dev All parameters are given in raw token decimal encoding. Requires the buffer to be initialized beforehand.
```

pkg/vault/contracts/VaultCommon.sol
```diff
@@ -241,6 +242,32 @@ abstract contract VaultCommon is IVaultEvents, IVaultErrors, VaultStorage, Reent
         return config.isPoolInitialized();
     }
 
+    /*******************************************************************************
+                        Buffer Initialization & validation
```
EndymionJkb, line 246:
```suggestion
                          Buffer Initialization & Validation
```

EndymionJkb (approved):
Starting to glaze over looking at the tests (hopefully Joao covered those better), but looks good!

pkg/vault/contracts/VaultAdmin.sol
```diff
@@ -432,49 +432,104 @@ contract VaultAdmin is IVaultAdmin, VaultCommon, Authentication {
...
+        );
+
+        _bufferTokenBalances[wrappedToken] = bufferBalances;
+
         // Amount of shares to issue is the total underlying token that the user is depositing.
         issuedShares = wrappedToken.convertToAssets(amountWrappedRaw) + amountUnderlyingRaw;
 
-        if (_bufferAssets[wrappedToken] == address(0)) {
-            // Buffer is not initialized yet, so we initialize it.
+        _ensureMinimumTotalSupply(issuedShares);
```
danielmkm, line 466: The flow here is a bit funky

```solidity
_ensureMinimumTotalSupply(issuedShares); // checks issuedShares > minimum

issuedShares -= _MINIMUM_TOTAL_SUPPLY; // subtracts minimum from issued shares
_mintBufferShares(wrappedToken, sharesOwner, issuedShares); // internally calls _ensureMinimumTotalSupply with the new smaller amount
```

We're calling `_ensureMinimumTotalSupply` twice with different numbers
- When calling `_mintBufferShares`, assuming the current supply is greater than the minimum the `newTotalSupply` can never be less than the minimum.
- The only place we need to check for the minimum is on init, no?
- If the argument is an additional security layer, why are you not checking for minimum on remove. This is arguably the more important place for this extra check.

pkg/vault/contracts/VaultAdmin.sol
```diff
@@ -432,49 +432,104 @@ contract VaultAdmin is IVaultAdmin, VaultCommon, Authentication {
...
 
-            // Burn MINIMUM_TOTAL_SUPPLY shares, so the buffer can never go back to zero liquidity
-            // (avoids rounding issues with low liquidity).
-            _bufferTotalShares[wrappedToken] = _MINIMUM_TOTAL_SUPPLY;
-            issuedShares -= _MINIMUM_TOTAL_SUPPLY;
-        } else if (_bufferAssets[wrappedToken] != underlyingToken) {
-            // Asset was changed since the first bufferAddLiquidity call.
-            revert WrongWrappedTokenAsset(address(wrappedToken));
-        }
+        emit LiquidityAddedToBuffer(wrappedToken, amountUnderlyingRaw, amountWrappedRaw);
```
danielmkm, line 472: I prefer the approach to having an add / remove event, but this is at odds to the flow we use in addLiquidity/removeLiquidity where we call the `PoolBalanceChanged` event.

pkg/vault/contracts/VaultAdmin.sol
danielmkm: Agree with juani, if the wrapper provides a malicious rate, it can only impact it's own balances. The goal here is to ensure isolation of balances.

pkg/vault/contracts/VaultAdmin.sol
```diff
@@ -486,22 +541,23 @@ contract VaultAdmin is IVaultAdmin, VaultCommon, Authentication {
...
         onlyVaultDelegateCall
         onlyWhenUnlocked
+        withInitializedBuffer(wrappedToken)
         authenticate
         nonReentrant
         returns (uint256 removedUnderlyingBalanceRaw, uint256 removedWrappedBalanceRaw)
     {
-        bytes32 bufferBalances = _bufferTokenBalances[wrappedToken];
-
         if (sharesToRemove > _bufferLpShares[wrappedToken][sharesOwner]) {
```
danielmkm, line 549: It feels like this check and the zero address check should be closer together.

PR #883, opened 2024-08-15
Change token rates of ERC4626 in tests to avoid confusion between underlying and wrapped amounts

joaobrunoah:
# Description

This code:
* Changes the rates of ERC4626 tokens to something different than 1, so confusions between underlying and wrapped amounts don't happen when testing;
* Fixes the calculated amounts returned by add and remove liquidity from ERC4626 pool (It was using wrapped amounts instead of underlying);

## Type of change

- [ ] Bug fix 
- [ ] New feature 
- [ ] Breaking change 
- [ ] Dependency changes
- [x] Code refactor / cleanup
- [ ] Other

## Checklist:


EndymionJkb (commented):
Before looking in detail, the description lists 3 things here; one is new functionality, one is for testing, and one is a bug fix. Should these be separate PRs?

If it's based of #881, should that be the target branch (for at least the new functionality one)?

joaobrunoah:
> Before looking in detail, the description lists 3 things here; one is new functionality, one is for testing, and one is a bug fix. Should these be separate PRs?
> 
> If it's based of #881, should that be the target branch (for at least the new functionality one)?

The first and second points are related (changing token rates breaks the tests, not because of a bad implementation of tests, but for a bug in the router).

The third point depends on the other two, but I can separate this into another PR and target this one.

pkg/vault/contracts/BatchRouter.sol
```diff
@@ -904,27 +910,27 @@ contract BatchRouter is IBatchRouter, BatchRouterStorage, RouterCommon, Reentran
...
         );
 
-        if (EVMCallModeHelpers.isStaticCall()) {
-            return amountsOut;
-        }
+        bool isStaticCall = EVMCallModeHelpers.isStaticCall();
 
         for (uint256 i = 0; i < erc4626PoolTokens.length; ++i) {
             IERC4626 wrappedToken = IERC4626(address(erc4626PoolTokens[i]));
             IERC20 underlyingToken = IERC20(wrappedToken.asset());
```
EndymionJkb, line 917:
```suggestion
```
Could move this down

pkg/vault/contracts/BatchRouter.sol
```diff
@@ -904,27 +910,27 @@ contract BatchRouter is IBatchRouter, BatchRouterStorage, RouterCommon, Reentran
...
                     wrappedToken: wrappedToken,
-                    amountGivenRaw: amountsOut[i],
+                    amountGivenRaw: wrappedAmountsOut[i],
                     limitRaw: params.minAmountsOut[i],
                     userData: params.userData
                 })
             );
 
-            _sendTokenOut(params.sender, underlyingToken, underlyingAmountsOut[i], params.wethIsEth);
+            if (isStaticCall == false) {
```
EndymionJkb, line 931:
```suggestion
            if (isStaticCall == false) {
                IERC20 underlyingToken = IERC20(wrappedToken.asset());
```
`isStaticCall` is fine where it is (outside the loop, so it only gets called once), but moving this down could save a little gas on queries.

pkg/vault/contracts/BatchRouter.sol
```diff
@@ -864,7 +864,7 @@ contract BatchRouter is IBatchRouter, BatchRouterStorage, RouterCommon, Reentran
 
     function addLiquidityERC4626PoolProportionalHook(
         AddLiquidityHookParams calldata params
-    ) external nonReentrant onlyVault returns (uint256[] memory amountsIn) {
+    ) external nonReentrant onlyVault returns (uint256[] memory underlyingAmountsIn) {
         IERC20[] memory erc4626PoolTokens = _vault.getPoolTokens(params.pool);
 
         uint256[] memory maxUint128TypeAmounts = new uint256[](erc4626PoolTokens.length);
```
EndymionJkb, line 870:
```suggestion
        uint256[] memory maxAmounts = new uint256[](erc4626PoolTokens.length);
```
Consider simplifying the name; people can look up _MAX_AMOUNT.

EndymionJkb (commented):
First pass. Need to write some docs for this; didn't want to start in case it's influenced by #892

pkg/vault/contracts/BatchRouter.sol
```diff
@@ -934,23 +940,33 @@ contract BatchRouter is IBatchRouter, BatchRouterStorage, RouterCommon, Reentran
...
         bool isStaticCall = EVMCallModeHelpers.isStaticCall();
 
-        // Wrap given underlying tokens for wrapped tokens
+        // Wrap given underlying tokens for wrapped tokens.
         for (uint256 i = 0; i < erc4626PoolTokens.length; ++i) {
-            // ERC4626 pool tokens are all wrapped tokens.
+            // Treat all ERC4626 pool tokens as wrapped. The next step will verify if we can use the wrappedToken as
+            // a valid ERC4626.
             IERC4626 wrappedToken = IERC4626(address(erc4626PoolTokens[i]));
             IERC20 underlyingToken = IERC20(wrappedToken.asset());
```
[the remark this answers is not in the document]
joaobrunoah: I've changed that in another PR: https://github.com/balancer/balancer-v3-monorepo/pull/892

pkg/vault/test/foundry/utils/BaseERC4626BufferTest.sol
```diff
@@ -30,32 +30,106 @@ abstract contract BaseERC4626BufferTest is BaseVaultTest {
     uint256 internal waUsdcIdx;
     address erc4626Pool;
 
+    // Rounding issues are introduced when dealing with tokens with rates different than 1:1. For example, to scale the
+    // tokens of an yield-bearing pool, the amount of tokens is multiplied by the rate of the token, which is
+    // calculated using `convertToAssets(FixedPoint.ONE)`. It generates an 18 decimal rate, but quantities bigger than
+    // 1e18 will have rounding issues. Another example is the different between convert (used to calculate query
+    // results of buffer operations) and the actual operation.
+    uint256 internal errorTolerance = 1e8;
```
EndymionJkb, line 38: Is this 1e-10%? (0.000000001%?) We haven't usually used assertApproxEqRel (usually assertApproxEqAbs), so consider renaming to show that this is a percentage (same as ours, where 1e18 = 100%, according to the docs).

```suggestion
    uint256 internal errorTolerancePercentage = 1e8; // 1e18 = 100%
```

pkg/vault/contracts/VaultAdmin.sol
```diff
@@ -614,6 +614,10 @@ contract VaultAdmin is IVaultAdmin, VaultCommon, Authentication, VaultGuard {
         return (_bufferTokenBalances[token].getBalanceRaw(), _bufferTokenBalances[token].getBalanceDerived());
     }
 
+    function isERC4626Initialized(IERC4626 token) external view returns (bool) {
```
EndymionJkb, line 617: This function doesn't seem to be used anywhere. Did you mean to add this to the external interface? (In which case it should be in IVaultAdmin.) Or is it left over?

EndymionJkb (approved):
Looks good to me now, though I think someone else should also review the tests; not sure I'm following everything there.

pkg/vault/contracts/VaultAdmin.sol
joaobrunoah: It's a left over, I'll remove it!

pkg/vault/test/foundry/utils/BaseERC4626BufferTest.sol
joaobrunoah: Since the BPT amount is also 18 decimals, the difference between Absolute and Relative is small. So, I just changed the test to Abs.

pkg/vault/contracts/BatchRouter.sol
```diff
@@ -960,6 +979,18 @@ contract BatchRouter is IBatchRouter, BatchRouterStorage, RouterCommon, Reentran
...
+
+            if (isStaticCall == false && kind == SwapKind.EXACT_OUT) {
+                // If SwapKind of wrap is EXACT_OUT, the limit of underlying tokens was taken from the user, so the
+                // difference between limit and exact underlying amount needs to be returned to the sender.
+                uint256 valueToSend;
+                unchecked {
+                    // `limits[i]` will always be bigger than `underlyingAmounts[i]` because
+                    // `erc4626BufferWrapOrUnwrap` checks it.
+                    valueToSend = limits[i] - underlyingAmounts[i];
+                }
```
jubeira, line 991: I think it's fine to use checked math here.

The comment is true, and even if it wasn't the transaction would revert. But i'd leave `unchecked` operations just for cases where it's blatantly obvious it won't over/underflow (i.e. `if (a > b) a - b`).

In this case the assumption is enforced in a separate contract, so I wouldn't say it's obvious just by looking at this code.

pkg/vault/test/foundry/BatchRouterERC4626Pool.t.sol
```diff
@@ -149,38 +146,40 @@ contract BatchRouterERC4626PoolTest is BaseERC4626BufferTest {
...
+                "Vault: wrong waDAI underlying buffer balance"
             );
         }
         {
-            (, , , uint256[] memory balances) = vault.getPoolTokenInfo(address(erc4626Pool));
-            assertEq(
-                balances[waDaiIdx],
-                erc4626PoolInitialAmount + exactWrappedAmountsIn[waDaiIdx],
-                "ERC4626 Pool: waDAI balance should increase"
+            (, , , uint256[] memory lastBalancesLiveScaled18) = vault.getPoolTokenInfo(address(erc4626Pool));
```
jubeira, line 158: hm why are we using live balances here? Shouldn't we be checking raw if we're comparing that against initial amounts / exact amounts in?

pkg/vault/test/foundry/utils/BaseERC4626BufferTest.sol
```diff
@@ -118,4 +192,30 @@ abstract contract BaseERC4626BufferTest is BaseVaultTest {
...
+        // Deposit sets the rate to 1.
+        vm.startPrank(lp);
+        dai.approve(address(waDAI), 10 * erc4626PoolInitialAmount);
+        waDAI.deposit(10 * erc4626PoolInitialAmount, lp);
+        usdc.approve(address(waUSDC), 10 * erc4626PoolInitialAmount);
+        waUSDC.deposit(10 * erc4626PoolInitialAmount, lp);
+        vm.stopPrank();
+        // Changing asset balances without minting shares changes the rate so that it is no longer 1.
+        dai.mint(address(waDAI), 2 * erc4626PoolInitialAmount);
+        usdc.mint(address(waUSDC), 4 * erc4626PoolInitialAmount);
```
jubeira, line 219: I know we're already converting back and forth between assets and shares, but can we add a very explicit assertion that checks that the rate of the wrappers is not FP(1)?

pkg/vault/test/foundry/YieldBearingPoolBufferAsVaultPrimitive.t.sol
```diff
@@ -360,40 +343,44 @@ contract YieldBearingPoolBufferAsVaultPrimitiveTest is BaseERC4626BufferTest {
...
-            assertApproxEqAbs(amounts[0], vars.expectedAliceDelta, 1, "Wrong amounts count");
+            assertApproxEqAbs(amounts[0], vars.expectedAliceDelta, MAX_ERROR, "Wrong amounts count");
             assertEq(tokens[0], address(usdc), "Wrong token for SwapKind");
         } else {
             // Rounding issues occurs in favor of vault, and are very small
-            assertGe(paths[0], vars.expectedAliceDelta, "paths AmountIn must be >= expected amountIn");
-            assertApproxEqAbs(paths[0], vars.expectedAliceDelta, 5, "Wrong path count");
-            assertGe(amounts[0], vars.expectedAliceDelta, "amounts AmountIn must be >= expected amountIn");
-            assertApproxEqAbs(amounts[0], vars.expectedAliceDelta, 5, "Wrong amounts count");
+            // TODO remove tolerance of 1 wei, when rounding issues with rates are solved
```
jubeira, line 352: Is this stale or is it a real todo?

jubeira (commented):
Thanks @joaobrunoah; great work.

Overall makes sense. I do have a few questions, and I'd like to do a second pass; there's a lot going on in here. But the direction seems correct in general.

pkg/vault/test/foundry/YieldBearingPoolBufferAsVaultPrimitive.t.sol
joaobrunoah: It's a real todo, related with the issue https://github.com/balancer/balancer-v3-monorepo/issues/884

pkg/vault/test/foundry/BatchRouterERC4626Pool.t.sol
joaobrunoah: You're right.. I just renamed it, forgot to check if it made sense. I'll change this.

pkg/vault/contracts/BatchRouter.sol
joaobrunoah: It saves 30 bytes of bytecode, but since we're going to split the BatchRouter anyway, it makes sense. I'll leave it checked then.

pkg/vault/test/foundry/BatchRouterERC4626Pool.t.sol
joaobrunoah: The test is correct. erc4626Pool has wrapped amounts, so the normal balance of the pool is wrapped balance. Since the tokens have 18 decimals, the lastBalancesLiveScaled18 return underlying amounts, which we can compare with the underlying amounts in.

jubeira (approved):
LGTM, great job @joaobrunoah !

One last final question: can we justify why the max error is 10 wei now? I understand each conversion is ok +/- 1 wei, so where does the 10 come from?

PR #886, opened 2024-08-15
Remove liquidity from nested pools

joaobrunoah:
# Description

This PR allows the liquidity of nested pools to be removed. Consider the scenario:

* PoolA has USDC and DAI;
* PoolB has PoolA-BPT and WETH;

Calling `removeLiquidityProportionalFromNestedPools`, using a PoolB-BPT exact amount, should return an amount of WETH, DAI and USDC.

## Type of change

- [ ] Bug fix 
- [x] New feature 
- [ ] Breaking change 
- [ ] Dependency changes
- [ ] Code refactor / cleanup
- [ ] Other

## Checklist:


pkg/solidity-utils/contracts/helpers/InputHelpers.sol
```diff
@@ -98,4 +98,30 @@ library InputHelpers {
...
+     *
+     * A bubble sort should be gas- and bytecode-efficient enough for such small arrays.
+     * Could have also done "manual" comparisons for each of the cases, but this is
+     * about the same number of operations, and more concise.
+     *
+     * This is less efficient for larger token count (i.e., above 4), but such pools should
+     * be rare. For now, sorting is only used to remove liquidity from nested pools in the
+     * batch router.
+     */
+    function sortAddresses(address[] memory addresses) internal pure returns (address[] memory) {
```
EndymionJkb, line 115: Can we just reuse the existing `sortTokens`? (Or generalize sortTokens to sortAddresses?)

pkg/vault/test/foundry/BatchRouterNestedPools.t.sol
```diff
@@ -0,0 +1,351 @@
...
+        // completely drained. We need to discount this amount of tokens from the total liquidity that we can extract
+        // from the child pools.
+        uint256 deadTokens = (MIN_BPT / 2).mulDown(proportionToRemove);
+
+        uint256[] memory expectedAmountsOut = new uint256[](4);
+        // DAI exists in childPoolB and parentPool, so we expect 2x more DAI than the other tokens.
+        // Since pools are in their initial state, we can use poolInitAmount as the balance of each token in the pool.
+        // Also, we only need to account deadTokens once, since we calculate the bpts in for the parent pool using
+        // totalSupply (so the burned MIN_BPT amount does not affect the bpt in calculation and the amounts out are
+        // perfectly proportional to the parent pool balance)
```
EndymionJkb, line 85:
```suggestion
        // Also, we only need to account for deadTokens once, since we calculate the BPT in for the parent pool using
        // totalSupply (so the burned MIN_BPT amount does not affect the BPT in circulation, and the amounts out are
        // perfectly proportional to the parent pool balance).
```
Is this what you mean?

pkg/vault/test/foundry/BatchRouterNestedPools.t.sol
```diff
@@ -0,0 +1,351 @@
...
+        );
+        assertEq(
+            vars.parentPoolAfter.childPoolBBpt,
+            vars.parentPoolBefore.childPoolBBpt - burnedChildPoolBBpts,
+            "ParentPool ChildPoolB BPT Balance is wrong"
+        );
+    }
+
+    function testRemoveLiquidityNestedPoolLimits() public {
+        // Remove 10% of pool liquidities.
```
EndymionJkb, line 225:
```suggestion
        // Remove 10% of pool liquidity.
```

pkg/vault/test/foundry/BatchRouterNestedPools.t.sol
```diff
@@ -0,0 +1,351 @@
...
+        minAmountsOut[vars.daiIdx] = poolInitAmount;
+        minAmountsOut[vars.wethIdx] = poolInitAmount;
+        minAmountsOut[vars.wstethIdx] = poolInitAmount;
+        minAmountsOut[vars.usdcIdx] = poolInitAmount;
+
+        // DAI exists in childPoolB and parentPool, so we expect 2x more DAI than the other tokens.
+        // Since pools are in their initial state, we can use poolInitAmount as the balance of each token in the pool.
+        // Also, we only need to account deadTokens once, since we calculate the bpts in for the parent pool using
+        // totalSupply (so the burned MIN_BPT amount does not affect the bpt in calculation and the amounts out are
+        // perfectly proportional to the parent pool balance)
```
EndymionJkb, line 251:
```suggestion
        // Also, we only need to account for deadTokens once, since we calculate the BPT in for the parent pool using
        // totalSupply (so the burned MIN_BPT amount does not affect the BPT in circulation, and the amounts out are
        // perfectly proportional to the parent pool balance).

```

pkg/vault/test/foundry/BatchRouterNestedPools.t.sol
```diff
@@ -0,0 +1,351 @@
...
+
+        NestedPoolTestLocals memory vars = _createNestedPoolTestLocals();
+
+        // During pool initialization, MIN_BPT amount of BPT is burned to address(0), so that the pool cannot be
+        // completely drained. We need to discount this amount of tokens from the total liquidity that we can extract
+        // from the child pools.
+        uint256 deadTokens = (MIN_BPT / 2).mulDown(proportionToRemove);
+
+        uint256[] memory minAmountsOut = new uint256[](4);
+        // Expect minAmountsOut to be the liquidity of the pool, which is more than what should return.
```
EndymionJkb, line 241:
```suggestion
        // Expect minAmountsOut to be the liquidity of the pool, which is more than what we should return,
        // causing it to revert.
```

EndymionJkb (commented):
Good to test repeated tokens - that will probably be common. If it was nested more than 1 level, it would just return the second-level BPT directly. I suppose we could make it recursive, allowing any depth (up to the gas limit). The expected out/limit calculations would get really complicated though: best to limit to 1 level for sanity.

pkg/solidity-utils/contracts/helpers/InputHelpers.sol
joaobrunoah: Done!

joaobrunoah:
> Will this be moved to a new router then, as discussed? (Given that it's 1.2k bytes over :)

Yes, but I think we said it should be done in another PR, because we will move Nested pools and ERC4626 pools too.

pkg/vault/contracts/CompositeLiquidityRouter.sol
```diff
@@ -0,0 +1,512 @@
...
+                from: params.sender,
+                maxBptAmountIn: params.maxBptAmountIn,
+                minAmountsOut: new uint256[](parentPoolTokens.length),
+                kind: params.kind,
+                userData: params.userData
+            })
+        );
+
+        for (uint256 i = 0; i < parentPoolTokens.length; i++) {
+            address childPool = address(parentPoolTokens[i]);
```
jubeira, line 430:
```suggestion
            address childToken = address(parentPoolTokens[i]);
```

Maybe this would be a bit clearer given that it's not necessarily a pool.

pkg/vault/contracts/CompositeLiquidityRouter.sol
```diff
@@ -0,0 +1,512 @@
...
+            } else {
+                // Token is not a BPT, so return the amount to the user.
+                _currentSwapTokensOut().add(childPool);
+                _currentSwapTokenOutAmounts().tAdd(childPool, parentPoolAmountsOut[i]);
+            }
+        }
+
+        // The hook writes current swap token and token amounts out.
+        // We copy that information to memory to return it before it is deleted during settlement.
+        tokensOut = InputHelpers.sortTokens(_currentSwapTokensOut().values().asIERC20()).asAddress();
```
jubeira, line 466: Maybe we don't need to sort the tokens out. Just returning the array of token addresses and output amounts should do.

pkg/vault/contracts/CompositeLiquidityRouter.sol
```diff
@@ -0,0 +1,512 @@
...
+        }
+
+        _settlePaths(params.sender, false);
+    }
+
+    /*******************************************************************************
+                                    Settlement
+    *******************************************************************************/
+
+    function _settlePaths(address sender, bool wethIsEth) internal {
```
jubeira, line 483: This is the same as with batch router right? Technically there are no 'paths' here.
We don't have a common layer between batch and composite router though :/ 

Do you think there's a good way of sharing this code between the routers?

pkg/vault/test/.contract-sizes/CompositeLiquidityRouter
```diff
@@ -0,0 +1,2 @@
+Bytecode	17.007
```
jubeira, line 1: :hushed:

pkg/vault/test/foundry/CompositeLiquidityRouterNestedPools.t.sol
```diff
@@ -0,0 +1,358 @@
...
+        // Since LP is the owner of all BPT supply, and part of the BPTs were burned in the initialization step, using
+        // totalSupply is more accurate to remove exactly the proportion that we intend from each pool.
+        uint256 exactBptIn = totalPoolBPTs.mulDown(proportionToRemove);
+
+        NestedPoolTestLocals memory vars = _createNestedPoolTestLocals();
+
+        // During pool initialization, POOL_MINIMUM_TOTAL_SUPPLY amount of BPT is burned to address(0), so that the
+        // pool cannot be completely drained. We need to discount this amount of tokens from the total liquidity that
+        // we can extract from the child pools.
+        uint256 deadTokens = (POOL_MINIMUM_TOTAL_SUPPLY / 2).mulDown(proportionToRemove);
```
jubeira, line 78: hm. This is correct, but if you don't want to deal with this another option is having `alice` add liquidity after initialization. Then, alice can remove all the added tokens, without any dead portion.

pkg/vault/test/foundry/CompositeLiquidityRouterNestedPools.t.sol
```diff
@@ -0,0 +1,358 @@
...
+        );
+        assertEq(
+            vars.childPoolAAfter.usdc,
+            vars.childPoolABefore.usdc - amountsOut[vars.usdcIdx],
+            "ChildPoolA Usdc Balance is wrong"
+        );
+
+        // Check ChildPoolB
+        // Since DAI amountOut comes from parentPool and childPoolB, we need to calculate the proportion that comes
+        // from childPoolB.
```
jubeira, line 191: :100: :fire:

jubeira (approved):
Just a few minor comments; otherwise LGTM. Great job!

Let's hold on merging until we decide whether to return the tokens in a sorted fashion or not.

pkg/vault/contracts/CompositeLiquidityRouter.sol
joaobrunoah: I thought about adding it to RouterCommon, but I wasn't sure since the Router doesn't use it. Anyway, I can add it in there.

pkg/vault/test/foundry/CompositeLiquidityRouterNestedPools.t.sol
joaobrunoah: I tried to replace `lp` by `alice`, and I still needed to deal with deadTokens because "removeLiquidityProportional" takes into account Alice's BPTs and `totalSupply`, which includes burned tokens.

EndymionJkb (commented):
Shaping up well; really took a bite out of BatchRouter :)

pkg/solidity-utils/test/foundry/utils/BaseTest.sol
```diff
@@ -138,6 +138,33 @@ abstract contract BaseTest is Test, GasSnapshot {
...
+        }
+
+        for (uint256 i = 0; i < length - 1; ++i) {
+            for (uint256 j = 0; j < length - 1; ++j) {
+                if (sortedAddresses[j] > sortedAddresses[j + 1]) {
+                    // Swap if they're out of order.
+                    (sortedAddresses[j], sortedAddresses[j + 1]) = (sortedAddresses[j + 1], sortedAddresses[j]);
+                }
+            }
+        }
```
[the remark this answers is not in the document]
joaobrunoah: Done!

pkg/vault/contracts/CompositeLiquidityRouter.sol
joaobrunoah: Done! As discussed, the user will pass tokensOut array, since he needs to know the token addresses beforehand to pass the minAmountsOut correctly.

pkg/vault/contracts/CompositeLiquidityRouter.sol
```diff
@@ -0,0 +1,472 @@
...
+                    direction: WrappingDirection.WRAP,
+                    wrappedToken: wrappedToken,
+                    amountGivenRaw: amountsIn[i],
+                    limitRaw: limits[i],
+                    userData: params.userData
+                })
+            );
+
+            if (isStaticCall == false && kind == SwapKind.EXACT_OUT) {
+                // If SwapKind of wrap is EXACT_OUT, the limit of underlying tokens was taken from the user, so the
```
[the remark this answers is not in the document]
joaobrunoah: wrap can be EXACT_IN or EXACT_OUT, and an EXACT_OUT SwapKind can be wrap or unwrap. I'd leave the wrap here, just to make sure it's clear.

pkg/vault/contracts/CompositeLiquidityRouter.sol
```diff
@@ -0,0 +1,472 @@
...
+
+                continue;
+            }
+
+            if (isStaticCall == false) {
+                if (kind == SwapKind.EXACT_IN) {
+                    // If SwapKind of wrap is EXACT_IN, take the exact amount in from the sender.
+                    _takeTokenIn(params.sender, underlyingToken, amountsIn[i], params.wethIsEth);
+                } else {
+                    // If SwapKind of wrap is EXACT_OUT, the exact amount in is not known, because amountsIn is the
```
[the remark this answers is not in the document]
joaobrunoah: wrap can be EXACT_IN or EXACT_OUT, and an EXACT_OUT SwapKind can be wrap or unwrap. I'd leave the wrap here, just to make sure it's clear.

PR #891, opened 2024-08-16
Add state influencing behavior/accuracy to dynamic data

EndymionJkb:
# Description

After discussions today, realized that the dynamic pool data can at times be misleading. For instance, if the pool is paused or non initialized, operations will fail. And if the pool is in Recovery Mode, live balances may be inaccurate. This adds documentation to that effect, and also adds the relevant pool config flags to the dynamic data output, so that users can check whether they can rely on the data.

I was going to update the tests, and realized we don't have any for this, I guess because it's an off-chain-only feature.

## Type of change

- [ ] Bug fix 
- [X] New feature 
- [ ] Breaking change 
- [ ] Dependency changes
- [ ] Code refactor / cleanup
- [ ] Other

## Checklist:

### DIFF
--- a/pkg/interfaces/contracts/pool-stable/IStablePool.sol
+++ b/pkg/interfaces/contracts/pool-stable/IStablePool.sol
@@ -22,6 +22,10 @@ struct StablePoolImmutableData {
 
 /**
  * @notice Snapshot of current Stable Pool data that can change.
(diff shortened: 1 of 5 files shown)


jubeira (approved):
That's a good point; thanks @EndymionJkb.

/cc @brunoguerios @johngrantuk

PR #892, opened 2024-08-16
Support add and remove liquidity to `partial ERC4626 pools`

joaobrunoah:
# Description

Adds support to add/remove liquidity in partial yield-bearing pools (Pools that have some but not all yield-bearing tokens)

## Type of change

- [ ] Bug fix 
- [x] New feature 
- [ ] Breaking change 
- [ ] Dependency changes
- [ ] Code refactor / cleanup
- [ ] Other

## Checklist:


EndymionJkb (approved):
We're eating up bytecode for the BatchRouter now; I hope all this stuff fits!

pkg/vault/contracts/BatchRouter.sol
```diff
@@ -951,7 +961,18 @@ contract BatchRouter is IBatchRouter, BatchRouterStorage, RouterCommon, Reentran
...
             IERC4626 wrappedToken = IERC4626(address(erc4626PoolTokens[i]));
-            IERC20 underlyingToken = IERC20(wrappedToken.asset());
+            IERC20 underlyingToken = IERC20(_vault.getBufferAsset(wrappedToken));
+
+            // If the vault returns address 0 as underlying, it means that the ERC4626 token buffer was not
+            // initialized. Thus, the router treats it as a non-ERC4626 token.
+            if (address(underlyingToken) == address(0)) {
+                if (isStaticCall == false) {
+                    underlyingAmounts[i] = amountsIn[i];
+                    wrappedAmounts[i] = amountsIn[i];
```
jubeira, line 971: Hmm don't we need to assign these two outside `if (isStaticCall == false)`?

jubeira: In a static call `underlyingAmounts[i]` and `wrappedAmounts[i]` will be 0.
In `addLiquidityERC4626PoolUnbalancedHook` we use those as max input amounts to add liquidity.

I don't get why no tests are failing because of this, since the query would fail when trying to add liquidity with 0 max inputs here: https://github.com/balancer/balancer-v3-monorepo/pull/892/files#diff-8c740872f6bae09a05c4cdf1ba7a2c0cae6ca40cf8ae0b14058d20e624fb14deR195

jubeira (commented):
Thanks @joaobrunoah, looks good and makes sense! Simple and on-point.

I have a question about `_wrapTokens`. I don't fully get why the test doesn't fail because of this, so I might be missing something. Could you please check it out before merging?

pkg/vault/contracts/BatchRouter.sol
joaobrunoah: Good catch! I created query tests for this now.

jubeira (approved):
LGTM, great job @joaobrunoah ! :100:

PR #893, opened 2024-08-16
Make `removeLiquidityFromBuffer` permissionless

jubeira:
# Description

In the current state, `removeLiquidityFromBuffer` relies on a permission for the Vault to get the rightful owner of the buffer shares.

Even though governance should be well behaved, this opens a vector to lock and even drain buffer funds if the permission is misused, which is far from ideal.

To get the rightful owner, we can instead make the Vault the entrypoint and send the correct sender to the function that executes the operation instead. This can be thought of as the Vault acting as a Router.

Regarding the remaining modifier modifications:
- The new internal hook can only be called by the Vault itself, otherwise anyone could just unlock the Vault and spoof the owner anyways.
- The new hook can't be non-reentrant because of the call to `sendTo` which is already non-reentrant. But given that `sendTo` is the only bit that makes external calls, it should still be non-reentrant as a whole in practice.
  - We could refactor the Vault and make `sendTo` public inside `VaultCommon`, but we don't really need to optimize this case. Calling both `supplyCredit` and `sendTo` from the vault itself is a pattern that we haven't used before, and is a consequence of the code acting both as the Vault and the Router. But still prefer this option instead of refactoring `sendTo`: using it from the vault should only be done with extreme care.

## Type of change

- [ ] Bug fix 
- [x] New feature 
- [ ] Breaking change 
- [ ] Dependency changes
- [ ] Code refactor / cleanup
- [ ] Other

## Checklist:

## Issue Resolution

N/A


pkg/vault/contracts/VaultAdmin.sol
```diff
@@ -556,7 +582,8 @@ contract VaultAdmin is IVaultAdmin, VaultCommon, Authentication {
         removedUnderlyingBalanceRaw = (bufferBalances.getBalanceRaw() * sharesToRemove) / totalShares;
         removedWrappedBalanceRaw = (bufferBalances.getBalanceDerived() * sharesToRemove) / totalShares;
 
-        _supplyCredit(IERC20(_bufferAssets[wrappedToken]), removedUnderlyingBalanceRaw);
+        IERC20 underlyingToken = IERC20(_bufferAssets[wrappedToken]);
```
EndymionJkb, line 585: Consider adding a comment here about why we're not calling `asset` on the wrapped token. We've guaranteed the asset on the way in, so we can trust the Vault's token assignment on the way out.

EndymionJkb (approved):
Weird pattern to be sure, but well worth it for improving the security guarantees around buffers. They're much more likely to be funded when they are fully non-custodial.

Great how all the existing tests work; just need to s/router/vault/g;

pkg/vault/test/foundry/mutation/vault/VaultAdmin.t.sol
```diff
@@ -322,41 +323,47 @@ contract VaultAdminMutationTest is BaseVaultTest {
...
         vm.expectRevert(IVaultErrors.NotVaultDelegateCall.selector);
-        vaultAdmin.removeLiquidityFromBuffer(IERC4626(address(0)), 0, address(0));
+        VaultAdmin(address(vaultAdmin)).removeLiquidityFromBufferHook(IERC4626(address(0)), 0, address(0));
     }
 
-    function testRemoveLiquidityFromBufferWhenNotUnlocked() public {
+    function testRemoveLiquidityFromBufferHookWhenVaultIsNotSender() public {
+        vm.expectRevert(abi.encodeWithSelector(IVaultErrors.SenderIsNotVault.selector, address(this)));
+        VaultAdmin(address(vault)).removeLiquidityFromBufferHook(IERC4626(address(0)), 0, address(0));
+    }
```
EndymionJkb, line 334:
```suggestion
```
Same test as `testRemoveLiquidityFromBufferHookWhenNotVault`

joaobrunoah (approved):
LGTM!

pkg/vault/test/foundry/mutation/vault/VaultAdmin.t.sol
jubeira: it's not :)
- `msg.sender` has to be the Vault
- current address has to be vault (i.e. be a vault delegate call, since the hook actually lives in the admin)

jubeira: In other words, this one checks `onlyVault` modifier, and the other one checks `onlyVaultDelegateCall` modifier.

Not having `onlyVault` as modifier is a critical bug, as it allows anyone to unlock the vault and call `remove` hook, spoofing the sender.

pkg/interfaces/contracts/vault/IVaultAdmin.sol
```diff
@@ -247,14 +248,12 @@ interface IVaultAdmin {
      * @param wrappedToken Address of the wrapped token that implements IERC4626
      * @param sharesToRemove Amount of shares to remove from the buffer. Cannot be greater than sharesOwner's
      * total shares
-     * @param sharesOwner Address that owns the deposited liquidity.
      * @return removedUnderlyingBalanceRaw Amount of underlying tokens returned to the user
      * @return removedWrappedBalanceRaw Amount of wrapped tokens returned to the user
      */
```
EndymionJkb, line 253: There are some stale comments in IVaultAdmin, such as initializeBuffer and addLiquidityToBuffer, where it says "...an yield-bearing buffer (one of the Vault's internal ERC4626 buffers)"

Should be:
"an internal ERC4626 buffer in the Vault"

pkg/vault/test/foundry/mutation/vault/VaultAdmin.t.sol
jubeira: Ah I see what you mean. Fixed now.

Issue #894, opened 2024-08-18
State Sharing Issue Between `onComputeDynamicSwapFeePercentage` and `onSwap`


andreiashu:
While adapting the `FXPool`'s fee calculation to Balancer V3, I've identified a limitation in the current architecture:  `onComputeDynamicSwapFeePercentage` and `onSwap` functions do not share state.

This forces redundant calculations in both functions, particularly for complex fee structures like FXPool's, where fees depend on the swap amount and its effect on pool ratios.

The current dynamic swap fee percentage mechanism only supports basic rules due to this lack of state sharing.

I think that allowing the Vault to pass mutable state data between these functions could solve the immediate issue, but it may not be the most elegant solution.

Are there more efficient or cleaner approaches to handle complex, interdependent fee calculations in Balancer V3?

andreiashu:
I'd like to add more context here: in the previous version of the FXPool (Bal V2) we would mint LP tokens as a proportion of the fee numeraire (equivalent value in USD). Minting LP tokens from a pool doesn't seem to be possible outside of the `initialize` or `addLiquidity` flow.

Hence the need to use `onComputeDynamicSwapFeePercentage` and the issue I stumbled on in the original description of this ticket

jubeira:
Thanks for bringing this up @andreiashu.

`onComputeDynamicSwapFeePercentage` is designed to be simple and specific, which is why it's just a `view` function.
`onBeforeSwap` on the contrary is more generic: you can save state there if needed (transient storage might be a good idea if it's just to store an intermediary result) and even reenter the vault and trigger another operation (e.g. `addLiquidity`). That would allow you to mint LP tokens before swapping for example (or anything that you want; could also be even another swap).

> This forces redundant calculations in both functions, particularly for complex fee structures like FXPool's, where fees depend on the swap amount and its effect on pool ratios.

By any chance is there some draft we can look at? Is this redundancy overly complex or expensive in terms of gas costs?

> I think that allowing the Vault to pass mutable state data between these functions could solve the immediate issue, but it may not be the most elegant solution.

This is probably the case, although in terms of architecture a hook to compute a dynamic fee percentage passing data to the main swap hook sounds a bit weird to me.

andreiashu:
@jubeira thanks for the suggestion, was focusing so much on trying to make `onComputeDynamicSwapFeePercentage` work for us that I forgot about `onBeforeSwap`.

[...]

PR #895, opened 2024-08-18
Rename swap params to improve devUX

EndymionJkb:
# Description

This always drove me a little nuts. And when thinking about issue #894, I realized it was difficult to understand the current state by just looking at the code.

There are (for good reasons) two separate swap parameter structs that we pass around. (The passing around of memory structs is not ideal, but necessary due to stack-too-deep limitations.)

One corresponds to the "external" API from the router: with token addresses and amounts in native token decimals. The other is the "internal" API used within the Vault (corresponding to `IBasePool`); e.g., to enforce consistency when calling the main operation and various hooks. This uses scaled values, indices instead of addresses, etc.

I've always found this a bit hard to read. It's confusing to have `swapParams` and `poolSwapParams`, especially when it's sometimes called `params`, and we also have `SwapState` with completely generic variables like `state`. It's especially confusing when we're converting between types.

This PR adds a block comment explaining why we have two "SwapParam" things, and what each does. It renames SwapParams to VaultSwapParams, so that we now have VaultSwapParams for the "Vault API" and PoolSwapParams for the "Pool/Hook API." It also explicitly names all the generic `params` and `state` variables so that we can keep them all straight.

As for the choice of names, I suppose VaultSwapParams could be RouterSwapParams; I just thought that might be confusing in the Vault. We could also say ExternalSwapParams and InternalSwapParams, but these aren't very specific. PoolSwapParams was already there, and it does mean "api for IBasePool and pool hooks," so I left it alone.

## Type of change

- [ ] Bug fix 
- [ ] New feature 
- [ ] Breaking change 
- [ ] Dependency changes
- [X] Code refactor / cleanup
- [ ] Other

## Checklist:


jubeira (commented):
I think this makes sense for swap given that we have pool / vault swap params coexisting very closely together in the code. The change itself looks good.

@danielmkm do you think this improves readability a bit?

pkg/interfaces/contracts/vault/VaultTypes.sol
```diff
@@ -196,17 +196,26 @@ enum SwapKind {
...
 }
 
+// There are two "SwapParams" structs defined below. `VaultSwapParams` corresponds to the external swap API defined
+// in the Router contracts, which uses explicit token addresses, the amount given and limit on the calculated amount
+// expressed in native token decimals, and optional user data passed in from the caller.
+//
+// `PoolSwapParams` passes some of this information through (kind, userData), but "translates" the parameters to fit
+// the internal swap API used by `IBasePool`. It scales amounts to full 18-decimal precision, adds the token balances,
+// converts the raw token addresses to indices, and adds the address of the router originating the request. It does
+// not need the limit, since this is checked at the Router level.
```
joaobrunoah, line 206: Nice!

joaobrunoah (approved):
Makes sense!

jubeira (approved):
Let's go!

PR #897, opened 2024-08-18
Add helper to set protocol fees by factory

### LINKED ISSUE #888: Set protocol fee percentage for a given pool factory
Add the capability to set protocol fee % per factory instead of per pool.

Should be done in a layer above the fee controller and the current system in place, with a contract that has permission to tweak it.

EndymionJkb:
# Description

This adds a `ProtocolFeePercentagesProvider` helper contract that allows governance to set protocol fee percentages at the factory level, and then a permissionless function to propagate these fee percentages to individual pools. (Well, permissionless after governance grants the provider contract permission to set pool fees.)

An important design constraint was building this on top of the existing infrastructure, with no new dependencies (e.g., not changing existing factories, or adding new requirements for factories, such as keeping an enumerable set of pools, or calling back into anything to register new pools, etc.)

There are no changes to existing contracts: except adding a getter for the maximum swap/yield fee constants to the ProtocolFeeController, which it arguably should have anyway. This avoids duplicating constants, and divergence from the current ProtocolFeeController.

It has two important functions:

`setFactorySpecificProtocolFeePercentages`: a permissioned call to set fee percentages by factory. It does some validation that isn't strictly necessary: ensuring that the fees being set are <= max allowed values, and a "best effort" check that the factory is a real factory (i.e., that responds correctly to `isPoolFromFactory` calls). Otherwise, you could set values that would fail when you tried to apply them.

`setProtocolFeePercentagesForPools`: a permissionless call to apply the fee percentages to a set of pools from a given factory. This was over-engineered at first (as I tend to do), taking just a set of pools, and having the contract figure out which factories they belonged to and fetch the corresponding fee percentages. But that required an enumerable set of factories, a "find" function, etc. Probably people are going to set the fees for one factory at a time, using the subgraph, so I think making it completely generic is overkill.

It should cut down on governance calls, since they need only approve the factory-wide defaults, not the list of individual pools. No further governance action would be needed for new pools from that factory, as the "propagation" call is permissionless.

## Type of change

- [ ] Bug fix 
- [X] New feature 
- [ ] Breaking change 
- [ ] Dependency changes
- [ ] Code refactor / cleanup
- [ ] Other

## Checklist:

## Issue Resolution

Resolves #888


pkg/vault/contracts/ProtocolFeePercentagesProvider.sol
```diff
@@ -0,0 +1,139 @@
...
+
+        // Best effort check that `factory` is the address of an IBasePoolFactory.
+        bool poolFromFactory = IBasePoolFactory(factory).isPoolFromFactory(address(0));
+        if (poolFromFactory) {
+            revert InvalidFactory(factory);
+        }
+
+        // Store the default fee percentages, and mark the factory as registered.
+        _factoryDefaultFeePercentages[IBasePoolFactory(factory)] = FactoryProtocolFees({
+            protocolSwapFeePercentage: protocolSwapFeePercentage.toUint64(),
```
EndymionJkb, line 96: Technically don't need SafeCast here, as I've validated the range above; could just cast it. But we don't care too much about gas here.

pkg/vault/contracts/ProtocolFeePercentagesProvider.sol
```diff
@@ -0,0 +1,143 @@
...
+    }
+
+    /// @inheritdoc IProtocolFeePercentagesProvider
+    function setProtocolFeePercentagesForPools(address factory, address[] memory pools) external {
+        FactoryProtocolFees memory factoryFees = _getValidatedProtocolFees(factory);
+
+        for (uint256 i = 0; i < pools.length; ++i) {
+            address currentPool = pools[i];
+
+            if (IBasePoolFactory(factory).isPoolFromFactory(currentPool) == false) {
```
jubeira, line 113: Technically if you set the fee percentage in a contract that just returns `true` for any address that is not 0, you can use that 'factory' to set any pool fees.

I think there's not much way around it. I don't see much value in splitting `setFactorySpecificProtocolFeePercentages` into 2 steps (one to register the factory, and another one to set the value) as in practice it'd be pretty much the same.

jubeira (commented):
First pass done. I think it makes sense overall; thanks @EndymionJkb !

I can't think of a way to validate that the factory is actually the factory without any extra requirements to the pool themselves, so this is probably fine. I'll give it some more thought just in case before the second pass.

elshan-eth (dismissed):
LGTM!

Grossbeatz:
Use `git status` to list all new or modified files that haven't yet been committed.

pkg/vault/contracts/ProtocolFeePercentagesProvider.sol
```diff
@@ -0,0 +1,149 @@
...
+        _protocolFeeController = protocolFeeController;
+        _trustedContractRegistry = trustedContractRegistry;
+
+        if (protocolFeeController.vault() != vault) {
+            revert WrongProtocolFeeControllerDeployment();
+        }
+
+        // These values are constant in the `ProtocolFeeController`.
+        (_maxProtocolSwapFeePercentage, _maxProtocolYieldFeePercentage) = protocolFeeController
+            .getMaximumProtocolFeePercentages();
```
jubeira, line 60: Unfortunately I don't think we'll migrate this once more in the short term. I'd just receive the constants and verify that they are correct in the fork test.

pkg/vault/contracts/ProtocolFeePercentagesProvider.sol
```diff
@@ -0,0 +1,149 @@
...
+            revert IProtocolFeeController.ProtocolSwapFeePercentageTooHigh();
+        }
+
+        if (protocolYieldFeePercentage > _maxProtocolYieldFeePercentage) {
+            revert IProtocolFeeController.ProtocolYieldFeePercentageTooHigh();
+        }
+
+        // Ensure precision checks will pass.
+        _protocolFeeController.ensureValidPrecision(protocolSwapFeePercentage);
+        _protocolFeeController.ensureValidPrecision(protocolYieldFeePercentage);
```
jubeira, line 95: The precision is fixed in the vault, so we can also replicate the logic in this contract. We don't necessarily need to rely on the fee controller for that now that the deployed version doesn't really have this public function.

pkg/interfaces/contracts/vault/IProtocolFeeController.sol
jubeira, line 1: This is fine I guess, although I don't think we'll do another migration for this.

pkg/vault/contracts/ProtocolFeePercentagesProvider.sol
jubeira: Let's add a small comment here, saying that the factory has been validated by the registry when you call `setFactorySpecificProtocolFeePercentages`, and then checked against the argument at the beginning of this function

pkg/vault/test/foundry/ProtocolFeePercentagesProvider.t.sol
```diff
@@ -0,0 +1,285 @@
...
+            poolFactory,
+            maxSwapFeePercentage,
+            maxYieldFeePercentage
+        );
+    }
+
+    function testRevertWhenSetFactorySpecificProtocolFeePercentageInvalidFactory() public {
+        _grantPermissions();
+
+        vm.expectRevert();
```
jubeira, line 104: Let's be specific with the revert reasons in the tests.

pkg/vault/test/foundry/ProtocolFeePercentagesProvider.t.sol
```diff
@@ -0,0 +1,285 @@
...
+
+    function testSetFactorySpecificProtocolFeePercentageHighPrecisionYield() public {
+        _grantPermissions();
+
+        vm.expectRevert(IProtocolFeeController.ProtocolYieldFeePercentageTooHigh.selector);
+        vm.prank(admin);
+        percentagesProvider.setFactorySpecificProtocolFeePercentages(
+            poolFactory,
+            1e16 + 234234234,
+            maxYieldFeePercentage + 1
```
jubeira, line 175:
```suggestion
            maxSwapFeePercentage,
            1e16 + 234234234
```

pkg/vault/test/foundry/ProtocolFeePercentagesProvider.t.sol
```diff
@@ -0,0 +1,285 @@
...
+            poolFactory,
+            maxSwapFeePercentage,
+            maxYieldFeePercentage + 1
+        );
+    }
+
+    function testSetFactorySpecificProtocolFeePercentageHighPrecisionYield() public {
+        _grantPermissions();
+
+        vm.expectRevert(IProtocolFeeController.ProtocolYieldFeePercentageTooHigh.selector);
```
jubeira, line 170: Test name doesn't match what it does :)

```suggestion
        vm.expectRevert(IProtocolFeeController.FeePrecisionTooHigh.selector);
```

jubeira (commented):
"better late than never" huh

This is looking good. Just a few comments, mostly to adjust to the deployed contracts. Otherwise this makes a lot of sense!

pkg/vault/contracts/ProtocolFeePercentagesProvider.sol
EndymionJkb: They're public, so I can still get them; just not with this function. I think to avoid confusion I will not make any changes to the controller in this PR. (We've already changed it since V2, but not in this way.) Just what happens when PRs languish too long.

pkg/interfaces/contracts/vault/IProtocolFeeController.sol
EndymionJkb: I think it's simplest to just revert all the protocol fee controller changes and work with what's available in the deployed version.

jubeira: agreed

pkg/vault/test/foundry/ProtocolFeePercentagesProvider.t.sol
EndymionJkb: Not sure why I didn't do that. Will add this to all of them.

jubeira (approved):
LGTM, awesome job!

PR #899, opened 2024-08-19
Make initialize non-reentrant

### LINKED ISSUE #898: Make `onInitialize` nonReentrant
There's no real use case to reenter before initialize that we can think of.

EndymionJkb:
# Description

There is possible reentrancy if we allow the pre-initialize hook to re-enter the Vault (before setting the initialized bit). Since we don't think the pre-initialize hook needs to be reentrant in the first place, just make the whole initialize function nonReentrant and avoid the issue.

## Type of change

- [ ] Bug fix 
- [ ] New feature 
- [ ] Breaking change 
- [ ] Dependency changes
- [X] Code refactor / cleanup
- [ ] Other

## Checklist:

## Issue Resolution

Resolves #898


jubeira (approved):
LGTM.

Can we add a test inside `mutation/VaultExtension` that double-checks that the modifier is present?

PR #908, opened 2024-08-20
Enforce memory-safe checks on assembly blocks

### LINKED ISSUE #906: Add `memory-safe` to assembly blocks
When you add the "memory-safe" option to an inline assembly block, it tells the Solidity compiler that the code within the assembly block adheres to certain safety rules regarding memory access. Specifically, it enforces the following:

**Bounds Checking**: The compiler ensures that memory operations (such as mload, mstore, and mstore8) do not access memory outside of the allocated bounds. This prevents memory corruption by ensuring that you only read from or write to valid memory locations.

**Memory Safety**: The compiler checks that any changes to the mload, mstore, or mstore8 instructions within the assembly block are safe and do not introduce vulnerabilities related to memory management.

**Stack and Memory Consistency**: The "memory-safe" option helps the compiler ensure that the stack and memory states are consistent and valid throughout the assembly block, reducing the likelihood of introducing bugs that could lead to unexpected behavior.

Note that there is also a directive (/// @solidity memory-safe-assembly), which is used in Create3 and the OZ libraries. This "asserts" that it's memory safe (e.g., to suppress warnings), but doesn't actively check. The "memory-safe" actually performs checks at compile time to ensure memory safety.

EndymionJkb:
# Description

Assembly code is always a bit dangerous; this makes it less so by adding directives to assembly blocks to enforce safe use of memory (e.g., that it doesn't overflow during memory copies). There is also a compiler hint used in OZ and create3 (/// @solidity memory-safe-assembly), which "asserts" that the memory is safe, but doesn't actually check. I've left that alone in imported libraries, and only changed it for our own code.

There was some talk about this affecting optimization (even though it's done at compile time), but it doesn't seem to have changed anything in that respect.

## Type of change

- [ ] Bug fix 
- [ ] New feature 
- [ ] Breaking change 
- [ ] Dependency changes
- [X] Code refactor / cleanup
- [ ] Other

## Checklist:

## Issue Resolution

Resolves #906


pkg/solidity-utils/contracts/helpers/CastingHelpers.sol
```diff
@@ -23,7 +23,7 @@ library CastingHelpers {
     /// @dev Returns a native array of addresses as an IERC20[] array. (Used in tests.)
     function asIERC20(address[] memory addresses) internal pure returns (IERC20[] memory tokens) {
         // solhint-disable-next-line no-inline-assembly
-        assembly {
+        assembly ("memory-safe") {
```
joaobrunoah, line 26: I just passed through a Stack-too-deep issue today because this function was missing "memory-safe". Cool!

joaobrunoah (approved):
LGTM!

PR #915, opened 2024-08-20
Dynamic swap fee guardrail

### LINKED ISSUE #914: Dynamic swap fee range check
Per TOB-BALV3-10: The dynamic swap fee is not explicitly limited to 100% which could lead to unexpected behavior.

EndymionJkb:
# Description

Limit the dynamic swap fee percentage to 100%, regardless of what the hook returns, to avoid unexpected behavior in downstream code. We already have these limits when setting fees (including calculating aggregate fees), but not on dynamic fees.

The custom error `InvalidPercentage` was local to the PoolConfigLib. Since we also need it in the HooksConfigLib, I moved it to IVaultErrors. Also removed the argument - do we really care about the exact value? - because this mitigates the increase in bytecode from adding the check. We get a few bytes back by removing that. It still fits with it in, so if we think it's necessary after all, we can add it back in.

## Type of change

- [X] Bug fix 
- [ ] New feature 
- [ ] Breaking change 
- [ ] Dependency changes
- [ ] Code refactor / cleanup
- [ ] Other

## Checklist:

## Issue Resolution

Resolves #914


pkg/interfaces/contracts/vault/IVaultErrors.sol
```diff
@@ -215,6 +215,9 @@ interface IVaultErrors {
      */
     error FeePrecisionTooHigh();
 
+    /// @dev Given percentage is above FixedPoint.ONE (1e18 wei).
+    error InvalidPercentage();
```
joaobrunoah, line 219: Should it be renamed to something that explains that the percentage was greater than 100%? Like, PercentageAboveMax, or something like this?

It's minor, the comment explains, but just a suggestion, since this error is only used when the percentage is bigger than a maximum number.

joaobrunoah (approved):
LGTM!

pkg/interfaces/contracts/vault/IVaultErrors.sol
EndymionJkb: Yes; currently it's always 100% (hence "Invalid" meaning outside 0-100%). However, in the code there is a check against `MAX_FEE_PERCENTAGE`, which could possibly be set < 100% at some point (e.g., for dynamic fees). So it makes sense to change.

PR #922, opened 2024-08-23
Add rounding direction to `computeInvariant`

jubeira:
# Description

To ensure path independency, we must ensure that any pair of do-undo operations don't result in funds getting extracted for free.

For unbalanced liquidity operations, we're using `computeInvariant` / `computeBalance` and doing math around them in `BasePoolMath` to solve BPT amounts in and / or out. 

While `BasePoolMath` already rounds in the right direction, the invariant computation itself can also have precision loss, which is unknown to the Vault. Therefore, rounding only in `BasePoolMath` might not be enough to offset this error for every circumstance.

Particularly for `WeightedPool`, the invariant calculation requires `pow` calls, which has errors much more significant than a few wei. The solution to prevent this from breaking path independence is to provide a rounding direction for the invariant. Then, pools can implement two internal variations for `computeInvariant`, and `BasePoolMath` will use them accordingly to round in the protocol's favor.

On the other hand, I've also analyzed the case of `computeBalance`. Considering how it is used in `BasePoolMath`, my conclusion is that it doesn't need a rounding direction: it should always round up, and that works in the protocol's favor (note for the reviewers: please verify that this is effectively the case).

Other notes:
- The only code change besides the rounding direction is the minimum swap fee for weighted pools, increased 10x based on fuzz tests.
- The liquidity approximation tests have been tightened up. In the current tests, the error between direct / indirect swaps is measured symmetrically around the result of the direct method. Truth is that it's fine to deviate a bit more penalizing indirect swaps, but we definitely don't want the indirect method to be better. There's still some tolerance, but now we check that whenever that happens, it's effectively a minor discount over the swap fees paid. In other words, the discount is not enough to make the operation reversible in any case.

## Type of change

- [x] Bug fix 
- [ ] New feature 
- [x] Breaking change 
- [ ] Dependency changes
- [ ] Code refactor / cleanup
- [ ] Other

## Checklist:

## Issue Resolution

N/A


pkg/interfaces/contracts/vault/IBasePool.sol
```diff
@@ -21,14 +21,23 @@ interface IBasePool is ISwapFeePercentageBounds, IUnbalancedLiquidityInvariantRa
     /**
      * @notice Computes the pool's invariant.
      * @dev This function computes the invariant based on current balances (and potentially other pool state).
+     * The given rounding direction must be respected for the Vault to round in its favor when calling this function.
```
joaobrunoah, line 24:
```suggestion
     * The given rounding direction must be respected for the Vault to round in pool's favor when calling this function.
```

I'd put in pool's favor, since we need the developers of the pool to understand why the rounding is important.

pkg/pool-weighted/contracts/WeightedPool.sol
```diff
@@ -13,7 +13,12 @@ import {
 } from "@balancer-labs/v3-interfaces/contracts/vault/IUnbalancedLiquidityInvariantRatioBounds.sol";
 import { IVault } from "@balancer-labs/v3-interfaces/contracts/vault/IVault.sol";
 import { IVaultErrors } from "@balancer-labs/v3-interfaces/contracts/vault/IVaultErrors.sol";
-import { SwapKind, PoolSwapParams, PoolConfig } from "@balancer-labs/v3-interfaces/contracts/vault/VaultTypes.sol";
+import {
+    SwapKind,
+    PoolSwapParams,
+    PoolConfig,
+    Rounding
+} from "@balancer-labs/v3-interfaces/contracts/vault/VaultTypes.sol";
```
joaobrunoah, line 21:
```suggestion
import "@balancer-labs/v3-interfaces/contracts/vault/VaultTypes.sol";
```

Since now it's multiline import, maybe we can leave the import of the file only?

pkg/pool-weighted/contracts/WeightedPool.sol
```diff
@@ -38,7 +43,7 @@ contract WeightedPool is IWeightedPool, BalancerPoolToken, PoolInfo, Version {
     // This means they have 0.00001% resolution (i.e., any non-zero bits < 1e11 will cause precision loss).
     // Minimum values help make the math well-behaved (i.e., the swap fee should overwhelm any rounding error).
     // Maximum values protect users by preventing permissioned actors from setting excessively high swap fees.
-    uint256 private constant _MIN_SWAP_FEE_PERCENTAGE = 1e12; // 0.0001%
+    uint256 private constant _MIN_SWAP_FEE_PERCENTAGE = 1e13; // 0.001%
```
joaobrunoah, line 46:
```suggestion
    uint256 private constant _MIN_SWAP_FEE_PERCENTAGE = 0.001e16; // 0.001%
```

pkg/pool-weighted/test/foundry/BigWeightedPool.t.sol
```diff
@@ -34,6 +34,9 @@ contract BigWeightedPoolTest is BasePoolTest {
         tokenIndexOut = 7;
 
         BasePoolTest.setUp();
+
+        poolMinSwapFeePercentage = 1e13;
```
joaobrunoah, line 38:
```suggestion
        poolMinSwapFeePercentage = 0.001e16; // 0.001%
```

pkg/pool-weighted/test/foundry/LiquidityApproximationWeighted.t.sol
```diff
@@ -74,7 +78,7 @@ contract LiquidityApproximationWeightedTest is LiquidityApproximationTest {
     ) public {
         // Weights can introduce some differences in the swap fees calculated by the pool during unbalanced add/remove
         // liquidity, so the error tolerance needs to be a bit higher than the default tolerance.
-        roundingDelta = 1e13;
+        excessRoundingDelta = 0.5e16; // 0.1%
```
joaobrunoah, line 81:
```suggestion
        excessRoundingDelta = 0.5e16; // 0.5%
```

pkg/pool-weighted/test/foundry/WeightedPool.t.sol
```diff
@@ -40,6 +41,9 @@ contract WeightedPoolTest is BasePoolTest {
         BasePoolTest.setUp();
 
         (daiIdx, usdcIdx) = getSortedIndexes(address(dai), address(usdc));
+
+        poolMinSwapFeePercentage = 1e13;
```
joaobrunoah, line 45:
```suggestion
        poolMinSwapFeePercentage = 0.001e16; // 0.001%
```

pkg/pool-weighted/test/foundry/WeightedPoolLimits.t.sol
```diff
@@ -7,7 +7,11 @@ import "forge-std/Test.sol";
 import { IERC20 } from "@openzeppelin/contracts/token/ERC20/IERC20.sol";
 
 import { IVaultAdmin } from "@balancer-labs/v3-interfaces/contracts/vault/IVaultAdmin.sol";
-import { PoolRoleAccounts, LiquidityManagement } from "@balancer-labs/v3-interfaces/contracts/vault/VaultTypes.sol";
+import {
+    PoolRoleAccounts,
+    LiquidityManagement,
+    Rounding
+} from "@balancer-labs/v3-interfaces/contracts/vault/VaultTypes.sol";
```
joaobrunoah, line 14:
```suggestion
import "@balancer-labs/v3-interfaces/contracts/vault/VaultTypes.sol";
```

pkg/pool-weighted/test/foundry/WeightedPoolLimits.t.sol
```diff
@@ -29,6 +33,8 @@ contract WeightedPoolLimitsTest is BaseVaultTest {
     using ArrayHelpers for *;
 
     uint256 constant DEFAULT_SWAP_FEE = 1e16; // 1%
+    uint256 constant MIN_SWAP_FEE = 1e13;
```
joaobrunoah, line 36:
```suggestion
    uint256 constant MIN_SWAP_FEE = 0.001e16; // 0.001%
```

joaobrunoah (approved):
LGTM! I checked computeBalance, the function already receives invariantRatio, which is simplified by dividing the newSupply by totalSupply, rounded up, so it doesn't look like it should have a rounding direction, as you mentioned. I also checked the rounding drection in add and remove liquidity, since both were rounding up, but it makes sense to round up in both cases, so all good.

pkg/interfaces/contracts/vault/IBasePool.sol
```diff
@@ -21,14 +21,23 @@ interface IBasePool is ISwapFeePercentageBounds, IUnbalancedLiquidityInvariantRa
     /**
      * @notice Computes the pool's invariant.
      * @dev This function computes the invariant based on current balances (and potentially other pool state).
+     * The given rounding direction must be respected for the Vault to round in pool's favor when calling this function.
```
EndymionJkb, line 24:
```suggestion
     * The rounding direction must be respected for the Vault to round in the pool's favor when calling this function.
```

pkg/pool-stable/contracts/StablePool.sol
```diff
@@ -14,7 +14,12 @@ import {
     IUnbalancedLiquidityInvariantRatioBounds
 } from "@balancer-labs/v3-interfaces/contracts/vault/IUnbalancedLiquidityInvariantRatioBounds.sol";
 import { IVault } from "@balancer-labs/v3-interfaces/contracts/vault/IVault.sol";
-import { SwapKind, PoolSwapParams, PoolConfig } from "@balancer-labs/v3-interfaces/contracts/vault/VaultTypes.sol";
+import {
```
EndymionJkb, line 17: Could optionally do a straight import (as the last element of the interfaces section), once it goes multi-line and isn't an example. (Arguably, StablePool might be used as an example.)

pkg/pool-weighted/contracts/WeightedPool.sol
```diff
@@ -13,7 +13,7 @@ import {
 } from "@balancer-labs/v3-interfaces/contracts/vault/IUnbalancedLiquidityInvariantRatioBounds.sol";
 import { IVault } from "@balancer-labs/v3-interfaces/contracts/vault/IVault.sol";
 import { IVaultErrors } from "@balancer-labs/v3-interfaces/contracts/vault/IVaultErrors.sol";
-import { SwapKind, PoolSwapParams, PoolConfig } from "@balancer-labs/v3-interfaces/contracts/vault/VaultTypes.sol";
+import "@balancer-labs/v3-interfaces/contracts/vault/VaultTypes.sol";
```
EndymionJkb, line 16: Could leave the explicit import, if it fits. Otherwise, straight imports should go at the bottom of the interfaces section.

pkg/pool-weighted/test/foundry/LiquidityApproximationWeighted.t.sol
```diff
@@ -104,7 +110,7 @@ contract LiquidityApproximationWeightedTest is LiquidityApproximationTest {
     ) public {
         // Weights can introduce some differences in the swap fees calculated by the pool during unbalanced add/remove
         // liquidity, so the error tolerance needs to be a bit higher than the default tolerance.
-        roundingDelta = 1e13;
+        defectRoundingDelta = 0.00001e16; // 0.00001%
```
EndymionJkb, line 113: These specific values were all determined empirically?

pkg/pool-weighted/test/foundry/WeightedPoolLimits.t.sol
```diff
@@ -7,7 +7,7 @@ import "forge-std/Test.sol";
 import { IERC20 } from "@openzeppelin/contracts/token/ERC20/IERC20.sol";
 
 import { IVaultAdmin } from "@balancer-labs/v3-interfaces/contracts/vault/IVaultAdmin.sol";
-import { PoolRoleAccounts, LiquidityManagement } from "@balancer-labs/v3-interfaces/contracts/vault/VaultTypes.sol";
+import "@balancer-labs/v3-interfaces/contracts/vault/VaultTypes.sol";
```
EndymionJkb, line 10: Either keep, since it fits, or put at the end.

pkg/solidity-utils/contracts/math/BasePoolMath.sol
```diff
@@ -186,13 +188,20 @@ library BasePoolMath {
...
         // total supply with the ratio of the change in invariant.
         // Since we multiply and divide we don't need to use FP math.
-        // Round down since we're calculating BPT amount out. `invariantWithFeesApplied` calculated with `newBalances`
-        // rounded down, which also contributes to a lower `bptAmountOut`.
+        // Round down since we're calculating BPT amount out. This is the most important result of this function,
+        // equivalent to:
+        // `totalSupply * (invariantWithFeesApplied / currentInvariant - 1)`
+        // Then, to round `bptAmountOut` down we use `invariantWithFeesApplied` rounded down and `currentInvariant`
+        // rounded up.
+        // If rounding makes `invariantWithFeesApplied` smaller or equal than `currentInvariant`, this would effectively
```
EndymionJkb, line 202:
```suggestion
        // If rounding makes `invariantWithFeesApplied` smaller or equal to `currentInvariant`, this would effectively
```

pkg/solidity-utils/contracts/math/BasePoolMath.sol
```diff
@@ -186,13 +188,20 @@ library BasePoolMath {
...
 
         // Calculate the amount of BPT to mint. This is done by multiplying the
         // total supply with the ratio of the change in invariant.
         // Since we multiply and divide we don't need to use FP math.
-        // Round down since we're calculating BPT amount out. `invariantWithFeesApplied` calculated with `newBalances`
-        // rounded down, which also contributes to a lower `bptAmountOut`.
+        // Round down since we're calculating BPT amount out. This is the most important result of this function,
+        // equivalent to:
+        // `totalSupply * (invariantWithFeesApplied / currentInvariant - 1)`
+        // Then, to round `bptAmountOut` down we use `invariantWithFeesApplied` rounded down and `currentInvariant`
```
EndymionJkb, line 200: Consider some spacing "//" lines around the "Then..." paragraph for better readability.

pkg/solidity-utils/contracts/math/BasePoolMath.sol
```diff
@@ -291,9 +300,13 @@ library BasePoolMath {
...
         // Calculate the new proportional balance by multiplying the new invariant ratio by the current balance.
         // Calculate the taxable amount by subtracting the new balance from the equivalent proportional balance.
-        uint256 currentInvariant = pool.computeInvariant(currentBalances);
+        // We round `currentInvariant` up as it affects the calculated `bptAmountIn` directly (see below).
+        uint256 currentInvariant = pool.computeInvariant(currentBalances, Rounding.ROUND_UP);
         // We round invariant ratio up (see reason below).
-        uint256 invariantRatio = pool.computeInvariant(newBalances).divUp(currentInvariant);
+        // This invariant ratio could be rounded up even more by rounding `currentInvariant` down. But since it only
+        // affects the taxable amount and the fee calculation, whereas `currentInvariant` affects BPT in more directly,
+        // we use `currentInvariant` rounded up here as well.
```
EndymionJkb, line 308: Would this (rounding) reduce the difference in fees between "paths" then? So we're making it "fairer" by reducing the delta in this direction, while rounding the invariant itself protects against unfavorable "round trips"

pkg/solidity-utils/contracts/math/BasePoolMath.sol
```diff
@@ -309,18 +322,20 @@ library BasePoolMath {
...
         swapFeeAmounts[tokenOutIndex] = fee;
 
-        // Calculate the amount of BPT to burn. This is done by multiplying the
-        // total supply with the ratio of the change in invariant.
-        // Since we multiply and divide we don't need to use FP math.
-        // Calculating BPT amount in, so we round up.
-        // Finally, lower `invariantWithFeesApplied` makes the subtraction larger, which also helps `bptAmountIn` to be
-        // larger since it's in the numerator.
+        // Calculate the amount of BPT to burn. This is done by multiplying the total supply with the ratio of the
+        // change in invariant.
```
EndymionJkb, line 332:
```suggestion
        // Calculate the amount of BPT to burn. This is done by multiplying the total supply by the ratio of the
        // invariant delta to the current invariant.
```
Maybe clearer?

pkg/solidity-utils/contracts/math/BasePoolMath.sol
```diff
@@ -309,18 +322,20 @@ library BasePoolMath {
...
+        // Calculating BPT amount in, so we round up. This is the most important result of this function, equivalent to:
+        // `totalSupply * (1 - invariantWithFeesApplied / currentInvariant)`.
+        // Then, to round `bptAmountIn` up we use `invariantWithFeesApplied` rounded down and `currentInvariant`
+        // rounded up.
+        // Since `currentInvariant` is rounded up and `invariantWithFeesApplied` is rounded down, the difference
+        // should always be positive. We revert if that is not the case with checked math.
```
EndymionJkb, line 338:
```suggestion
        // should always be positive. The checked math will revert if that is not the case.
```

pkg/solidity-utils/contracts/math/WeightedMath.sol
```diff
@@ -66,7 +66,7 @@ library WeightedMath {
     // The invariant is used to collect protocol swap fees by comparing its value between two times.
     // So we can round always to the same direction. It is also used to initiate the BPT amount and,
     // because there is a minimum BPT, we round down the invariant.
```
EndymionJkb, line 68: Stale comment now. (And not well-written besides.) The first sentence is left-over from v2.
```suggestion
    // The invariant functions are called by the Vault during various liquidity operations, and require a specific
    // rounding direction in order to ensure safety (i.e., that the final result is always rounded in favor of the protocol).
```

EndymionJkb (commented):
This is changing out from under me... adding current (minor) comments. Reviewed the main code that inspired the change, and it looks good!

pkg/solidity-utils/contracts/math/BasePoolMath.sol
jubeira: More or less.

The most important thing is that the overall amount we ask from the sender in this case makes the roundtrip lossy. Fee calculation is always behind a fee percentage which is < 1, so if you round in detriment of the fees but you still ask more tokens from the user it's not necessarily that you're making the indirect math 'fairer'.

The problem here is that we're computing 2 invariants already to get the invariant ratio. My understanding is that the way in this PR is the one that favors the protocol the most computing only these 2 invariants, which is the goal.

We could round in the protocol's favor even more by favoring the rounding for fees as well, but that would require computing one extra invariant which is too much to ask.

pkg/pool-weighted/test/foundry/LiquidityApproximationWeighted.t.sol
jubeira: Yes. I was a bit on the fence on this one, but it gives some insights to have them differentiated.
The test verifies that the defect is absorbed by the fees anyways, which is why I bumped the minimum swap fee for the pools.

This is just a first rough approximation though. In the next iteration of the tests we'll be measuring invariants before and after the operations to ensure that they are not going down in any case, which is a stronger assurance (even for 0 fees).

PR #932, opened 2024-08-27
Fix add liquidity to buffer

### LINKED ISSUE #923: Fix buffer shares calculation
Init is correct; add should consider total shares supply.

EndymionJkb:
# Description

Consider the current balances and rates when calculating the shares to be issued when adding liquidity to a buffer. It now more closely parallels the pools, calculating a "buffer share rate" that decreases as the convertToAssets rate increases (i.e., as the wrapped tokens are worth more underlying), such that fewer shares will be issued.

The test replicates the Certora PoC: add equal amounts underlying/wrapped once, then - with a rate of 2 - add twice the amount of underlying only, then immediately withdraw. The existing balances and rate will discount the shares received the second time (i.e., will be less than instead of equal to the amount of underlying tokens added). If they immediately withdraw, the total value of the tokens withdrawn will be <= the total value just added.

I also simplified the initialize logic a bit, inlining a function that was only called once, not doing unnecessary additions, etc., for clarity and to reduce the impact of the new logic on the bytecode size.

## Type of change

- [X] Bug fix 
- [ ] New feature 
- [ ] Breaking change 
- [ ] Dependency changes
- [ ] Code refactor / cleanup
- [ ] Other

## Checklist:

- [X] The diff is legible and has no extraneous changes (well, there's a lint fix from current master)

## Issue Resolution

Resolves #923.


pkg/vault/contracts/VaultAdmin.sol
```diff
@@ -429,22 +430,23 @@ contract VaultAdmin is IVaultAdmin, VaultCommon, Authentication, VaultGuard {
...
+        // depositing the tokens to initialize the buffer, which receives the balance.
+        _bufferTotalShares[wrappedToken] = _MINIMUM_TOTAL_SUPPLY;
+        _bufferLpShares[wrappedToken][address(0)] = _MINIMUM_TOTAL_SUPPLY;
+
+        emit BufferSharesMinted(wrappedToken, address(0), _MINIMUM_TOTAL_SUPPLY);
+
         issuedShares -= _MINIMUM_TOTAL_SUPPLY;
+
         _mintBufferShares(wrappedToken, sharesOwner, issuedShares);
-        _mintMinimumBufferSupplyReserve(wrappedToken);
```
jubeira, line 447: I'd just reverse the order of the calls here and keep it that way.

I know it goes a bit against the rule of not putting code into a function that we only use once, but it's the same pattern we use for pool initialization, and here we're essentially doing the same thing. We're in the admin here so bytecode size is not an issue.

pkg/vault/contracts/VaultAdmin.sol
```diff
@@ -472,16 +474,30 @@ contract VaultAdmin is IVaultAdmin, VaultCommon, Authentication, VaultGuard {
...
+        // The buffer share rate (equivalent to pool's "BPT rate") is the ratio of the total shares
+        // (equivalent to pool's "total supply") to the buffer invariant. As the rate goes up (i.e., wrapped tokens
+        // become more valuable), the share rate will go down, and the user will receive fewer shares.
+        // The buffer must have been initialized to get here, so the balances cannot be zero, and the rate calculation
+        // will never revert.
+        uint256 currentBufferShareRate = _bufferTotalShares[wrappedToken].divDown(bufferInvariant);
+
+        uint256 bufferInvariantDelta = wrappedToken.convertToAssets(amountWrappedRaw) + amountUnderlyingRaw;
+        // The amount of shares to issue is the added liquidity in underlying terms multiplied by the share rate.
+        issuedShares = bufferInvariantDelta.mulDown(currentBufferShareRate);
```
jubeira, line 491: So, this is basically `issuedShares = bufferTotalShares * (invariantRatio - 1)`, where `invariantRatio = (bufferInvariant + bufferInvariantDelta) / bufferInvariant`.

Replace and simplify, and you get 
```
issuedShares = bufferTotalShares * bufferInvariantDelta / currentInvariant
```

I think you can do that with raw multiplication and division without FP math, since it's a product and a division rounded down. The current code uses two FP operations instead, and divides first, multiplies second.

pkg/vault/contracts/VaultAdmin.sol
```diff
@@ -492,23 +508,12 @@ contract VaultAdmin is IVaultAdmin, VaultCommon, Authentication, VaultGuard {
             revert BufferSharesInvalidReceiver();
         }
 
-        uint256 newTotalSupply = _bufferTotalShares[wrappedToken] + amount;
-
-        _ensureMinimumTotalSupply(newTotalSupply);
```
jubeira, line 497: yeah I know this one is redundant, but it's just about being overly defensive and following the same pattern as with `mint` for BPT. I'd keep it symmetrical.

pkg/vault/test/foundry/BufferVaultPrimitive.t.sol
```diff
@@ -455,6 +456,50 @@ contract BufferVaultPrimitiveTest is BaseVaultTest {
...
+        (uint256 bufferUnderlyingBalance, uint256 bufferWrappedBalance) = vault.getBufferBalance(waDAI);
+
+        uint256 expectedUnderlyingOut = proportionalWithdrawPct.mulDown(bufferUnderlyingBalance);
+        uint256 expectedWrappedOut = proportionalWithdrawPct.mulDown(bufferWrappedBalance);
+        // Will get 1333.333/3333.333 = 40% of value:
+        // [0.4 * 3000, 0.4 * 1000] = [1200 underlying, 400 wrapped] - worth 2000 = amount in
+        vm.prank(lp);
+        (uint256 removedUnderlying, uint256 removedWrapped) = vault.removeLiquidityFromBuffer(waDAI, secondAddLpShares);
+        assertApproxEqAbs(removedUnderlying, expectedUnderlyingOut, 1e6, "Wrong underlying amount removed");
+        assertApproxEqAbs(removedWrapped, expectedWrappedOut, 1e6, "Wrong wrapped amount removed");
```
jubeira, line 496: Let's see if this improves even further using raw math instead of FP (see comment above).

jubeira (commented):
Thanks for pushing this @EndymionJkb !

Looking good now. Just a few comments with the new math, and the old checks. I know the shares validation around the minimum supply are redundant, but they follow the same pattern as with `mint` for BPTs. I don't see a reason to break the symmetry, since we're doing essentially the same. The problem in `init` was the order of the calls.

Other than that, now the math adds up which is the main point.

pkg/vault/test/foundry/BufferVaultPrimitive.t.sol
EndymionJkb: It didn't, except for the add/remove test at the end.

pkg/vault/contracts/VaultAdmin.sol
EndymionJkb: Physically painful... but ok; I'll add a comment.

jubeira (commented):
Thanks @EndymionJkb, this looks good!

Before wrapping up, if I'm not mistaken the rounding considerations we're applying to the Vault in general also apply here.

`addLiquidityToBuffer` is transient, so we can add whatever and remove what we get without paying a dime. In other words, `addToBuffer` + `removeFromBuffer` should always be lossy.

We're rounding down issued shares, but there are some calls to `convert` in the middle. Can we add a fuzz test that adds liquidity with random amounts and random wrapped rates and then removes the shares, checking that the whole operation is lossy always?

/cc @joaobrunoah

pkg/solidity-utils/contracts/test/ERC4626TestToken.sol
```diff
@@ -39,6 +43,10 @@ contract ERC4626TestToken is ERC4626, IRateProvider {
         return _convertToAssets(FixedPoint.ONE, Math.Rounding.Floor);
     }
 
+    function mockRate(uint256 newRate) external {
+        _mockRate = newRate;
+    }
```
joaobrunoah, line 48: Mocking rate like this is probably good enough for this test, but it does not cover rounding issues with the rate calculation itself. Not sure if it should be in this PR, but maybe we could create a mockRate that receives `assets` and `totalSupply`, minting underlying tokens and minting the wrapped itself. (minting underlying increases the rate, minting wrapped decreases the rate).

pkg/vault/test/foundry/BufferVaultPrimitive.t.sol
```diff
@@ -508,6 +509,134 @@ contract BufferVaultPrimitiveTest is BaseVaultTest {
...
+    }
+
+    function _addLiquidityToBufferWithRate(
+        uint256 firstDepositUnderlying,
+        uint256 secondDepositUnderlying,
+        uint256 wrappedAmount,
+        uint256 rate
+    ) internal {
+        // Ensure we're adding more than the minimum, or it will revert.
+        vm.assume(firstDepositUnderlying + wrappedAmount >= MINIMUM_TOTAL_SUPPLY);
```
joaobrunoah, line 589: Nice!

pkg/vault/test/foundry/BufferVaultPrimitive.t.sol
```diff
@@ -508,6 +509,134 @@ contract BufferVaultPrimitiveTest is BaseVaultTest {
...
+        vm.prank(lp);
+        uint256 firstAddLpShares = router.initializeBuffer(waDAI, firstDepositUnderlying, wrappedAmount);
+        assertEq(
+            firstAddLpShares,
+            firstDepositUnderlying + wrappedAmount - MINIMUM_TOTAL_SUPPLY,
+            "Wrong first lpShares added"
+        );
+
+        // Change the rate after initialization.
+        waDAI.mockRate(rate);
```
joaobrunoah, line 600: As I said, if we mockRate by minting underlying/wrapped, mockRate would receive a certain amount of underlying and wrapped to mint. The rate would be `waDAI.getRate()`, and then this test would also check rounding issues between getRate() and the convert operations.

pkg/vault/test/foundry/BufferVaultPrimitive.t.sol
```diff
@@ -508,6 +509,134 @@ contract BufferVaultPrimitiveTest is BaseVaultTest {
...
+
+        // Predict the amount of shares to receive
+        (uint256 bufferUnderlyingBalance, uint256 bufferWrappedBalance) = vault.getBufferBalance(waDAI);
+        uint256 currentInvariant = bufferUnderlyingBalance + bufferWrappedBalance.mulDown(rate);
```
joaobrunoah, line 604:
```suggestion
        uint256 currentInvariant = bufferUnderlyingBalance + waDAI.convertToAssets(bufferWrappedBalance);
```

Multiplying by rate is different than convertToAssets. Multiply by the rate is the same as doing `bufferWrappedBalance.mulDown(underlyingBalanceOfWrapper).divDown(wrapperTotalSupply)`, which introduces rounding issues since mulDown divides by 1e18. 

ConvertToAssets is `bufferWrappedBalance * underlyingBalanceOfWrapper / wrapperTotalSupply`, which is a more precise operation.

pkg/vault/test/foundry/BufferVaultPrimitive.t.sol
```diff
@@ -508,6 +509,134 @@ contract BufferVaultPrimitiveTest is BaseVaultTest {
...
+
+        uint256 expectedUnderlyingOut = proportionalWithdrawPct.mulDown(bufferUnderlyingBalance);
+        uint256 expectedWrappedOut = proportionalWithdrawPct.mulDown(bufferWrappedBalance);
+
+        vm.prank(lp);
+        (uint256 removedUnderlying, uint256 removedWrapped) = vault.removeLiquidityFromBuffer(waDAI, secondAddLpShares);
+        assertApproxEqAbs(removedUnderlying, expectedUnderlyingOut, 1e8, "Wrong underlying amount removed");
+        assertApproxEqAbs(removedWrapped, expectedWrappedOut, 1e8, "Wrong wrapped amount removed");
+
+        uint256 totalUnderlyingValue = removedUnderlying + rate.mulDown(removedWrapped);
```
joaobrunoah, line 634: I'd call this bufferInvariantAfter, or something with the buffer invariant, to be consistent with the naming we're using. Besides, I'd prefer convertToAssets instead of using the rate directly here.

joaobrunoah (approved):
A few comments, but looks nice! Not sure if my comments should be addressed in this PR, in another PR or even addressed at all, since we're measuring the imprecisions with the rate in other places too.

jubeira (approved):
LGTM, thanks @EndymionJkb !

Just a few suggestions for the tests to check the invariant of the buffer more directly. Other than it's good to :ship:

EndymionJkb:
Ok; will address and merge tonight. That "invariantAfter" was suggested in a previous comment :)

PR #933, opened 2024-08-28
Round up rate when scaling a token out amount

### LINKED ISSUE #884: Consider adding 1 wei to token rates when scaling down from scaled18 to raw amounts
Consider the following swap using a Linear Pool with two tokens, and each token has a rate provider
```
tokenARate = 1200000000000000000
tokenBRate = 3299999999999999999

---SWAP---
amountInRaw  4166666666666666666666666
amountInSca  4999999999999999999999999
amountOutSca 4999999999999999999999999
amountOutRaw 1515151515151515151974288 => This number is a bit inflated
```

When tokenBRate was calculated, it was rounded down, and since the scaled18 number is divided by the rate to get the raw amount, the amountOutRaw was inflated a bit.

This would be the scenario with the correct rate (rounding up the tokenBRate when scaling down)

```
tokenARate = 1200000000000000000
tokenBRate = 3300000000000000000 (only when scaling down)

---SWAP---
amountInRaw  4166666666666666666666666
amountInSca  4999999999999999999999999
amountOutSca 4999999999999999999999999
amountOutRaw 1515151515151515151515151
```

With this scenario in mind, we should consider rounding up (or adding 1 wei) to token rates when scaling down to raw amounts.

joaobrunoah:
# Description

We identified that a swap in a pool with rate providers was giving an inconsistent result due to rounding errors in the rate calculation. Basically, in a swap, the following formula needs to be true (considering a linear pool for simplicity, but the logic applies to any pool):

`amountInRaw * rateTokenIn = amountOutRaw * rateTokenOut`

It implies that, for ExactIn:
`amountOutRaw = amountInRaw * rateTokenIn / rateTokenOut`

We were using rates rounded down, but if the `rateTokenOut` is rounded down, in some cases the pool may be drained. The same happens with ExactOut swap:

`amountInRaw = amountOutRaw * rateTokenOut / rateTokenIn`

If `rateTokenOut` is rounded down, the user may pay less tokens than it should, which also drains the pool.

The solution is to add 1 wei in the rate of tokens out when swapping.

Also, fixing this rate didn't fix all issues with ERC4626 rates. The following changes were also implemented in this PR:

* Improved BatchRouterERC4626 tests to don't use lastBalancesLiveScaled18 when checking pool balances, since this number is calculated using a rate rounded down which does not reflect the actual operation;
* Added a convert factor when using the buffer liquidity to wrap/unwrap tokens without calling the wrapper protocol, since the convert operation is less conservative than the actual operation.

## Type of change

- [x] Bug fix 
- [ ] New feature 
- [ ] Breaking change 
- [ ] Dependency changes
- [x] Code refactor / cleanup
- [ ] Other

## Checklist:

## Issue Resolution

Closes #884


pkg/vault/test/foundry/BatchRouterERC4626Pool.t.sol
```diff
@@ -155,16 +158,16 @@ contract BatchRouterERC4626PoolTest is BaseERC4626BufferTest {
             );
         }
         {
-            (, , , uint256[] memory lastBalancesLiveScaled18) = vault.getPoolTokenInfo(address(erc4626Pool));
+            (, , uint256[] memory balances, ) = vault.getPoolTokenInfo(address(erc4626Pool));
```
joaobrunoah, line 161: In theory, lastBalancesLive should be equal to the underlying amount, but it's not. When we call "convertToShares", we multiply the amount by the totalSupply and divide by the assets allocated in the wrapper directly (in most cases), which does not introduce rounding issues.

However, lastBalancesLive is the `wrappedAmount` multiplied by convertToAssets(FixedPoint.ONE), which is the rate of one share in a 18 decimal number. This introduces a rounding imprecision proportional to the `wrappedAmount`.

pkg/vault/test/foundry/utils/BaseERC4626BufferTest.sol
```diff
@@ -223,6 +223,6 @@ abstract contract BaseERC4626BufferTest is BaseVaultTest {
         vm.stopPrank();
         // Changing asset balances without minting shares changes the rate so that it is no longer 1.
         dai.mint(address(waDAI), 2 * erc4626PoolInitialAmount);
-        usdc.mint(address(waUSDC), 4 * erc4626PoolInitialAmount);
+        usdc.mint(address(waUSDC), 23 * erc4626PoolInitialAmount);
```
joaobrunoah, line 226: Changing 4 by 23 creates a wrapped rate that breaks the old tests.

pkg/solidity-utils/contracts/helpers/ScalingHelpers.sol
```diff
@@ -317,4 +317,14 @@ library ScalingHelpers {
         uint256 decimalsDifference = 18 - tokenDecimals;
         return FixedPoint.ONE * 10 ** decimalsDifference;
     }
+
+    /**
+     * @notice Rounds up a rate informed by a rate provider.
+     * @dev The rates are calculated by an external rate provider have rounding errors. Intuitively, a rate provider
```
joaobrunoah, line 323:
```suggestion
     * @dev Rates calculated by an external rate provider have rounding errors. Intuitively, a rate provider
```

jubeira (commented):
Man, what a lift!

I agree with the changes at the vault, and the tests are pretty good. They are pretty complex though, which is why I'd propose adding a handful of extra comments here and there to help the reader interpret the intention behind the tests.

Besides that I think it's ready to go in. I haven't finished reviewing `YieldBearingPoolSwapBase`, but everything looks correct. I'll give it one last look before the final approval.

pkg/solidity-utils/contracts/helpers/ScalingHelpers.sol
```diff
@@ -317,4 +317,14 @@ library ScalingHelpers {
...
     }
+
+    /**
+     * @notice Rounds up a rate informed by a rate provider.
+     * @dev Rates calculated by an external rate provider have rounding errors. Intuitively, a rate provider
+     * rounds the rate down so the pool math is executed with conservative amounts. However, when upscaling or
+     * downscaling the amount out, the rate should be rounded up to make sure the amounts scaled are conservative.
+     */
+    function computeRateRoundUp(uint256 rate) internal pure returns (uint256) {
+        return (rate / FixedPoint.ONE) * FixedPoint.ONE == rate ? rate : rate + 1;
```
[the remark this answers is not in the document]
joaobrunoah: It saves 15 bytes, so I unchecked the division and multiplication, makes sense.

pkg/vault/test/foundry/BatchRouterERC4626Pool.t.sol
```diff
@@ -933,4 +863,39 @@ contract BatchRouterERC4626PoolTest is BaseERC4626BufferTest {
...
+        testBalances.waDAIBuffer.wrapped = waDAIBufferBalanceWrapped;
+
+        (uint256 waUSDCBufferBalanceUnderlying, uint256 waUSDCBufferBalanceWrapped) = vault.getBufferBalance(waUSDC);
+        testBalances.waUSDCBuffer.underlying = waUSDCBufferBalanceUnderlying;
+        testBalances.waUSDCBuffer.wrapped = waUSDCBufferBalanceWrapped;
+
+        testBalances.daiIdx = 0;
+        testBalances.usdcIdx = 1;
+        testBalances.waDaiIdx = 2;
+        testBalances.waUsdcIdx = 3;
```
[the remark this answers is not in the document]
joaobrunoah: The index of each token is defined by the order of tokenArray. Since the tokenArray is defined in the same function, we can for sure know each index and hardcode it. I could use sort methods too, but looked much more complex than it should be. I added a comment in the code explaining this.

pkg/vault/test/foundry/fork/YieldBearingPoolSwapBase.t.sol
```diff
@@ -466,27 +319,42 @@ abstract contract YieldBearingPoolSwapBase is BaseVaultTest {
...
+            uint256 wrappedAmountInScaled18 = vars.expectedWrappedDeltaTokenIn.divDown(_ybToken1Factor);
             // PoolMock is linear, so wrappedAmountInScaled18 = wrappedAmountOutScaled18
-            uint256 wrappedAmountOutRaw = wrappedAmountInScaled18.mulDown(_ybToken2Factor);
-            vars.expectedDeltaTokenOut = _previewRedeemWithBuffer(ybToken2, wrappedAmountOutRaw, shouldUseConvert);
+            vars.expectedWrappedDeltaTokenOut = wrappedAmountInScaled18.mulDown(_ybToken2Factor);
+            (
+                vars.expectedUnderlyingDeltaTokenOut,
+                vars.expectedUnderlyingSurplusTokenOut,
+                vars.expectedWrappedSurplusTokenOut
+            ) = _previewRedeemWithBuffer(ybToken2, vars.expectedWrappedDeltaTokenOut, withBufferLiquidity);
```
jubeira, line 341: So if I understand this correctly, now we also have expected surplus for wrapped and underlying (one for buffer in, one for buffer out). Is that correct?

pkg/vault/test/foundry/fork/YieldBearingPoolSwapBase.t.sol
```diff
@@ -512,10 +380,20 @@ abstract contract YieldBearingPoolSwapBase is BaseVaultTest {
...
+        // Query and actual operation can return different results, depending on the difference of decimals. The error
+        // is amplified by the rate of the token out and by `vaultConvertFactor` when using the buffer liquidity in
+        // the actual operation.
+        uint256 sharesError = vars.ybTokenOut.convertToAssets(
+            (withBufferLiquidity ? vaultConvertFactor : 1) * decimalError
+        );
+        uint256 absTolerance = sharesError + (withBufferLiquidity ? vaultConvertFactor : 1);
+        // If convertToAssets return 0, absTolerance may be smaller than the error introduced by the difference of
+        // decimals, so keep the decimalError.
+        absTolerance = absTolerance > decimalError ? absTolerance : decimalError;
```
jubeira, line 396: All of this is just to get an upper bound for the absolute error between the query and the actual operation right?
I'd clarify that in the comments: "The error _between the query and the actual operation_ is proportional..."

jubeira: I'm happy knowing that the relative error is bounded by 0.01%.
Other than that, with the conversion factor is the error always one-sided? i.e. is the query always worse than the actual operation for the user then? That would be a good thing to verify: besides using `assertApproxEqAbs` or `assertApproxEqRel` we can use `gt` / `lt` (not saying we should remove the existing assertions, maybe add one more if it's easy for documentation purposes as well).

pkg/vault/test/foundry/fork/YieldBearingPoolSwapBase.t.sol
```diff
@@ -744,45 +649,42 @@ abstract contract YieldBearingPoolSwapBase is BaseVaultTest {
...
-                int256(vars.bufferBeforeSwapYbTokenIn) +
-                    (
-                        vars.expectedBufferDeltaTokenIn < int256(0)
-                            ? int256(vars.ybTokenIn.convertToShares(uint256(-vars.expectedBufferDeltaTokenIn)))
-                            : -int256(vars.ybTokenIn.convertToShares(uint256(vars.expectedBufferDeltaTokenIn)))
-                    )
-            ),
+            vars.bufferBeforeSwapYbTokenIn +
+                vars.expectedWrappedSurplusTokenIn -
+                (vars.withBufferLiquidity ? vars.expectedWrappedDeltaTokenIn : 0),
```
jubeira, line 666: :+1:  this one looks better.

pkg/vault/test/foundry/fork/YieldBearingPoolSwapBase.t.sol
```diff
@@ -1024,96 +929,94 @@ abstract contract YieldBearingPoolSwapBase is BaseVaultTest {
...
             // rebalance the buffer. It can introduce rounding issues, so we should consider the rebalance in our
             // result preview.
-            uint256 underlyingSurplus = bufferBalances.getBufferUnderlyingSurplus(wToken);
-            uint256 wrappedSurplusToLeaveInTheBuffer = wToken.convertToShares(underlyingSurplus);
-            uint256 vaultUnderlyingDelta = wToken.previewMint(wrappedToMint + wrappedSurplusToLeaveInTheBuffer);
-            return vaultUnderlyingDelta - underlyingSurplus;
+            underlyingSurplus = bufferBalances.getBufferUnderlyingSurplus(wToken);
+            wrappedSurplus = wToken.convertToShares(underlyingSurplus);
+            uint256 vaultUnderlyingDelta = wToken.previewMint(wrappedToMint + wrappedSurplus);
+            depositedAssets = vaultUnderlyingDelta - underlyingSurplus;
```
jubeira, line 972: This bit is replicating the logic in the vault right?

jubeira (approved):
Just finished checking the tests. Only a few more small questions and observations, but I think this is looking good! Thanks for pushing it @joaobrunoah !

pkg/vault/test/foundry/fork/YieldBearingPoolSwapBase.t.sol
joaobrunoah: We were calculating the surplus before elsewhere, in the test function, and this calculation was different depending n the case. Now the expected surplus is standard.

pkg/vault/test/foundry/fork/YieldBearingPoolSwapBase.t.sol
joaobrunoah: Done!

pkg/vault/test/foundry/fork/YieldBearingPoolSwapBase.t.sol
joaobrunoah: Yes, I added some comments to explain this better.

pkg/vault/test/foundry/YieldBearingPoolBufferAsVaultPrimitive.t.sol
```diff
@@ -118,20 +137,37 @@ contract YieldBearingPoolBufferAsVaultPrimitiveTest is BaseERC4626BufferTest {
         (uint256[] memory pathAmountsIn, address[] memory tokensIn, uint256[] memory amountsIn) = batchRouter
             .swapExactOut(paths, MAX_UINT256, false, bytes(""));
 
+        // For ExactOut, the last step is computed first (Unwrap -> Swap -> Wrap).
+        // Compute Unwrap. The exact amount out in USDC is `swapAmount` and the token out is USDC, so the unwrap
+        // occurs in the waUSDC buffer. Since the buffer has liquidity, we need to consider the vaultConvertError.
+        // `waUsdcAmountOutRaw` is the ExactOut amount of the pool swap, and input to the unwrap, that's why the
+        // vaultConvertFactor is added.
```
EndymionJkb, line 144:
```suggestion
        // `waUsdcAmountOutRaw` is the ExactOut amount of the pool swap, and the input to the unwrap.
        // That's why the `vaultConvertFactor` is added.
```

EndymionJkb (commented):
A few minor comments (and one math refactor suggestion)

pkg/vault/test/foundry/fork/YieldBearingPoolSwapBase.t.sol
joaobrunoah: I couldn't assert that queries are always greater or smaller than the actual operation. Not related to the vaultConvertFactor, maybe related to a rate calculation that introduces some uncertainty. I can check later why this is not the case.

pkg/solidity-utils/contracts/helpers/ScalingHelpers.sol
```diff
@@ -317,4 +317,20 @@ library ScalingHelpers {
...
+     * downscaling the amount out, the rate should be rounded up to make sure the amounts scaled are conservative.
+     */
+    function computeRateRoundUp(uint256 rate) internal pure returns (uint256) {
+        uint256 roundedRate;
+        // If rate is divisible by FixedPoint.ONE, roundedRate and rate will be equal. It means that rate has 18 zeros,
+        // so there's no rounding issue and the rate should not be rounded up.
+        unchecked {
+            roundedRate = (rate / FixedPoint.ONE) * FixedPoint.ONE;
+        }
+        return roundedRate == rate ? rate : rate + 1;
```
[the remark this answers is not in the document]
joaobrunoah: We didn't want to uncheck the `rate + 1` because the rate value is unlimited.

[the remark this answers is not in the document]
joaobrunoah: Replacing by remainder, even unchecked, raises the code by 16 bytes, so I'll leave the division and multiplication, ok?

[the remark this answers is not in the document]
EndymionJkb: Sure; you'd think the mod way would be smaller... but maybe the division/multiplication way is optimized better.

PR #934, opened 2024-08-29
Add StableSurge hook example with tests

johngrantuk:
# Description

Add StableSurge hook example based of ZenDragons [initial doc](https://www.notion.so/Peg-Protection-Swap-Fee-Surge-Weights-Based-8fa8f525a3a14698afcdcd0d26ad3778?pvs=4).

## Type of change

- [ ] Bug fix 
- [ ] New feature 
- [ ] Breaking change 
- [ ] Dependency changes
- [ ] Code refactor / cleanup
- [X] Other - Hook example

## Checklist:


pkg/pool-hooks/contracts/StableSurgeHookExample.sol
```diff
@@ -0,0 +1,235 @@
...
+    /// @inheritdoc IHooks
+    function onRegister(
+        address factory,
+        address pool,
+        TokenConfig[] memory,
+        LiquidityManagement calldata
+    ) public override onlyVault returns (bool) {
+        // This hook implements a restrictive approach, where we check if the factory is an allowed factory and if
+        // the pool was created by the allowed factory.
+        emit StableSurgeHookExampleRegistered(address(this), factory, pool);
```
MattPereira, line 94: Is this best practice I should add to scaffold examples? 

Looks like the `PoolRegistered` event already contains all the same info, so maybe this just enables easier indexing / filtering?

https://github.com/balancer/balancer-v3-monorepo/blob/184494da6351fc9daf61b738ea2ed7da63694ee8/pkg/vault/contracts/VaultExtension.sol#L316-L326

EndymionJkb: I introduced this; there are already PRs in docs about this, and the monorepo has these added. It's up to the subgraph / front end people if they find this useful, but I thought this was a simpler, focused event in case we have very popular hooks that are used a lot. Since there could be multiple instances of the same hook with different parameters (in some cases).

pkg/pool-hooks/contracts/StableSurgeHookExample.sol
```diff
@@ -0,0 +1,235 @@
...
+/**
+ * @notice Hook that applies a fee for out of range or undesirable amounts of tokens in relation to a threshold.
+ * @dev Uses the dynamic fee mechanism to apply a directional fee.
+ */
+contract StableSurgeHookExample is BaseHooks, VaultGuard, Ownable {
+    using FixedPoint for uint256;
+    // Only pools from a specific factory are able to register and use this hook.
+    address private immutable _allowedFactory;
+    // Defines the range in which surging will not occur; meaning surging will occur when:
+    // B(i-after swap) / (sum(B(n-after swap))) > 1/n + gamma
```
EndymionJkb, line 33: I'm not sure what all the variables mean: i, n, B(), etc. Range of what? Maybe give some more details / a numerical example.

pkg/pool-hooks/contracts/StableSurgeHookExample.sol
```diff
@@ -0,0 +1,235 @@
...
+    uint256 private _threshold;
+    // An amplification coefficient to amplify the degree a fee increases upon the threshold being triggered
```
EndymionJkb, line 35:
```suggestion
    // An amplification coefficient to amplify the degree to which a fee increases after the threshold is met.
```
Maybe this is clearer? Non-inline comments always end in a period.

pkg/pool-hooks/contracts/StableSurgeHookExample.sol
```diff
@@ -0,0 +1,235 @@
...
+        IVault vault,
+        address allowedFactory,
+        uint256 threshold,
+        uint256 surgeCoefficient
+    ) VaultGuard(vault) Ownable(msg.sender) {
+        _allowedFactory = allowedFactory;
+        _threshold = threshold;
+        _surgeCoefficient = surgeCoefficient;
+        emit ThresholdChanged(address(this), threshold);
+        emit SurgeCoefficientChanged(address(this), surgeCoefficient);
```
EndymionJkb, line 77: What we typically do here is encapsulate the event with the set function. 
```suggestion
        _setThreshold(threshold);
        _setSurgeCoefficient(surgeCoefficient);
```
Then, for both the threshold and the surgeCoefficient:

```
function setThreshold(uint64 newThreshold) external onlyOwner {
    _setThreshold(newThreshold);
}

function _setThreshold(uint64 newThreshold) private {
    // This should be validated; e.g. <= FixedPoint.ONE or some max value?
    _threshold = newThreshold;

    emit ThresholdChanged(address(this), newThreshold);
}
```

pkg/pool-hooks/contracts/StableSurgeHookExample.sol
EndymionJkb: Would move the event down after the check, though. No sense emitting and then reverting, plus it separates the comment from the logic.

pkg/pool-hooks/contracts/StableSurgeHookExample.sol
```diff
@@ -0,0 +1,235 @@
...
+    /// @inheritdoc IHooks
+    function onComputeDynamicSwapFeePercentage(
+        PoolSwapParams calldata params,
+        address pool,
+        uint256 staticSwapFeePercentage
+    ) public view override onlyVault returns (bool, uint256) {
+        uint256 amp;
+        (amp, , ) = IStablePool(pool).getAmplificationParameter();
+
+        // In order to calculate weightAfterSwap we need balances after swap so we must compute swap amount
```
EndymionJkb, line 108:
```suggestion
        // In order to calculate `weightAfterSwap` we need balances after swap, so we must compute the swap amount.
```

pkg/pool-hooks/contracts/StableSurgeHookExample.sol
```diff
@@ -0,0 +1,235 @@
...
+    }
+
+    /// @inheritdoc IHooks
+    function onComputeDynamicSwapFeePercentage(
+        PoolSwapParams calldata params,
+        address pool,
+        uint256 staticSwapFeePercentage
+    ) public view override onlyVault returns (bool, uint256) {
+        uint256 amp;
+        (amp, , ) = IStablePool(pool).getAmplificationParameter();
```
EndymionJkb, line 106: Does this all still work if the amp factor's changing?

pkg/pool-hooks/contracts/StableSurgeHookExample.sol
```diff
@@ -0,0 +1,235 @@
...
+
+    /**
+     * Defines the range in which surging will not occur.
+     * @dev An expected value for threshold in a 2 token (n=2) would be 0.1.
+     * This would mean surging would occur if any token reaches 60% of the total of balances.
+     * @param numberOfAssets Number of assets in the pool.
+     * @param threshold Theshold value.
+     */
+    function getThresholdBoundary(uint256 numberOfAssets, uint256 threshold) public pure returns (uint256) {
+        return 1e18 / numberOfAssets + threshold;
```
EndymionJkb, line 172:
```suggestion
        return FixedPoint.ONE / numberOfAssets + threshold;
```
We've been using the defined constant for this vs. hard-coding numbers. So this would create an upper bound on the threshold then. With 2 tokens, the max threshold would be 0.5? With 8 tokens, 0.875? Assuming the boundary doesn't go over 1?

How can this be set on the hook then? If I understand it correctly, and the number of pool tokens constrains the maximum threshold, it would be different for every pool then - or at least every pool size. So, you'd set the threshold for all 2-token pools, then all 3-token pools, etc. (I think I don't understand this well.)

pkg/pool-hooks/contracts/StableSurgeHookExample.sol
EndymionJkb: Also, if this is a percentage, it should be "thresholdPercentage". We're naming all percentage things explicitly.

pkg/pool-hooks/contracts/StableSurgeHookExample.sol
EndymionJkb: How is surge coefficient validated? (i.e., what is the range)? You can't go over 100% on a dynamic fee, or the Vault will revert.

pkg/pool-hooks/test/foundry/StableSurgeExample.t.sol
```diff
@@ -0,0 +1,333 @@
...
+    uint256 SURGE_COEFFICIENT = 50e18;
+
+    // The following reference values were calculated manually using settings above
+    uint256 initialAmount = 50e18;
+    uint256[] balancesScaled18 = [initialAmount, initialAmount].toMemoryArray();
+    uint256 expectedThresholdBoundary = 0.6e18;
+    uint256 amountInBelowThreshold = 1e18;
+    uint256 amountInAboveThreshold = 30e18;
+    // Return from GivenIn swap, input=amountInAboveThreshold, minus fee
+    uint256 amountOutAboveThreshold = 29.850106042102014697e18;
```
EndymionJkb, line 47: These would be calculated in the code based on the constants above, but we can look at the tests once the main contract is done.

EndymionJkb (commented):
See questions in the comments; maybe we can sync and talk about the intent and how this should work.

pkg/pool-hooks/contracts/StableSurgeHookExample.sol
johngrantuk: Based of Daniel conclusion during the Office Hours chat it seems like we should leave this as is?

pkg/pool-hooks/contracts/StableSurgeHookExample.sol
johngrantuk: Tried to clarify this a bit more in a more detailed comment. Let me know if this works?

pkg/pool-hooks/contracts/StableSurgeHookExample.sol
johngrantuk: From my understanding `getAmplificationParameter` will return the instantaneous value of the amp which is then used in the calculations so I believe this works.

pkg/pool-hooks/contracts/StableSurgeHookExample.sol
johngrantuk: Will check with Zen on best way to validate.

pkg/pool-hooks/contracts/StableSurgeHookExample.sol
EndymionJkb: Sure; it's fine. Clearer this way, and we don't need to optimize for a case when it reverts.

pkg/pool-hooks/contracts/StableSurgeHookExample.sol
johngrantuk: From Zen:
* I don’t think the threshold should matter so much as long as it’s less than 1 - 1/n 
* surge coefficient: baseFee * 𝜇 / (1/n) < 100 would be the test then

EndymionJkb (commented):
See this Notion doc for my thoughts here:
https://www.notion.so/Stable-Surge-algorithm-113e395e2281801ba9f1d452e3a9be29

jubeira:
Superseded by #1085

PR #936, opened 2024-08-30
Separate buffer and pool minimum total supply limits

EndymionJkb:
# Description

Well this was one of those "pull on a sweater thread" ones, where the diff got a little out-of-hand for a relatively simple change.

We had a single `TotalSupplyTooLow` error, which I've separated into `PoolTotalSupplyTooLow` and `BufferTotalSupplyTooLow`. The limits themselves are likewise now called `POOL_MINIMUM_TOTAL_SUPPLY` and `MINIMUM_MINIMUM_TOTAL_SUPPLY`, and have different values.

When propagating this to the tests, I found we had lots of duplicate definitions of these values; I moved everything to `BaseVaultTest`, and named them the same as in the Vault code.

In an attempt to reduce bytecode (only ~10 bytes), I added getters for the minimum values, which we should probably have anyway (certainly for the newly non-constant minimum trade size), and removed the constant from the error.

Finally, when updating the interfaces, I noticed they'd diverged a bit (and from the VaultExplorer). Plus we were missing a recently added buffer function. So it discharges some tech debt relating to comments that drifted a bit out of sync, and updates the Vault Explorer with the new functions. (These are separable, if there's too much going on here.)

This explicitly sets the "production" trade amounts for tests that still expect them. We can now make versions that leave the minimums at zero (out of scope for this PR).

## Type of change

- [ ] Bug fix 
- [ ] New feature 
- [ ] Breaking change 
- [ ] Dependency changes
- [X] Code refactor / cleanup
- [ ] Other

## Checklist:


pkg/vault/contracts/Vault.sol
```diff
@@ -1567,6 +1574,16 @@ contract Vault is IVaultMain, VaultCommon, Proxy {
...
 
+    /// @inheritdoc IVaultMain
+    function getMinimumTradeAmount() external view returns (uint256) {
+        return _MINIMUM_TRADE_AMOUNT;
+    }
+
+    /// @inheritdoc IVaultMain
+    function getMinimumWrapAmount() external view returns (uint256) {
+        return _MINIMUM_WRAP_AMOUNT;
+    }
```
jubeira, line 1585: Hm; I've thought about this.

How much of the bytecode size change is because of these getters? I think we can't afford them unfortunately.
Maybe I'd move the immutables to the base storage layer and the getters to VaultAdmin, and then set them in the constructor of the vault extension and the main vault. It's a bit more complex (at the constructor levels at least), but it will save some bytes that we need here.

jubeira (commented):
LGTM, thanks for pushing this and all those corrections!

I only have a comment regarding the getters. I thought about this when making the pool min trade amount immutable, and I didn't want to overcomplicate things.

To save some bytes, I'd move it to the admin and propagate to the extension and the main vault (actually just to the main vault is enough). It's a bit tricky because we need to make sure that the values get to the right place, but the only extra complexity will lie in the constructors.

pkg/vault/contracts/Vault.sol
EndymionJkb: I tried that though - there seems to be an issue doing this with immutables (vs. regular storage, like all the other times we do this). I will try again; maybe I was missing something.

jubeira (approved):
LGTM! The save seems to be worth it.

PR #937, opened 2024-08-30
Implement a way to increase or decrease ERC4626 rate

joaobrunoah:
# Description

We need to increase or decrease the ERC4626 rate when testing to verify rounding issues and differentiate underlying amounts and wrapped amounts.  Suppose we only set a rate for the token in 18 decimals. In that case, we cover some rounding issues that may occur with convertToAssets and convertToShares, which use a direct division between underlying balance and total supply.

This PR creates a way to increase or decrease rates by inflating underlying and wrapped balances of the ERC4626 test token, so we can fully test the token rate.

## Type of change

- [ ] Bug fix 
- [x] New feature 
- [ ] Breaking change 
- [ ] Dependency changes
- [ ] Code refactor / cleanup
- [ ] Other

## Checklist:


pkg/solidity-utils/contracts/test/ERC4626TestToken.sol
```diff
@@ -39,6 +40,22 @@ contract ERC4626TestToken is ERC4626, IRateProvider {
         return _convertToAssets(FixedPoint.ONE, Math.Rounding.Floor);
     }
 
+    /**
+     * @notice Mints underlying and/or wrapped amount to the ERC4626 token.
+     * @dev If we set a rate to the ERC4626 token directly, we do not reproduce rounding issues when dividing assets
+     * by total supply. The best way to mock a rate is to inflate underlying and wrapped amounts.
```
EndymionJkb, line 46: I'd just add a numerical example illustrating how to use it; seems like there's more than one way. To change it to 2, should you inflate one twice as much as the other, according to their existing balances? Or set an amount only on one, leaving the other 0?

Also consider a "helper" mockRate function that calls `inflateUnderlyingOrWrapped` in a standard way to set the rate, while retaining the ability to call the inflation directly for fine-tuned control.

EndymionJkb (approved):
Going to lunch so I don't want to hold up your other PR. Looks good; see comment for some suggestions.

pkg/solidity-utils/contracts/test/ERC4626TestToken.sol
joaobrunoah: I'll add the mockRate suggestion, makes sense to use inflate only for fine-tuned control, as you mentioned

pkg/solidity-utils/contracts/test/ERC4626TestToken.sol
```diff
@@ -39,6 +41,53 @@ contract ERC4626TestToken is ERC4626, IRateProvider {
...
+        }
+        if (wrappedDelta > 0) {
+            // Mint wrapped to address 0, decreasing the token rate.
+            _mint(address(0), wrappedDelta);
+        }
+    }
+
+    /**
+     * @notice Use inflateUnderlyingOrWrapped to inflate underlying or total supply amounts, which sets the current
+     * token rate to the desired rate.
```
EndymionJkb, line 67:
```suggestion
     * @notice Set the token rate by inflating either the underlying or total supply amount.
```
Try to keep the notice to 1 line.

pkg/solidity-utils/contracts/test/ERC4626TestToken.sol
```diff
@@ -39,6 +41,53 @@ contract ERC4626TestToken is ERC4626, IRateProvider {
...
+     * @dev Although this function is a shortcut to inflateUnderlyingOrWrapped, it has a limited power: it cannot
+     * reproduce a rate that is not exact between assets and shares. So, if we want to test a rate that is problematic,
+     * we need to use the inflateUnderlyingOrWrapped and not this one.
```
EndymionJkb, line 70:
```suggestion
     * @dev Although this function is a shortcut to inflateUnderlyingOrWrapped, it has limited power: it cannot
     * reproduce a rate that is not an exact ratio between assets and shares. Use `inflateUnderlyingOrWrapped`
     * directly to test with rate values of arbitrary precision.
```

EndymionJkb (approved):
Just some comment suggestions; fine to merge after test/CI

pkg/solidity-utils/contracts/test/ERC4626TestToken.sol
```diff
@@ -39,6 +41,52 @@ contract ERC4626TestToken is ERC4626, IRateProvider {
...
+     */
+    function mockRate(uint256 newRate) external {
+        uint256 totalWrappedAmount = ERC4626TestToken(address(this)).totalSupply();
+        uint256 totalUnderlyingAmount = ERC4626TestToken(address(this)).totalAssets();
+
+        uint256 underlyingDelta;
+        uint256 wrappedDelta;
+
+        // If rate is lower than one, inflates the total supply. Else, inflates the underlying amount.
+        if (newRate < FixedPoint.ONE) {
```
jubeira, line 79: shouldn't this be the current rate?

jubeira (approved):
LGTM, just one observation to make it more generic :)

pkg/solidity-utils/contracts/test/ERC4626TestToken.sol
joaobrunoah: Yes, it should! Nice catch!

PR #938, opened 2024-08-30
Expand invariant approximation tests for weighted pool

jubeira:
# Description

This PR expands existing liquidity approximation tests for weighted pools by replicating the existing tests with zero swap fees.
Also paves the way to test against small numbers (not included in this PR; locally it looks good, but base pool math sometimes rejects small requests because of underflows, which is correct but makes the tests more complicated).

Also fixes one rounding direction for live balances after adding liquidity (no reason to round up there).

## Type of change

- [ ] Bug fix 
- [ ] New feature 
- [ ] Breaking change 
- [ ] Dependency changes
- [ ] Code refactor / cleanup
- [x] Other

## Checklist:

## Issue Resolution

N/A


pkg/vault/test/foundry/LiquidityApproximation.t.sol
```diff
@@ -199,11 +200,21 @@ contract LiquidityApproximationTest is BaseVaultTest {
 
         assertEq(dai.balanceOf(alice), dai.balanceOf(bob), "Bob and Alice DAI balances are not equal");
 
+        if (usdc.balanceOf(alice) <= defaultBalance) {
```
EndymionJkb, line 203: Can it actually be less (due to rounding, I guess)? If not, it would be self-documenting to move the `aliceAmountOut` calculation up above this line and just check for `aliceAmountOut == 0` here. If it can actually be less though, you'd have to do it this way, as otherwise you'd get an underflow calculating the difference. Here and one place below.

pkg/vault/contracts/Vault.sol
```diff
@@ -740,7 +740,7 @@ contract Vault is IVaultMain, VaultCommon, Proxy {
             poolData.updateRawAndLiveBalance(
                 i,
                 poolData.balancesRaw[i] + amountInRaw - locals.totalFeesRaw,
-                Rounding.ROUND_UP
+                Rounding.ROUND_DOWN
```
EndymionJkb, line 743: Should we just always round down when computing live balances? This is currently rounded up in only one place: tokenIn on a swap (Vault.sol: L461). Raw amounts were already calculated rounding up.

If so, we could just remove the rounding argument and save some bytes/gas.

This rounding has no effect on the token transfers; those use the raw values. It could only have an impact on future yield fees, which use the lastLive balances.

EndymionJkb (approved):
LGTM; just a couple questions in the comments.

pkg/vault/test/foundry/LiquidityApproximation.t.sol
jubeira: It can be less for very small amounts (this PR just paves the way to fuzz small ranges harder). Might be related to this: https://github.com/balancer/balancer-v3-monorepo/pull/946.

Indeed, I've implemented the test this way to avoid underflows downstream.

pkg/vault/contracts/Vault.sol
jubeira: > This rounding has no effect on the token transfers; those use the raw values. It could only have an impact on future yield fees, which use the lastLive balances.

Exactly. The thing is that this should only be used for the next operation; we've already done the main hook math to get raw results at this point.

> Should we just always round down when computing live balances? This is currently rounded up in only one place: tokenIn on a swap (Vault.sol: L461). Raw amounts were already calculated rounding up.

That should probably be the case yes. I think I'll open a PR just with these two changes.

jubeira: > If so, we could just remove the rounding argument and save some bytes/gas.

My understanding is that there's one case where it does make sense to round up: in `addLiquidity` before the main hook. Rounding up there helps to compute less BPT out, as stated in the comments: https://github.com/balancer/balancer-v3-monorepo/blob/fb2753a76cf3cb650c2adf412735d84d931b67af/pkg/vault/contracts/Vault.sol#L518-L531.

If that's the case, we can't really remove the direction from the function.

jubeira: I've removed this change from here in favor of #947

joaobrunoah (approved):
LGTM!

PR #939, opened 2024-08-31
Remove vault dependencies in solidity-utils

EndymionJkb:
# Description

This started out as noticing that Rounding might not need to be in the Vault. In the process, we noted several places where solidity-utils contracts were dependent on the Vault (even if only through interfaces, so not technically circular). Still, this isn't ideal.

The main driver of this seemed to be putting BasePoolMath in solidity-utils (while IBasePool and friends were in the Vault). Assuming we still want BasePoolMath down in utils, the easiest thing was to move IBasePool to solidity-utils.

Along with Rounding, some basic types (e.g., TokenType, PoolSwapParams, SwapKind) were then referenced in solidity-utils. So a simple thing was to create a BasePoolTypes file similar to VaultTypes, and move those few that were needed at a lower level out of VaultTypes.

The problem then was SingletonAuthentication - which needed IVault and `getAuthorizer`. There was a way to handle this, but it involved adding a new interface, changing IVaultMain, changing function signatures, adding typecasts, etc. It got messy fast, and I didn't like it. It was far simpler to go the other way, and move SingletonAuthentication up to the Vault, where importing IVault was fine. This solved all the issues and didn't require other changes.

There now should be nothing in solidity-utils that references the Vault. The diff is 95% imports; there aren't many code changes.

## Type of change

- [ ] Bug fix 
- [ ] New feature 
- [ ] Breaking change 
- [ ] Dependency changes
- [X] Code refactor / cleanup
- [ ] Other

## Checklist:


jubeira (commented):
Thanks for pushing @EndymionJkb.

Looking at the changes I'm having second thoughts about this. This would move things related to pools and stuff like that to `solidity-utils`, which is supposed to be really basic dependencies and building blocks. Doesn't feel entirely right.

The new base pool types in solidity utils has some struct definitions for swaps, but not for add / remove.

> The main driver of this seemed to be putting BasePoolMath in solidity-utils (while IBasePool and friends were in the Vault). Assuming we still want BasePoolMath down in utils, the easiest thing was to move IBasePool to solidity-utils.

Perhaps this is the problem? `BasePoolMath` is kind of a legacy term at this point. Truth is that this is just code that is part of the Vault now, only that is separated into a library for convenience.

What if we do it the other way around? I.e. move `BasePoolMath` to `Vault`, and just move `Rounding` to solidity utils? (or maybe even leave it there? I'm open to input here).

EndymionJkb:
Yeah; I wasn't thrilled about the diff, even after simplifying as much as
possible. Will try it another way. I did already move some things up

PR #940, opened 2024-09-01
Remove vault dependencies in solidity-utils (v2)

EndymionJkb:
# Description

Still a giant diff (again, mostly imports), this takes a different approach, and moves BasePoolMath up into the Vault. This means fewer things need to go down a level to solidity-utils. We still need a BasePoolTypes, but it only has Rounding and Token stuff: nothing about swaps or pool operations. This feels cleaner, even though there's more stuff in the Vault than before, and ideally we were trying to move things *out* of the Vault.

This should also achieve the other goal of not having "circular" dependencies - no references to the Vault (albeit just interfaces) in solidity-utils.

## Type of change

- [ ] Bug fix 
- [ ] New feature 
- [ ] Breaking change 
- [ ] Dependency changes
- [X] Code refactor / cleanup
- [ ] Other

## Checklist:


EndymionJkb:
Superseded by #942

PR #943, opened 2024-09-02
Update Solidity to 0.8.26

jubeira:
# Description

Update every contract but `VaultExplorer` (see https://github.com/ethereum/solidity/issues/15004) to build with 0.8.26, using the default optimizer sequence from 0.8.25 (see [release announcement](https://soliditylang.org/blog/2024/05/21/solidity-0.8.26-release-announcement)).

Gas and bytecode looks better. Waiting for 0.8.27 this week, which should fix `VaultExplorer` as well.

## Type of change

- [ ] Bug fix 
- [ ] New feature 
- [ ] Breaking change 
- [x] Dependency changes
- [ ] Code refactor / cleanup
- [ ] Other

## Checklist:

## Issue Resolution

N/A


joaobrunoah (approved):
Just to make sure, we should not change the pragma of all contracts, right? If that's the case, LGTM.

EndymionJkb:
> Just to make sure, we should not change the pragma of all contracts, right? If that's the case, LGTM.

Also because we're hopefully moving to 0.8.27 right away; we can change it then. (We would truly be dependent on 0.8.27, since the code wouldn't compile otherwise, especially if we can remove the VaultExplorer exception.)

A bit dangerous to be on the bleeding edge of the compiler. Looks like Uni is already on 0.8.26 though.

jubeira:
> Just to make sure, we should not change the pragma of all contracts, right? 

I don't think that's needed. The code does compile with .24, and that's the minimum required because of transient storage for now.

If we start using `require` with custom errors we should update the pragma for example.

PR #945, opened 2024-09-03
Add missing comment on _settleWrap and _settleUnwrap

joaobrunoah:
# Description

Auditors skipped an important row when reading `_wrapWithBuffer` and `_unwrapWithBuffer`. This PR makes sure these rows are explained and not easy to miss.

## Type of change

- [ ] Bug fix 
- [ ] New feature 
- [ ] Breaking change 
- [ ] Dependency changes
- [ ] Code refactor / cleanup
- [x] Documentation or wording changes
- [ ] Other

## Checklist:

### DIFF
--- a/pkg/vault/contracts/Vault.sol
+++ b/pkg/vault/contracts/Vault.sol
@@ -1489,9 +1489,9 @@ contract Vault is IVaultMain, VaultCommon, Proxy {
         if (underlyingBalancesAfter < expectedUnderlyingReservesAfter) {
             // If Vault's underlying balance is smaller than expected, the Vault was drained and the operation should
             // revert. It may happen in different ways, depending on the wrap/unwrap operation:
-            // * deposit: the wrapper didn't respect the exact amount in of underlying.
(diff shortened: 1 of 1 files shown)


pkg/vault/contracts/Vault.sol
```diff
@@ -1224,6 +1224,8 @@ contract Vault is IVaultMain, VaultCommon, Proxy {
             // drain the vault.
             underlyingToken.forceApprove(address(wrappedToken), 0);
 
+            // Check if the Vault's underlying balance decreased by `vaultUnderlyingDeltaHint` and the Vault's
+            // wrapped balance increased by `vaultWrappedDeltaHint`. If not, reverts.
```
EndymionJkb, line 1228:
```suggestion
            // wrapped balance increased by `vaultWrappedDeltaHint`. If not, it reverts.
```
"it reverts" or "revert"; former sounds more natural. Could say how it reverts, but I believe it could revert with either in both cases, so that would probably make it more confusing.

EndymionJkb (approved):
Definitely an improvement, but I think we could add even more.

PR #946, opened 2024-09-03
Bypass undo-scaling for exact amounts

jubeira:
# Description

Exact amounts in / out are not exact to the very last wei when there are rounding issues. In other words, in some cases `unscale(scale(amount)) != amount`.

This PR bypasses unscaling when there's a raw amount given that needs to be respected.

## Type of change

- [x] Bug fix 
- [ ] New feature 
- [ ] Breaking change 
- [ ] Dependency changes
- [ ] Code refactor / cleanup
- [ ] Other

## Checklist:

## Issue Resolution

N/A


pkg/vault/contracts/Vault.sol
```diff
@@ -647,6 +648,7 @@ contract Vault is IVaultMain, VaultCommon, Proxy {
             poolData.poolConfigBits.requireUnbalancedLiquidityEnabled();
 
             amountsInScaled18 = maxAmountsInScaled18;
+            amountsInRaw = params.maxAmountsIn;
```
jubeira, line 651: Note that here we're discarding the original array and pointing to `maxAmountsIn`.

In this case `maxAmountsIn` should not be modified. It might be better to just make `amountsInRaw` a copy of `maxAmountsIn` instead.

jubeira: I've added a simple deep-copy function for arrays.

The compiler doesn't seem to be optimizing the loop, even in 0.8.26. `mcopy` could be an alternative to save some bytes, but I don't want to optimize early.

pkg/vault/contracts/Vault.sol
```diff
@@ -647,6 +648,10 @@ contract Vault is IVaultMain, VaultCommon, Proxy {
             poolData.poolConfigBits.requireUnbalancedLiquidityEnabled();
 
             amountsInScaled18 = maxAmountsInScaled18;
+            // Deep copy given max amounts in raw to calculated amounts in raw to avoid scaling later, ensuring that
+            // `maxAmountsIn` is preserved.
+            ScalingHelpers.copyToArray(params.maxAmountsIn, amountsInRaw);
+
```
joaobrunoah, line 654: Why do we need to do a deep copy here? the amountsInRaw values are not edited, unless it's 0 (which would probably not enter this `if` clause anyway).

jubeira: True, we don't strictly need it, but you need to reason about that to figure out that you're not really modifying `maxAmountsIn`. Sounds a bit safer to just do a copy.

EndymionJkb: And I guess this is the only place, since we don't have a remove with exact (multiple) tokens out.

pkg/vault/test/foundry/VaultLiquidityRate.t.sol
```diff
@@ -213,4 +213,25 @@ contract VaultLiquidityWithRatesTest is BaseVaultTest {
             bytes("")
         );
     }
+
+    function testRemoveLiquiditySingleTokenExactOutWithRate__Fuzz(uint256 wstEthRate, uint256 wstEthAmountOut) public {
```
EndymionJkb, line 217: This test passes either way, but doesn't hurt to add. We'd need to have a Vault mock that overrode and mutated the amountsIn to "really" test this, I guess, but that seems like overkill for a simple thing.

joaobrunoah (approved):
LGTM!

EndymionJkb (approved):
Per offline discussion; this seems to be the best way to handle this

pkg/vault/contracts/Vault.sol
jubeira: Correct.
More so, unless we create the array inside each the block corresponding to each kind, we've already allocated the memory by this point. The copy only loops and assigns the values from the given argument to the new, empty array.

pkg/vault/test/foundry/VaultLiquidityRate.t.sol
jubeira: Actually it can be done in existing tests: https://github.com/balancer/balancer-v3-monorepo/pull/946/commits/87de278d4a8844b5e099bb80dae61bd3ab60cd6a.

It's an overkill, but I'd say it's worth it. Thanks for the idea.

pkg/solidity-utils/contracts/helpers/ScalingHelpers.sol
```diff
@@ -155,6 +155,15 @@ library ScalingHelpers {
                                     Array Functions
     ***************************************************************************/
 
+    function copyToArray(uint256[] memory from, uint256[] memory to) internal pure {
+        uint256 length = from.length;
```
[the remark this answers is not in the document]
jubeira: yes, but I didn't want to go into assembly and optimize early. Unfortunately the compiler doesn't seem to be optimizing this part on its own.

PR #947, opened 2024-09-03
Round down scaled balances after main operation (swap / add)

jubeira:
# Description

We're rounding pool balances up after swap and add, but it really doesn't make much sense to do so.

This should only affect the fees on the next operation, since the math for the main hook (`onSwap`, `onAdd`) has already been computed, and this only affects last live balances for pools (no relationship with tokens in / out). It shouldn't really have any impact beyond it, so flagging as cleanup.

## Type of change

- [ ] Bug fix 
- [ ] New feature 
- [ ] Breaking change 
- [ ] Dependency changes
- [x] Code refactor / cleanup
- [ ] Other

## Checklist:

## Issue Resolution

N/A


joaobrunoah (approved):
LGTM!

EndymionJkb (approved):
Looks good; too bad we can't remove it entirely. Everytime it's called by itself it's rounding down, but it's also called from the main `load`, which does need to round up in one case, passing the direction down to `updateRawAndLiveBalance`

PR #948, opened 2024-09-03
Test RemoveLiquiditySingleTokenExactIn to ensure the invariant does not decrease with rates != 1

joaobrunoah:
# Description

Swap operations had an issue when rounding token out amounts (given or calculated). So, we understood that we should round up the rate when upscaling or downscaling a token out amount, ensuring that any rounding error related to the rate calculation is paid by the user, and the Vault cannot be drained.

Since remove liquidity also has downscaling of token amounts out, we wanted to make sure that the same issue does not happen for these kind of operations.

This PR merges a test to removeLiquiditySingleTokenExactIn that makes sure rate changes cannot drain a pool.

## Type of change

- [ ] Bug fix 
- [ ] New feature 
- [ ] Breaking change 
- [ ] Dependency changes
- [ ] Code refactor / cleanup
- [x] Other: Test

## Checklist:


pkg/vault/test/foundry/VaultLiquidityRate.t.sol
```diff
@@ -213,4 +223,67 @@ contract VaultLiquidityWithRatesTest is BaseVaultTest {
...
+        // Considering a pool with exactly poolInitAmount of each token, we need to add tokens in the side where the
+        // balance scaled18 is smaller. If rate of WstETH is 1, and DAI is 2, we have less WstETH than DAI, so we need
+        // to add more WstETH. The amount of WstETH to add is `(daiBalance * rateDai / rateWstEth - wstEthBalance)`.
+        // Since daiBalance and wstEthBalance is poolInitAmount, this formula becomes
+        // `poolInitAmount * (rateDai / rateWstEth - 1)`.
+        uint256[] memory amountsToAdd = new uint256[](2);
+        if (rateWstEth < rateDai) {
+            amountsToAdd[wstethIdx] = poolInitAmount * ((rateDai / rateWstEth) - 1);
+        } else {
+            amountsToAdd[daiIdx] = poolInitAmount * ((rateWstEth / rateDai) - 1);
```
jubeira, line 249: Hmm I think there's a mistake here.

The formula is correct, but in FP math this should probably be (e.g. for `if` clause):

```
poolInitAmount.mulDown(rateDai.divDown(rateWstEth) - 1e18)
```

There's no need to simplify further given than this is a test. I'd just use FP math with no shortcuts (it's a multiplication and a division after all so it can be done without FP math, but the `-1` will make it confusing.

jubeira: Doing the division first and subtracting a single wei is OK on paper, but you're losing a lot of precision here.

jubeira (approved):
LGTM, thanks @joaobrunoah! I think one of the formulas can be improved considering FP math; PTAL.

pkg/vault/test/foundry/VaultLiquidityRate.t.sol
joaobrunoah: Yeap, totally right.. I was multiplying and dividing first, which didn't lose precision. But I needed to remove poolInitAmount from the result, so I did the math simplification and wrote in the code as I thought... Anyway, I'll change it.

elshan-eth (approved):
LGTM!

PR #951, opened 2024-09-04
Global protocol fee precision check

EndymionJkb:
# Description

Per auditor comment, we are checking precision of the *aggregate* fees, but not global fees. This lets global fees be set that could force creator fees to be set to only 0% or 100%. Of course, these are permissioned calls, and it could always be undone, but it gives governance an unintended degree of control over pool creator fees, so I think it should be fixed.

## Type of change

- [X] Bug fix 
- [ ] New feature 
- [ ] Breaking change 
- [ ] Dependency changes
- [ ] Code refactor / cleanup
- [ ] Other

## Checklist:


jubeira (approved):
LGTM!

PR #952, opened 2024-09-04
Remove regions comments

elshan-eth:
# Description

Region comments don't work for all. This PR removes all these comments.

## Type of change

- [ ] Bug fix 
- [ ] New feature 
- [ ] Breaking change 
- [ ] Dependency changes
- [ ] Code refactor / cleanup
- [ ] Other

## Checklist:

### DIFF
--- a/pkg/vault/contracts/lib/HooksConfigLib.sol
+++ b/pkg/vault/contracts/lib/HooksConfigLib.sol
@@ -35,7 +35,7 @@ library HooksConfigLib {
     using WordCodec for bytes32;
     using HooksConfigLib for PoolConfigBits;
 
-    // #region Bit offsets for hooks config
+    // Bit offsets for hooks config
 
     function enableHookAdjustedAmounts(PoolConfigBits config) internal pure returns (bool) {
         return PoolConfigBits.unwrap(config).decodeBool(PoolConfigConst.ENABLE_HOOK_ADJUSTED_AMOUNTS_OFFSET);
@@ -167,9 +167,7 @@ library HooksConfigLib {
             });
     }
 
-    // #endregion
-
-    // #region Hooks helper functions
+    // Hooks helper functions
 
     /**
      * @dev Call the `onComputeDynamicSwapFeePercentage` hook and return the result. Reverts on failure.
@@ -491,6 +489,4 @@ library HooksConfigLib {
             revert IVaultErrors.AfterInitializeHookFailed();
         }
     }
-
-    // #endregion
 }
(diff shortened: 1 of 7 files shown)


pkg/vault/contracts/lib/HooksConfigLib.sol
```diff
@@ -35,7 +35,7 @@ library HooksConfigLib {
     using WordCodec for bytes32;
     using HooksConfigLib for PoolConfigBits;
 
-    // #region Bit offsets for hooks config
+    // Bit offsets for hooks config
```
EndymionJkb, line 38:
```suggestion
    // Bit offsets for hooks config.
```

EndymionJkb: Here and elsewhere - it was like this before, but non-inline comments should end in a period.

pkg/vault/test/foundry/unit/ERC20MultiTokenTest.t.sol
```diff
@@ -27,7 +27,7 @@ contract ERC20MultiTokenTest is Test, IERC20Errors, ERC20MultiToken {
         token = new ERC20MultiTokenMock();
     }
 
-    // #region Init values
+    // Init values
```
EndymionJkb, line 30:
```suggestion
    // Initialize values
```
Might as well spell it out

pkg/vault/test/foundry/unit/ERC20MultiTokenTest.t.sol
```diff
@@ -36,9 +36,7 @@ contract ERC20MultiTokenTest is Test, IERC20Errors, ERC20MultiToken {
         assertEq(token.totalSupply(POOL), 0, "Unexpected total supply");
     }
 
-    // #endregion
-
-    // #region Approve & Allowance & SpendAllowance
+    // Approve & Allowance & SpendAllowance
```
EndymionJkb, line 39: Since this is a test, I don't think you need this comment at all without it being a region header. Same with QueryModeBalanceIncrease. If the comment is just the name of the function, we don't need it (especially in tests).

Here and elsewhere.

pkg/vault/test/foundry/unit/ERC20MultiTokenTest.t.sol
```diff
@@ -284,9 +272,7 @@ contract ERC20MultiTokenTest is Test, IERC20Errors, ERC20MultiToken {
         token.manualTransfer(POOL, OWNER, OWNER2, transferAmount);
     }
 
-    // #endregion
-
-    // #region Private functions
+    // Private functions
```
EndymionJkb, line 275: Here again, I think this can just be removed. We used to do this sometimes in v2, but in v3 we haven't been labelling groups by visibility. We use big block headers /****. ****/ in code, but I don't think we need those in tests.

EndymionJkb (approved):
Searched to ensure they're all gone. Some suggestions, but I think removing these improves consistency / readability. Certainly a lot easier than putting them in everywhere else.

PR #954, opened 2024-09-04
Solidity 0.8.27

jubeira:
# Description

Not much difference wrt. 0.8.26 in terms of the overall result really.
The main difference is that it does compile `VaultExplorer` as expected. Other than that, we still need the original optimization sequence.

Another observation is that if we were to change reverts for require we would need this to compile forge tests properly (in 0.8.26 require with custom errors does not work with the legacy pipeline).

## Type of change

- [ ] Bug fix 
- [ ] New feature 
- [ ] Breaking change 
- [ ] Dependency changes
- [ ] Code refactor / cleanup
- [ ] Other

## Checklist:


jubeira:
Not happening; thanks for participating.

PR #955, opened 2024-09-05
SafeCast in Protocol Fee Controller and Router

EndymionJkb:
# Description

Per the audit comment, this uses SafeCast everywhere we were doing simple truncation casting (where the type could possibly overflow).

I don't believe it can *actually* overflow in the Fee Controller, since the values are bound elsewhere, but it does create a dependency. I can make it happen in the BatchRouter (e.g., initial token transfer) or `_takeTokenIn`, though it would probably blow up elsewhere at some point later, such as settlement. I don't think it can happen in the Router, as the permit2 operations are done at the end, and any values that would overflow would blow up when we try to store the balances. Doing it everywhere for consistency, though it does cost ~100 bytes and 100-200 gas.

## Type of change

- [X] Bug fix 
- [ ] New feature 
- [ ] Breaking change 
- [ ] Dependency changes
- [ ] Code refactor / cleanup
- [ ] Other

## Checklist:


jubeira (commented):
Agreed on the admin functions; there's not much to lose over there.

Not 100% sold on the routers. I'm not a huge fan of spending more gas over this, although it might be OK. It's one more jump + an if; a check can't really get any cheaper than that. Thoughts?

jubeira (approved):
Ok I'm sold via offchain thread.

PR #960, opened 2024-09-05
Optimize VaultFactory

### LINKED ISSUE #176: Vault deployer contract
In V2 we have this profanity address that generates a pretty address in the 5th nonce -which is an arguably brittle mechanism-. The private key was compromised, which is super annoying when deploying to new chains.

For V3 we should resort to create2 or create3, and deploy the Vault via a deployer contract that can generate another pretty address without depending on the EOA that triggers the transaction.

elshan-eth:
# Description

Reducing the size of VaultFactory by moving bytecode to arguments. 

## Type of change

- [ ] Bug fix 
- [ ] New feature 
- [ ] Breaking change 
- [ ] Dependency changes
- [ ] Code refactor / cleanup
- [X] Optimization: [ ] gas / [ ] bytecode
- [] Other

## Checklist:

## Issue Resolution

Closes #176


pkg/vault/contracts/VaultFactory.sol
```diff
@@ -64,28 +74,67 @@ contract VaultFactory is Authentication {
...
 
-        VaultAdmin vaultAdmin = new VaultAdmin(
-            IVault(vaultAddress),
-            _pauseWindowDuration,
-            _bufferPeriodDuration,
-            _minTradeAmount,
-            _minWrapAmount
+        VaultAdmin vaultAdmin = VaultAdmin(
+            payable(
+                Create2.deploy(
```
joaobrunoah, line 99: It'd be nice to have a comment on why we're using payable and `create2` here.

pkg/vault/contracts/VaultFactory.sol
```diff
@@ -64,28 +74,67 @@ contract VaultFactory is Authentication {
...
+                    bytes32(uint256(0x01)),
+                    abi.encodePacked(vaultExtensionCreationCode, abi.encode(vaultAddress, vaultAdmin))
+                )
+            )
+        );
+
         ProtocolFeeController feeController = new ProtocolFeeController(IVault(vaultAddress));
 
-        address deployedAddress = _create(abi.encode(vaultExtension, _authorizer, feeController), salt);
+        require(vaultCreationCodeHash == keccak256(vaultCreationCode), "VaultFactory: invalid vaultCreationCode");
```
joaobrunoah, line 132: Are we still using `if - revert` with custom error? I remember we could have an issue with the compiler if we use require with custom error, but I'm not sure if we should use a require with a custom text. @EndymionJkb ?

joaobrunoah (commented):
Hey Elshan, nice! I have a question about the deployment process, I'm not sure if it'll get more complicated than today with the extra 3 parameters, but if it's easy to fetch them with the deployment script, then it's all good.

Since this PR focus on reducing the bytecode of VaultFactory, should we create a test to measure that in hardhat?

pkg/vault/contracts/VaultFactory.sol
EndymionJkb: I guess because we don't care what the address is for the proxy contracts, only the main Vault?

pkg/vault/contracts/VaultFactory.sol
EndymionJkb: We use require with text for "preconditions" in tests, or things that shouldn't happen. In production code, for things that *could* happen, we should use custom errors.

I guess this is technically production code, though it's only used by us for deployments, not by regular users. We already have custom errors defined, so the question is whether this is a pre/post condition that should never be violated, or it's something that could reasonably happen. It shouldn't if you pass the right arguments, but it would be the only one, so I'd lean toward another custom error, though I don't feel strongly.

I tried using require with custom errors (which only works in 0.8.27). It increased the bytecode fairly dramatically (hundreds of bytes); not sure why. So I think we'll very likely stay with our current method, even if we go to 0.8.27. (We'll probably just stay on 0.8.26, though.)

pkg/vault/contracts/VaultFactory.sol
elshan-eth: yes, we don't care about addresses for the proxy contracts

elshan-eth:
> Since this PR focus on reducing the bytecode of VaultFactory, should we create a test to measure that in hardhat?

@joaobrunoah, I think no because the main task was to ensure that the factory did not exceed the limit

joaobrunoah:
> > Since this PR focus on reducing the bytecode of VaultFactory, should we create a test to measure that in hardhat?
> 
> @joaobrunoah, I think no because the main task was to ensure that the factory did not exceed the limit

Got it! But, if we need to add something to VaultFactory in the future, aren't we at risk of increasing the code above the limit again? I mean, for maintainability, maybe it'd be nice to measure the contract size (in short, if we needed to reduce now, maybe we will need to reduce in the future)

pkg/vault/contracts/VaultFactory.sol
```diff
@@ -64,28 +74,67 @@ contract VaultFactory is Authentication {
...
         }
 
-        VaultAdmin vaultAdmin = new VaultAdmin(
-            IVault(vaultAddress),
-            _pauseWindowDuration,
-            _bufferPeriodDuration,
-            _minTradeAmount,
-            _minWrapAmount
+        VaultAdmin vaultAdmin = VaultAdmin(
+            payable(
```
joaobrunoah, line 98: If I got right, the vault should not receive eth from any contract. Should VaultAdmin and VaultExtension be payable?

pkg/vault/contracts/VaultFactory.sol
```diff
@@ -64,28 +74,67 @@ contract VaultFactory is Authentication {
...
+                        )
+                    )
+                )
+            )
+        );
+
+        require(
+            vaultExtensionCreationCodeHash == keccak256(vaultExtensionCreationCode),
+            "VaultFactory: invalid vaultExtensionCreationCode"
         );
```
jubeira, line 119: Aren't we already checking this above?

pkg/vault/contracts/VaultFactory.sol
jubeira, line 132: Same; aren't we already checking this at the beginning of the function?

pkg/vault/contracts/VaultFactory.sol
```diff
@@ -64,28 +74,67 @@ contract VaultFactory is Authentication {
      * @param targetAddress Expected Vault address. The function will revert if the given salt does not deploy the
      * Vault to the target address.
      */
-    function create(bytes32 salt, address targetAddress) external authenticate {
-        if (isDisabled) {
```
jubeira, line 68: I think I'd keep the flag. We don't want to deploy more than once.

pkg/vault/contracts/VaultFactory.sol
```diff
@@ -64,28 +74,67 @@ contract VaultFactory is Authentication {
...
-            revert VaultAlreadyCreated();
+    function create(
+        bytes32 salt,
+        address targetAddress,
+        bytes memory vaultCreationCode,
+        bytes memory vaultAdminCreationCode,
+        bytes memory vaultExtensionCreationCode
+    ) external authenticate {
```
jubeira, line 83: Not strictly related to this PR, but is there a reason to keep `authenticate`?

The factory should be fool-proof, meaning that it should only be able to deploy the right contracts in the right order with the right argument to the right addresses.

I think that's already (almost?) the case:
- You are storing the hashes for the creation code when the factory is deployed, which means you can only deploy the right contracts.
- The arguments are stored upon creation time. Maybe we could even make them constants so as to prevent construction time mistakes.
- The final Vault address is stored and checked against the actual deployment, so there's no possibility to deploy it elsewhere.

Am I missing something?

pkg/vault/contracts/VaultFactory.sol
```diff
@@ -64,28 +74,67 @@ contract VaultFactory is Authentication {
...
      */
-    function create(bytes32 salt, address targetAddress) external authenticate {
-        if (isDisabled) {
-            revert VaultAlreadyCreated();
+    function create(
+        bytes32 salt,
+        address targetAddress,
+        bytes memory vaultCreationCode,
+        bytes memory vaultAdminCreationCode,
+        bytes memory vaultExtensionCreationCode
```
jubeira, line 82: Perhaps it's worth using `calldata` here?

jubeira (approved):
LGTM, thanks @elshan-eth ! Factory can now be deployed.

Just a few comments and observations; other than that I think this makes sense overall.

pkg/vault/contracts/VaultFactory.sol
EndymionJkb: Should save a little, and be insurance against it changing (though it clearly doesn't get changed)

pkg/vault/contracts/VaultFactory.sol
EndymionJkb: It also just seems weird to need governance approval to deploy the Vault

EndymionJkb (approved):
Pending all the comments; looks good!

pkg/vault/contracts/VaultFactory.sol
elshan-eth: create2 doesn't allow the deployment of contracts again. We don't need a flag to protect double deployment

pkg/vault/contracts/VaultFactory.sol
elshan-eth: It seems that if you remove an authorization, someone can perform a front-run attack. For example, we deploy a contract with salt and someone intercepts it and deploys it with their arguments.

pkg/vault/contracts/VaultFactory.sol
```diff
@@ -26,7 +26,12 @@ contract VaultFactory is Authentication {
     /// @notice The given salt does not match the generated address when attempting to create the Vault.
     error VaultAddressMismatch();
 
-    bool public isDisabled;
```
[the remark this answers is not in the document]
elshan-eth: create2 doesn't allow the deployment of contracts again. We don't need a flag to protect double-deployment

elshan-eth:
> Got it! But, if we need to add something to VaultFactory in the future, aren't we at risk of increasing the code above the limit again? I mean, for maintainability, maybe it'd be nice to measure the contract size (in short, if we needed to reduce bytecode now, maybe we will need to reduce bytecode in the future again, and having this measurement of bytecode size would point this out sooner)

I think with the current approach, such a factory should never become larger than the bytecode limit.

pkg/vault/contracts/VaultFactory.sol
elshan-eth: If it should be access-free, then the salt should be saved in the constructor. Otherwise, anyone can call with any salt and the factory will no longer be able to deploy other contracts.

pkg/vault/contracts/VaultFactory.sol
elshan-eth: the compiler requires it to be payable

PR #965, opened 2024-09-08
Separate liquidity flags

### LINKED ISSUE #957: Consider disabling unbalanced add/remove separately
It could be useful to disable only unbalanced removes, so that hooks could charge a fee on exit safely, yet still support unbalanced adds.

EndymionJkb:
# Description

By default, the Vault supports all non-custom add/remove liquidity types: single token, unbalanced, and proportional. However, there are use cases where unbalanced liquidity operations may be unsafe or unsupported.

For instance, pricing in pools that support donation could be manipulated if it were possible to add or remove liquidity unbalanced. Also, if the hook enables adjusting swap deltas, it cannot be registered to a pool that supports unbalanced liquidity, or the fees (always on the calculated amount), might have to be calculated in BPT, which isn't supported. (The Vault prevents this.)

However, this check might be too restrictive. Hooks with exit fees should enforce proportional withdrawals to ensure proper fee accounting: but might still allow unbalanced adds, as this is the default in the UI and would be most convenient. This PR splits the `disableUnbalancedLiquidity` flag in LiquidityManagement into two separate flags: `disableAddLiquidityUnbalanced` and `disableRemoveLiquidityUnbalanced`.

## Type of change

- [ ] Bug fix 
- [X] New feature 
- [ ] Breaking change 
- [ ] Dependency changes
- [ ] Code refactor / cleanup
- [ ] Other

## Checklist:

## Issue Resolution

Closes #957.


pkg/pool-hooks/test/foundry/ExitFeeHookExample.t.sol
```diff
@@ -58,7 +58,8 @@ contract ExitFeeHookExampleTest is BaseVaultTest {
         roleAccounts.poolCreator = lp;
 
         LiquidityManagement memory liquidityManagement;
-        liquidityManagement.disableUnbalancedLiquidity = true;
+        liquidityManagement.disableAddLiquidityUnbalanced = true;
```
jubeira, line 61: This one can actually be enabled right? That's the main use-case for this :)

pkg/vault/contracts/VaultExtension.sol
```diff
@@ -284,7 +287,8 @@ contract VaultExtension is IVaultExtension, VaultCommon, Proxy {
                 // input amount (EXACT_OUT).
                 if (
                     hookFlags.enableHookAdjustedAmounts &&
-                    params.liquidityManagement.disableUnbalancedLiquidity == false
+                    (params.liquidityManagement.disableAddLiquidityUnbalanced == false ||
+                        params.liquidityManagement.disableRemoveLiquidityUnbalanced == false)
```
jubeira, line 291: hm.
The problem is that we want to enable this when the adjusted amount is only for one operation.

E.g. we can enable after exit adjusted amounts with unbalanced adds. Which means that we'd need to also split `hookAdjustedAmounts` flag as well, unless I'm missing something.

jubeira (commented):
Thanks for putting this together @EndymionJkb !

Overall looks good for the most part, but I think we'll need to split `hookAdjustedAmounts` flag in two as well to make this work :/

pkg/pool-hooks/test/foundry/ExitFeeHookExample.t.sol
EndymionJkb: Yes, but I didn't want also change the example in this PR (and also update the tests, etc.)

pkg/vault/contracts/VaultExtension.sol
EndymionJkb: We could have hookAdjustedAmountsIn and hookAdjustedAmountsOut. We also check this on swaps, though... do we need three?

pkg/pool-weighted/test/gas/.hardhat-snapshots/[WeightedPool - Standard - BatchRouter] swap exact in with one token and fees - cold slots
```diff
@@ -1 +1 @@
-186.1k
\ No newline at end of file
+186.5k
```
Comment (jubeira) [inline pkg/pool-weighted/test/gas/.hardhat-snapshots/[WeightedPool - Standard - BatchRouter] swap exact in with one token and fees - cold slots:1]: :face_exhaling:

pkg/vault/contracts/Vault.sol
```diff
@@ -646,7 +646,7 @@ contract Vault is IVaultMain, VaultCommon, Proxy {
             bptAmountOut = 0;
             amountsInScaled18 = maxAmountsInScaled18;
         } else if (params.kind == AddLiquidityKind.UNBALANCED) {
```
jubeira, line 648: not strictly related with this PR, but what about donations? Those are also a kind of unbalanced liquidity.

pkg/vault/test/.contract-sizes/Vault
```diff
@@ -1,2 +1,2 @@
-Bytecode	23.883
-InitCode	25.224
\ No newline at end of file
+Bytecode	23.980
```
jubeira, line 1: :melting_face:

pkg/vault/test/foundry/unit/VaultExtensionUnit.t.sol
```diff
@@ -15,6 +15,10 @@ contract VaultExtensionUnitTest is BaseVaultTest {
         BaseVaultTest.setUp();
     }
 
+    function testItWorks() public pure {
+        assertTrue(true);
+    }
```
jubeira, line 20: remove?

jubeira (commented):
Thanks @EndymionJkb, I think this looks good.

I don't fully buy the extra complexity and gas that this brings tbh. Let's discuss with the rest of the team whether this is worth it or not now that we have a solid implementation and some numbers on the table. Great job!

pkg/vault/test/foundry/unit/VaultExtensionUnit.t.sol
EndymionJkb: Ha :) Yes, I do this when debugging an error in the setup.

pkg/vault/contracts/Vault.sol
EndymionJkb: Not sure what you're asking here. We have separate types and flags for donations and regular add unbalanced, and the flags are specific to the operation.

EndymionJkb:
Consensus was this was too complex; close as not planned.

PR #969, opened 2024-09-10
Overhaul liquidity approximation tests for weighted pool

jubeira:
# Description

Improve `LiquidityApproximationWeighted` by restructuring the test and expanding the coverage.

In the current test, the inputs are bounded at several levels, which complicates the analysis and in some cases broke certain assumptions (e.g. we were fuzzing swap fees when they should have been 0, and in some cases bounding amounts in / out more than once).

This PR moves input bounds to the outer levels (i.e. entrypoints / functions that start with `test`) to ensure that it only happens once. It also repeats the existing tests for very small amounts without swap fees to verify that the math is still sound (or it that it reverts because of rounding).

## Type of change

- [ ] Bug fix 
- [ ] New feature 
- [ ] Breaking change 
- [ ] Dependency changes
- [ ] Code refactor / cleanup
- [x] Other

## Checklist:

## Issue Resolution

N/A


pkg/pool-weighted/test/foundry/LiquidityApproximationWeighted.t.sol
```diff
@@ -71,47 +72,84 @@ contract LiquidityApproximationWeightedTest is LiquidityApproximationTest {
 
     // Tests varying weight
 
+    // #region addLiquidityUnbalanced
```
EndymionJkb, line 75: These region tags were removed in another PR; shouldn't reintroduce here. Here and elsewhere.

pkg/pool-weighted/test/foundry/LiquidityApproximationWeighted.t.sol
```diff
@@ -126,159 +164,286 @@ contract LiquidityApproximationWeightedTest is LiquidityApproximationTest {
...
+        uint256 exactAmountOut,
+        uint256 weightDai
+    ) public {
+        exactAmountOut = bound(exactAmountOut, 1, 1e6);
+        weightDai = bound(weightDai, 1e16, 99e16);
+
+        try this.removeLiquiditySingleTokenExactOutWeights(exactAmountOut, 0, weightDai) {
+            // OK, test passed.
+        } catch (bytes memory result) {
+            assertEq(bytes4(result), bytes4(stdError.arithmeticError), "Unexpected error");
```
EndymionJkb, line 328: Would also add a comment here, like "Can legitimately fail due to arithmetic underflow; live system will be protected by minimum amounts." Could even point to the line where it would be expected to fail, if we're close enough that we don't expect maintenance issues with pointing to specific lines.

EndymionJkb: You have a similar comment below at L397: 

  // For very small invariant ratios, `BasePoolMath` reverts when calculated amount out < 0 because of rounding.
  // Perform an external call so that `expectRevert` catches the error.

EndymionJkb (approved):
LGTM

pkg/pool-weighted/test/foundry/LiquidityApproximationWeighted.t.sol
```diff
@@ -126,159 +164,286 @@ contract LiquidityApproximationWeightedTest is LiquidityApproximationTest {
...
+
+    function testAddLiquidityProportionalAndRemoveExactOutWeightsSmallAmounts__Fuzz(
+        uint256 exactBptAmountOut,
+        uint256 weightDai
+    ) public {
+        exactBptAmountOut = bound(exactBptAmountOut, 1, 1e6);
+        weightDai = bound(weightDai, 1e16, 99e16);
+
+        // Remove will ask more BPT than what the sender has; we care about the revert reason, not the exact amount.
+        // TODO: use `expectPartialRevert` once forge is updated with `IVaultErrors.BptAmountInAboveMax.selector`
```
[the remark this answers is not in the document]
jubeira: it's been merged a few weeks ago, but there's no new tag yet. I don't want to use a nightly in the meantime just for this.

PR #971, opened 2024-09-12
Add LBPool

gerrrg:
# Description
tl;dr: Create a weight shifting LBPool for V3.

## What
* Port over the main functionality of LBPool from V2 to V3
* Port over the gradual value change library
* ~~Create a WeightValidation library for improved portability (could use this in standard WP if desired)~~
* Adhere to events for V2 LBP to improve the ease of integration/compatibility
* Make the pool contract its own hook contract and do before{join,swap} hooks for onlyOwner LP and swaps only when enabled
* Store info for only a single weight for higher precision/lower gas since weights must sum to `FP.ONE`
* Add relevant tests
* Modify some inherited tests (`BasePoolTest.sol`) to be more generalizable/override-able

## Why
Modernize a highly popular feature from v2

## Type of change

- [ ] Bug fix 
- [x] New feature 
- [ ] Breaking change 
- [ ] Dependency changes
- [ ] Code refactor / cleanup
- [ ] Other

## Checklist:

- [tbd] Tests have 100% code coverage

## Issue Resolution
n/a


EndymionJkb:
Might need to be updated again. Can you run `yarn prettier` to fix lint, so that CI will run?

gerrrg:
@EndymionJkb looks like it's linted just fine, but `yarn build` is failing due to `DocstringParsingError`s. I can either fix this, or if there's a pending fix already in the works I can grab that. Whatever's faster/easier

gerrrg:
@EndymionJkb it looks like it wasn't a lint issue but rather a mismatch between docstring return variable and what the var was actually called. I made them match but would appreciate a double check on that commit to make sure it aligns with what was intended

EndymionJkb:
> @EndymionJkb it looks like it wasn't a lint issue but rather a mismatch between docstring return variable and what the var was actually called. I made them match but would appreciate a double check on that commit to make sure it aligns with what was intended

I saw that, but it's not wrong in our main... so maybe it wasn't fully updated? If you pull current main into yours it should have fixed that. You shouldn't need to change it yourself.

gerrrg:
@EndymionJkb I'm not exactly sure what was going wrong, but I overwrote the branch version of WeightedMath with the one in `main`, so we're all good on that end now

gerrrg:
I've now incorporated all the changes that @EndymionJkb made in https://github.com/balancer/balancer-v3-monorepo/pull/974 and included them in this PR. I believe we can now close that one; I'll summarize the notes/comments from that review here for anything known outstanding.

TODOs from PR https://github.com/balancer/balancer-v3-monorepo/pull/974
* Consider checking for `swapsEnabled` in a minimally overridden `onSwap` rather than in an `onBeforeSwap` hook to reduce gas costs (using this hook results in a double storage read for balances)
* Consider changing the fn name for `updateWeightsGradually` since it describes the action as "update" rather than "schedule".
  * pro: it's more clear 👓 
  * con: it breaks previous version convention (do we care? 🤷)
* Consider renaming `NormalizedWeightInvariant` in WeightedPool (inherited by LBPool) to be more descriptive
* Port over v2 LBP tests

Other TODOs
* LBPool test coverage
* LBPFactory test(s)

Misc comments/down-the-line TODOs
* Seedless version
* `{,I}TrustedRouterProvider.sol`

PR #974, opened 2024-09-13
V3 LBP (v2)

EndymionJkb:
# Description

Based on #971 (thanks, gerrg!), this introduces a LiquidityBootstrappingPool, v3-style. (It was just too hard looking at it from a fork, where I couldn't switch to the branch.)

We were wondering about people using pools as hooks, hoping that would work: here's an example that does. I tried to make minimal changes, mostly in these areas:

1) Code style conventions that have evolved over the course of v3 (block headers, how NatSpec works, imports, etc.) Lint now passes. Also timestamps - everywhere else in the system they're uint32, so I kept that convention.

2) Simplification / reduction of duplication: there were constants, errors, etc. that could be inherited instead of repeated (sometimes with a tweak to the base contract). Removed separate functions that were only called once, made sure we were doing the 1 - w0 calculation in just one place, etc. Generally we try not to "pre-optimize."

3) Encapsulation: there were functions doing separate things and emitting unrelated events, controlled by flags. Simplified stuff like that.

4) Adding documentation (NatSpec for externals, plus more detail inside functions).

I made almost no changes to existing forge tests (just had to change one that was testing a function I removed), which indirectly validates that no functionality was changed. (I also didn't really review the tests; just made sure they still passed.) We should have all the tests we had in v2. One of the hardhat tests was calling non-existent router functions, so I did have to change that one.

One way to review would be to go commit by commit. The first one is the baseline code (just getting it to compile), then each commit describes a refactoring step (or two).

Questions: Should we spell out the full name as we did in v2? `LiquidityBootstrappingPool` vs. `LBPool`? To me it sounds confusable with the v1 `BPool`. (Relevant because of CowAMM.)

## Type of change

- [ ] Bug fix 
- [X] New feature 
- [ ] Breaking change 
- [ ] Dependency changes
- [ ] Code refactor / cleanup
- [ ] Other

## Checklist:


pkg/pool-weighted/contracts/LBPool.sol
```diff
@@ -0,0 +1,312 @@
...
+    // to ensure the PoolState fits in a single slot.
+    struct PoolState {
+        uint32 startTime;
+        uint32 endTime;
+        uint64 startWeight0;
+        uint64 endWeight0;
+        bool swapEnabled;
+    }
+    // `{start,end}Time` are `uint32`s. Ensure that no input time (passed as `uint256`) will overflow.
+    uint256 private constant _MAX_TIMESTAMP = type(uint32).max;
```
gerrrg, line 41: If we're using SafeCast now, I assume we can axe this line, yeah, @EndymionJkb?

pkg/pool-weighted/contracts/LBPool.sol
```diff
@@ -0,0 +1,312 @@
...
+    }
+
+    /*******************************************************************************
+                                  Internal Functions
+    *******************************************************************************/
+
+    // This is unused in this contract, but must be overridden from WeightedPool for consistency.
+    function _getNormalizedWeight(uint256 tokenIndex) internal view virtual override returns (uint256) {
+        if (tokenIndex < _NUM_TOKENS) {
+            return _getNormalizedWeights()[tokenIndex];
```
gerrrg, line 234: Is the conciseness of this preferable to only calculating one weight?

pkg/pool-weighted/contracts/LBPool.sol
```diff
@@ -0,0 +1,312 @@
...
+                revert MinWeight();
+            }
+            normalizedSum += normalizedWeight;
+        }
+
+        // Ensure that the normalized weights sum to ONE
+        if (normalizedSum != FixedPoint.ONE) {
+            revert NormalizedWeightInvariant();
+        }
+    }
```
gerrrg, line 311: Why get rid of the 2-token optimized function? Also if it's in a library like it was originally, it could be used from any weighted pool, reducing duplicated code among different Weighted(style)Pool variants. 

For example, vanilla WeightedPool, LBPool, (a hypothetical 2-token optimized v3) WeightedPool2Tokens could all use the same library.

For extra context, I had put it the (single-use) function in a library contract since I didn't want to overstep and change the vanilla WeightedPool contract to use it w/o discussing with y'all

pkg/pool-weighted/contracts/lib/GradualValueChange.sol
```diff
@@ -0,0 +1,92 @@
...
+     * change has not yet started, and FixedPoint.ONE means it has fully completed.
+     */
+    function calculateValueChangeProgress(uint256 startTime, uint256 endTime) internal view returns (uint256) {
+        if (block.timestamp >= endTime) {
+            return FixedPoint.ONE;
+        } else if (block.timestamp <= startTime) {
+            return 0;
+        }
+
+        // No need for SafeMath as it was checked right above: endTime > block.timestamp > startTime
```
gerrrg, line 85: Could probably ditch this vestigial comment from the v2 code since 0.8.* makes SafeMath irrelevant here

pkg/pool-weighted/test/foundry/LBPool.t.sol
```diff
@@ -0,0 +1,477 @@
...
+        vm.prank(bob);
+        vm.expectRevert(IVaultErrors.SwapFeePercentageTooLow.selector);
+        vault.setStaticSwapFeePercentage(pool, poolMinSwapFeePercentage - 1);
+    }
+
+    function testSetSwapFeeTooHigh() public override {
+        vm.prank(bob);
+        vm.expectRevert(abi.encodeWithSelector(IVaultErrors.SwapFeePercentageTooHigh.selector));
+        vault.setStaticSwapFeePercentage(pool, poolMaxSwapFeePercentage + 1);
+    }
```
gerrrg, line 371: I had originally defined these tests here as `override`s, but ultimately decided it was cleaner to have `BasePoolTest` handle them with a more intelligent way to handle the swap fee manager. Is it preferable to have this in this file? Feels like needlessly duplicating code to me

pkg/pool-weighted/contracts/LBPool.sol
gerrrg, line 41:
```suggestion
```

If we're using SafeCast now, it should be safe to axe this line and associated comment

gerrrg:
Misc style question: for the before swap/add liquidity hooks, I'm currently returning false if they action shouldn't proceed. This seems to be the preferred pattern, but it does ultimately result in a error that isn't the most descriptive to the user. Does it make more sense for these hook functions to revert with their own errors like `LpIsNotOwner` and `SwapsNotEnabled`?

pkg/pool-weighted/contracts/LBPool.sol
EndymionJkb: It's mainly to not have the "1 - weight" in multiple places, which is a relatively weak argument. It's not directly called here (but can be invoked from WeightedPool functions). This favors readability over gas, but I don't feel strongly.

pkg/pool-weighted/contracts/LBPool.sol
EndymionJkb: I just considered that pre-optimizing, which we tend to avoid (i.e., adding new libraries or functions that *might* be used in the future). The main weighted pool has immutable weights (as will probably most variants), so this would only be called once on deployment.

Ideally this could go in the base WeightedPool contract and just be called from here, but it can't because that pool is also storing immutables in the middle. We could do it anyway, and just do two loops in `WeightedPool`:

Instead of:
```
uint256 normalizedSum = 0;
for (uint8 i = 0; i < _totalTokens; ++i) {
    uint256 normalizedWeight = params.normalizedWeights[i];

    if (normalizedWeight < _MIN_WEIGHT) {
        revert MinWeight();
    }
    normalizedSum = normalizedSum + normalizedWeight;

    // prettier-ignore
    if (i == 0) { _normalizedWeight0 = normalizedWeight; }
    else if (i == 1) { _normalizedWeight1 = normalizedWeight; }
    else if (i == 2) { _normalizedWeight2 = normalizedWeight; }
    else if (i == 3) { _normalizedWeight3 = normalizedWeight; }
    else if (i == 4) { _normalizedWeight4 = normalizedWeight; }
    else if (i == 5) { _normalizedWeight5 = normalizedWeight; }
    else if (i == 6) { _normalizedWeight6 = normalizedWeight; }
    else if (i == 7) { _normalizedWeight7 = normalizedWeight; }
}

// Ensure that the normalized weights sum to ONE
if (normalizedSum != FixedPoint.ONE) {
    revert NormalizedWeightInvariant();
}
```

we'd have `_ensureValidWeights` defined in `WeightedPool`, and the code above would be:

```
_ensureValidWeights(params.normalizedWeights);

for (uint8 i = 0; i < _totalTokens; ++i) {
    uint256 normalizedWeight = params.normalizedWeights[i];

    // prettier-ignore
    if (i == 0) { _normalizedWeight0 = normalizedWeight; }
    else if (i == 1) { _normalizedWeight1 = normalizedWeight; }
    else if (i == 2) { _normalizedWeight2 = normalizedWeight; }
    else if (i == 3) { _normalizedWeight3 = normalizedWeight; }
    else if (i == 4) { _normalizedWeight4 = normalizedWeight; }
    else if (i == 5) { _normalizedWeight5 = normalizedWeight; }
    else if (i == 6) { _normalizedWeight6 = normalizedWeight; }
    else if (i == 7) { _normalizedWeight7 = normalizedWeight; }
}
```

pkg/pool-weighted/contracts/lib/GradualValueChange.sol
EndymionJkb: Good catch! It should be this, then (I'll change it.)

```
// No need for checked math as the magnitudes are verified above: endTime > block.timestamp > startTime
uint256 totalSeconds;
uint256 secondsElapsed;

unchecked {
    totalSeconds = endTime - startTime;
    secondsElapsed = block.timestamp - startTime;
}
```

pkg/pool-weighted/contracts/LBPool.sol
EndymionJkb: Yes; I guess we're not using this anymore. I'd originally kept it as documentation, but I can just expand the comment to show this. Good catch! (I think I did a search and thought it was being used - but it was the version in FactoryWidePauseWindow, which is private and can't be referenced from here, and I certainly don't want to change anything there.)

EndymionJkb:
> Misc style question: for the before swap/add liquidity hooks, I'm currently returning false if they action shouldn't proceed. This seems to be the preferred pattern, but it does ultimately result in a error that isn't the most descriptive to the user. Does it make more sense for these hook functions to revert with their own errors like `LpIsNotOwner` and `SwapsNotEnabled`?

I believe that is the standard, which is why we had it return a value. It should do one or the other; if we expect it to revert, it shouldn't return a value. If we return a value, it should be used vs. reverting.

It should be rare for a hook to revert under normal circumstances (e.g., it's an attack from a malicious router or something), and hooks should be relatively simple, so it's not that hard to check the code to see what happened.

gerrrg:
> so it's not that hard to check the code to see what happened
I agree that this is a totally logical next step for _developers_ interacting with these contracts, but my counter-argument is that any non-technical user (most people) will just see that the `onBefore<Action>Hook` error with no descriptive reason.

Only a tiny slice of folks would actually dig into the code to see why their txn failed; most probably don't know _how_ to look at it, and many of the remaining folks simply won't bother.

I'm happy to follow whatever the conventions are, just sharing my own perspective on users

pkg/pool-weighted/contracts/LBPool.sol
gerrrg: Ok fair enough to bring it back into the pool contract if it's not useful for WPs then.

Considering that an LBP (or other similar dynamic pools) could have multiple weight shifts over its lifetime, are you against an in-pool-contract version of the weight validator fn that uses the 2-token version? IMO it was pretty simple, clean code. Copying here w/o comments for easy reference:
```
function validateTwoWeights(uint256 normalizedWeight0, uint256 normalizedWeight1) internal pure {
    if (normalizedWeight0 < _MIN_WEIGHT || normalizedWeight1 < _MIN_WEIGHT) {
        revert MinWeight();
    }
    if (normalizedWeight0 + normalizedWeight1 != FixedPoint.ONE) {
        revert NormalizedWeightInvariant();
    }
}
```

pkg/pool-weighted/contracts/LBPool.sol
gerrrg: Sure. I was trying to gas optimize here but don't have a strong preference either. We can leave it as is

pkg/pool-weighted/contracts/LBPool.sol
EndymionJkb: Given that this is only called in one place (because the base contract WeightedPool handles it on construction), I can just move this code inline. That's even simpler, and optimized.

EndymionJkb:
> I agree that this is a totally logical next step for _developers_ interacting with these contracts, but my counter-argument is that any non-technical user (most people) will just see that the `onBefore<Action>Hook` error with no descriptive reason.
> 
> Only a tiny slice of folks would actually dig into the code to see why their txn failed; most probably don't know _how_ to look at it, and many of the remaining folks simply won't bother.

Well, it's not a hard and fast rule. Generally, if there's only one way a hook can fail, returning false is sufficient and the easiest way to handle it. (Another reason for having a return value is to catch pathological (or malicious) cases where hooks say they implement something but don't, and it goes to the fallback function, etc.)

If you have a hook that can fail in several ways, it makes sense to have custom reverts. In your example here, that is the case. It returns false if it fails for the "normal" reason (not being called by the owner): but reverts if you use a non-standard router (error/attack case), which is fine.

pkg/pool-weighted/contracts/LBPool.sol
```diff
@@ -0,0 +1,294 @@
...
+        endWeights = new uint256[](_NUM_TOKENS);
+        endWeights[0] = poolState.endWeight0;
+        endWeights[1] = FixedPoint.ONE - poolState.endWeight0;
+    }
+
+    /**
+     * @notice Indicate whether swaps are enabled or not for the given pool.
+     * @return swapEnabled True if trading is enabled
+     */
+    function getSwapEnabled() external view returns (bool) {
```
danielmkm, line 125: It's probably not worth it, but this is a type of "pause" and we could consider adding a "swapsPaused" state to the vault. The goal of the pause manager was to have a consistent pattern for all pools when pausing, so we have less instances where internally the pool disables itself and we have to react to that state change. It increases operational burden for all of the off-chain tooling.

pkg/pool-weighted/contracts/LBPool.sol
```diff
@@ -0,0 +1,294 @@
...
+    /**
+     * @notice Start a gradual weight change. Weights will change smoothly from current values to `endWeights`.
+     * @dev This is a permissioned function that can only be called by the owner.
+     * If the `startTime` is in the past, the weight change will begin immediately.
+     *
+     * @param startTime The timestamp when the weight change will start
+     * @param endTime  The timestamp when the weights will reach their final values
+     * @param endWeights The final values of the weights
+     */
+    function updateWeightsGradually(
```
danielmkm, line 151: Super minor, but since this has a `startTime`, you'd think it should be named something like `scheduleGradualWeightsUpdate`

pkg/pool-weighted/contracts/LBPool.sol
```diff
@@ -0,0 +1,294 @@
...
+    }
+
+    function _setSwapEnabled(bool swapEnabled) private {
+        _poolState.swapEnabled = swapEnabled;
+
+        emit SwapEnabledSet(swapEnabled);
+    }
+
+    /**
+     * @dev When calling updateWeightsGradually again during an update, reset the start weights to the current weights,
```
danielmkm, line 272: The comment here is true, but it doesn't describe what the function is doing. Anything can be passed to this function as `startWeights`, it is not necessarily the current weights. This comment is more appropriate at line 169.

pkg/pool-weighted/contracts/LBPool.sol
```diff
@@ -0,0 +1,294 @@
...
+        uint256 endTime,
+        uint256[] memory endWeights
+    ) external onlyOwner {
+        InputHelpers.ensureInputLengthMatch(_NUM_TOKENS, endWeights.length);
+
+        if (endWeights[0] < _MIN_WEIGHT || endWeights[1] < _MIN_WEIGHT) {
+            revert MinWeight();
+        }
+        if (endWeights[0] + endWeights[1] != FixedPoint.ONE) {
+            revert NormalizedWeightInvariant();
```
danielmkm, line 162: This is a more global point, but it'd be great if we had errors that are more descriptive in their naming. This is very clear in context, but much less clear as an integrator.

danielmkm (commented):
Thanks for all of the work @gerrrg, overall looking good!
- Can we consider moving this into it's own package.
- Copying the code over has erased all of greg's commit history, it'd be great if we can find a way to preserve that so that his contributions are recognized in the repo.

pkg/pool-weighted/contracts/LBPool.sol
```diff
@@ -0,0 +1,294 @@
...
+
+        revert RouterNotTrusted();
+    }
+
+    /**
+     * @notice Called before a swap to let the pool block swaps if not enabled.
+     * @return success True if the pool has swaps enabled.
+     */
+    function onBeforeSwap(PoolSwapParams calldata, address) public view override onlyVault returns (bool) {
+        return _getPoolSwapEnabled();
```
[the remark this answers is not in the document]
EndymionJkb: More than just the overhead of calling the hook itself, this is mainly because it needs to reload balances and rates after calling a before hook, since they might have changed: even though all we're doing here is checking a flag. We'd briefly considered having before hooks return whether they need to be reloaded, but that's asking for trouble (a malicious one could say they didn't need a reload when they did).

pkg/pool-weighted/contracts/LBPool.sol
EndymionJkb: I'd hesitate to add something to the Vault to accommodate a specialized kind of pool (and all the checks through the Vault code; unbalanced add is a swap, etc.)

We could just use the regular pause, if we don't think the owner needs to add liquidity while swaps are paused. Of course, it would complicate the initialization, as they wouldn't be able to fund the pool if it starts paused. Or they'd have to unpause/initialize/pause in one transaction.

pkg/pool-weighted/contracts/LBPool.sol
EndymionJkb: This is using the legacy terminology, all the way back to v1. If it's not important to maintain that, it could be renamed.

pkg/pool-weighted/contracts/LBPool.sol
EndymionJkb: Legacy v2 again :) (v1 had denormalized weights.) We could review all the error messages / ask partners/auditors if any are confusing.

pkg/pool-weighted/contracts/LBPool.sol
gerrrg: Looks like @EndymionJkb beat me to it. I personally don't mind if we want to modernize the function name; I was just trying to maintain the [v2 function names](https://github.com/balancer/balancer-v2-monorepo/blob/master/pkg/pool-weighted/contracts/lbp/LiquidityBootstrappingPoolSettings.sol#L206-L219) for easiest interfacing with existing integrators

pkg/pool-weighted/contracts/LBPool.sol
```diff
@@ -0,0 +1,294 @@
...
+        address owner,
+        bool swapEnabledOnStart,
+        address trustedRouter
+    ) WeightedPool(params, vault) Ownable(owner) {
+        // WeightedPool validates `numTokens == normalizedWeights.length`, and ensures valid weights.
+        // Here we additionally enforce that LBPs must be two-token pools.
+        InputHelpers.ensureInputLengthMatch(_NUM_TOKENS, params.numTokens);
+
+        // Set the trusted router (passed down from the factory).
+        _TRUSTED_ROUTER = trustedRouter;
```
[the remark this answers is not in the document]
gerrrg: 💯 

I had initially put in placeholders for `ITrustedRouterProvider` in my PR, but removed them to reduce clutter/improve clarity. I'd lean towards keeping the LBPool using `_TRUSTED_ROUTER` until the point that `ITrustedRouterProvider` exists. I briefly considered writing the `TrustedRouterProvider` myself but deemed it out of scope for this PR/likely overstepping

EndymionJkb:
> Thanks for all of the work @gerrrg, overall looking good!
> 
> * Can we consider moving this into it's own package.
I don't know about a new package; in v2 we had an /lbp directory under weighted. We could do that again. It is a weighted pool type (directly derived from it).

> * Copying the code over has erased all of greg's commit history, it'd be great if we can find a way to preserve that so that his contributions are recognized in the repo.

As noted above, it's difficult to review a fork that I can't pull and look at. We could copy it back to the fork at this point, and work from that, now that it's a lot closer.

pkg/pool-weighted/contracts/LBPool.sol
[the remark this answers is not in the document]
gerrrg: I don't have a strong preference either way on this one. The hook seemed most elegant, but overriding `onSwap` seems easy enough. Realistically I'm going to need to override `onSwap` for the seedless version anyway, so that's fine with me

pkg/pool-weighted/contracts/LBPool.sol
gerrrg: I think this comment to less intended to describe the function's behavior and more to direct any developer how to use it. FWIW, this was a [direct copy/paste from v2](https://github.com/balancer/balancer-v2-monorepo/blob/master/pkg/pool-weighted/contracts/lbp/LiquidityBootstrappingPoolSettings.sol#L280-L281)

pkg/pool-weighted/contracts/LBPool.sol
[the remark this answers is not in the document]
EndymionJkb: It would add significant gas to every swap - thousands at least - with all the reloading. It's the side effects that make it expensive: and aren't apparent until you look at the Vault code. If you're going to do it anyway, I think it makes sense to introduce that now.

pkg/pool-weighted/contracts/LBPool.sol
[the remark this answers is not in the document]
EndymionJkb: Yes; I'd considered it pre-optimizing, which is why I removed it in this version. It's not that unreasonable for us to just migrate the LBP factory when we migrate the standard router, and it's not affected by adding a new (unrelated) router.

Could be a follow-on PR, but I agree we don't need it in this initial one.

gerrrg:
> it's difficult to review a fork that I can't pull and look at

Is the possible to get permissions on this repo to create a branch rather than from my own forked repo then? 

> I don't know about a new package; in v2 we had an /lbp directory under weighted. We could do that again. It is a weighted pool type (directly derived from it).

I'm inclined to agree with @EndymionJkb here, but obv I don't have the final say. Just my two cents

gerrrg:
@EndymionJkb I was able to take my fork's branch and apply all of the changes during the review with the asterisk that now it wipes out all of _your_ git history. That branch is [here](https://github.com/gerrrg/balancer-v3-monorepo/tree/add-lbpool-post-prelim-review).

What we could do to properly credit each thing (assuming I'm able to get branch permissions here) is:
* I copy my branch over to this repo
* I send you the patch file from the `diff` between `gerrrg/balancer-v3-monorepo:add-lbpool` and `balancer/balancer-v3-monorepo:v3-lbp` 
* You apply it to my branch

EndymionJkb:
> @EndymionJkb I was able to take my fork's branch and apply all of the changes during the review with the asterisk that now it wipes out all of _your_ git history. That branch is [here](https://github.com/gerrrg/balancer-v3-monorepo/tree/add-lbpool-post-prelim-review).
> ...
> Then from there, I could move the LBP stuff to its own subdirectory of the wp package if that's the direction we end up deciding on

I'm not worried about my commits here. The goal here was to ensure you're in the commit log for v3 (otherwise it just looks like I wrote the LBP). I already have elebenty thousand v3 commits :)

gerrrg:
In that case, I've got a branch (in my forked repo) ready to go with the current state of things w/ my git history there 🙂

pkg/pool-weighted/contracts/LBPool.sol
gerrrg: I would lean towards keeping it as it. Using the existing notion of paused pools feels like trying to shoe-horn in something that's similar but not exactly what we need. The unpause/initialize/pause seems like bending over backwards just to use that tool. I do also agree with @EndymionJkb that adding different `<Something>Managers` could spiral out of control. I think it's evident enough how it should work w/ the pool `owner`, but it someone strongly disagrees, I'm happy to discuss.

pkg/pool-weighted/contracts/LBPool.sol
gerrrg: resolving and adding as a comment to https://github.com/balancer/balancer-v3-monorepo/pull/971

pkg/pool-weighted/contracts/LBPool.sol
gerrrg: marking resolved and adding as a comment to https://github.com/balancer/balancer-v3-monorepo/pull/971

EndymionJkb:
Revert back to #971, after changes from here applied.

PR #975, opened 2024-09-13
Support negative imbalance in ERC4626 buffers

joaobrunoah:
# Description

In the current buffer implementation, the buffer only rebalances when the buffer doesn't have enough liquidity to fulfill the trade and:
* Wrap operation, buffer with more underlying than wrapped (underlying positive surplus)
* Unwrap operation, buffer with more wrapped than underlying (wrapped positive surplus)

The current PR makes the buffer rebalance also in the following situations
* Wrap operation, buffer with more wrapped than underlying (underlying negative surplus)
* Unwrap operation, buffer with more underlying than wrapped (wrapped negative surplus)

We call it negative surplus because it's a surplus in the opposite direction of the main operation. For example, a Wrap operation usually would wrap the underlying tokens from the user and the surplus of underlying in the same deposit/mint call. But, if the buffer has a surplus of wrapped tokens, we need to deposit/mint `user tokens - wrappedSurplusConvertedToUnderlying`, since we will use part of the wrapped balance of the buffer to give token to the user, and leave user underlying tokens in the buffer.

**_IMPORTANT: Notice that now we rely on the result of the 'preview' operation to tell how many tokens to return to the sender. Pay attention when reviewing this to make sure there isn't an attack vector to drain the buffer. Also, buffer and vault reserves should be fully isolated, we need to check if this isolation wasn't broken by allowing a negative delta._**

## Type of change

- [x] Bug fix 
- [ ] New feature 
- [ ] Breaking change 
- [ ] Dependency changes
- [ ] Code refactor / cleanup
- [ ] Other

## Checklist:


pkg/vault/contracts/Vault.sol
```diff
@@ -1201,27 +1203,25 @@ contract Vault is IVaultMain, VaultCommon, Proxy {
                 // The amount of underlying tokens to deposit is the necessary amount to fulfill the trade
                 // (amountInUnderlying), plus the amount needed to leave the buffer rebalanced 50/50 at the end
                 // (bufferUnderlyingSurplus).
-                vaultUnderlyingDeltaHint = amountInUnderlying + bufferUnderlyingSurplus;
+                vaultUnderlyingDeltaHint = (amountInUnderlying.toInt256() + bufferUnderlyingSurplus).toUint256();
```
elshan-eth, line 1206: I don't quite understand why there is `int256` + `uint256`

joaobrunoah: Hey Elshan! Good point, I'll comment it better. bufferUnderlyingSurplus is an int256 that can be a positive or a negative number. So, we need to convert amountINUnderlying to int256, sum both values, then convert again to uint256 because vaultUnderlyingDeltaHint needs to be a positive number (and so does amountInUnderlying)

elshan-eth (approved):
LGTM!

pkg/vault/contracts/Vault.sol
```diff
@@ -1323,9 +1331,9 @@ contract Vault is IVaultMain, VaultCommon, Proxy {
             // We unwrap the user's tokens via an external call and additionally rebalance the buffer if it has a
             // surplus of wrapped tokens.
 
-            // Gets the amount of wrapped tokens to unwrap in order to rebalance the buffer.
-            uint256 bufferWrappedSurplus = bufferBalances.getBufferWrappedSurplus(wrappedToken);
-            uint256 bufferUnderlyingSurplus;
+            // Gets the surplus of wrapped in the buffer. If negative, the buffer has more underlying than wrapped.
+            int256 bufferWrappedToRebalance = bufferBalances.getBufferWrappedSurplus(wrappedToken);
+            int256 bufferUnderlyingToRebalance;
```
jubeira, line 1336: Same as above; how about `bufferWrappedRebalanceDelta`?

pkg/vault/contracts/Vault.sol
```diff
@@ -1336,23 +1344,24 @@ contract Vault is IVaultMain, VaultCommon, Proxy {
...
-                    // underlying tokens the buffer should receive to rebalance, based on the wrapped token surplus.
-                    // The redeem function is used here, as it performs the inverse of a withdraw operation.
-                    bufferUnderlyingSurplus = wrappedToken.previewRedeem(bufferWrappedSurplus);
+                // The amount of underlying tokens to withdraw is calculated in terms of the wra.
+                if (bufferWrappedToRebalance != 0) {
+                    vaultWrappedDeltaHint = (amountInWrapped.toInt256() + bufferWrappedToRebalance).toUint256();
+                    vaultUnderlyingDeltaHint = wrappedToken.previewRedeem(vaultWrappedDeltaHint);
+                } else {
+                    // If the buffer does not have a surplus, unwrap the exact `amountOutUnderlying` amount.
+                    vaultUnderlyingDeltaHint = amountOutUnderlying;
```
jubeira, line 1362: No assignment for `vaultWrappedDeltaHint` in this case?

I get that it's assigned below unconditionally. But maybe it should be symmetrical with `_wrap`.

pkg/vault/contracts/Vault.sol
```diff
@@ -1200,28 +1202,43 @@ contract Vault is IVaultMain, VaultCommon, Proxy {
...
+                    // `abs(bufferUnderlyingToRebalance) > amountInUnderlying` and `bufferUnderlyingToRebalance < 0`,
+                    // it means the buffer has enough liquidity to fulfill the trade (i.e.
+                    // `bufferBalances.getBalanceDerived() >= amountOutWrapped` is true).
+                    vaultUnderlyingDeltaHint = (amountInUnderlying.toInt256() + bufferUnderlyingToRebalance)
+                        .toUint256();
+                    vaultWrappedDeltaHint = wrappedToken.previewDeposit(vaultUnderlyingDeltaHint);
+                } else {
+                    // If the buffer does not have a surplus, wrap the exact `amountOutWrapped` amount.
+                    vaultUnderlyingDeltaHint = amountInUnderlying;
+                    vaultWrappedDeltaHint = amountOutWrapped;
```
jubeira, line 1233: This one is assigned below; can it be skipped? (not so much for the assignment itself, but this should be symmetrical wrt. `_unwrap`.

pkg/vault/test/.contract-sizes/Vault
```diff
@@ -1,2 +1,2 @@
-Bytecode	23.712
-InitCode	25.053
\ No newline at end of file
+Bytecode	23.486
```
jubeira, line 1: :fire:

pkg/vault/test/foundry/BufferVaultPrimitive.t.sol
jubeira, line 1: One thing here: all the `deposit` tests (with buffer liquidity, without buffer liquidity and underlying surplus, without buffer liquidity and underlying deficit) should give out the exact same results right in terms of inputs and outputs right? After all, getting a quote from the buffer should be the same as getting a quote from the wrapper, and that should also not depend on the buffer conditions.
Plus, for not-enough-liquidity the buffer should also end up in the same state (balanced).

Can we also add that check?

jubeira: Well, I see that `_wrapAmount` is fixed across tests and is always the same. So perhaps this is already covered?

jubeira (commented):
This is looking pretty good. Code is simpler, less gas and bytecode, and more accurate queries and operations. It's just a massive win on every front.

First pass done. I have some comments / suggestions / questions, mostly minor details; PTAL. About naming: I think it'd be good to use `delta` as opposed to `surplus` now that it can go negative, and try to keep variable names consistent with the getters.

pkg/solidity-utils/contracts/helpers/BufferHelpers.sol
```diff
@@ -20,29 +23,25 @@ library BufferHelpers {
      * - 3 underlying = 1.5 wrapped
      * - final balances: 3.5 wrapped (2 existing + 1.5 new) and 7 underlying (10 existing - 3)
      */
-    function getBufferUnderlyingSurplus(bytes32 bufferBalance, IERC4626 wrappedToken) internal view returns (uint256) {
-        uint256 underlyingBalance = bufferBalance.getBalanceRaw();
+    function getBufferUnderlyingSurplus(bytes32 bufferBalance, IERC4626 wrappedToken) internal view returns (int256) {
```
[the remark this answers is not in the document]
EndymionJkb: Yes; it would be surplus / deficit. Or really, surplus / surplus in the opposite type. A negative "underlying surplus" is a wrapped surplus. Delta is a good generic term to capture the difference.

pkg/vault/contracts/Vault.sol
```diff
@@ -1336,23 +1344,24 @@ contract Vault is IVaultMain, VaultCommon, Proxy {
...
-                // (bufferUnderlyingSurplus).
-                if (bufferWrappedSurplus > 0) {
-                    // An unwrap operation withdraws wrapped tokens from the buffer and returns underlying tokens.
-                    // For the buffer to rebalance during an unwrap operation, it must have an excess (surplus) of
-                    // wrapped tokens. Since this is an EXACT_OUT unwrap (withdraw), we need to calculate the amount of
-                    // underlying tokens the buffer should receive to rebalance, based on the wrapped token surplus.
-                    // The redeem function is used here, as it performs the inverse of a withdraw operation.
-                    bufferUnderlyingSurplus = wrappedToken.previewRedeem(bufferWrappedSurplus);
+                // The amount of underlying tokens to withdraw is calculated in terms of the wra.
+                if (bufferWrappedToRebalance != 0) {
```
EndymionJkb, line 1357: For symmetry, should have a comment inside the if corresponding to `The amount of underlying tokens to deposit is the necessary amount to fulfill the trade...` in `_wrapWithBuffer`.

pkg/vault/contracts/Vault.sol
```diff
@@ -1200,28 +1202,43 @@ contract Vault is IVaultMain, VaultCommon, Proxy {
...
-                    // For the buffer to rebalance during a wrap operation, it must have an excess (surplus) of
-                    // underlying tokens. Since this is an EXACT_OUT wrap (mint), we need to calculate the amount of
-                    // wrapped tokens the buffer should receive to rebalance, based on the underlying token surplus.
-                    // The deposit function is used here, as it performs the inverse of a mint operation.
-                    bufferWrappedSurplus = wrappedToken.previewDeposit(bufferUnderlyingSurplus);
+                // We need to calculate both `vaultUnderlyingDeltaHint` and `vaultWrappedDeltaHint`. The underlying
+                // delta will be used to approve transfer from underlying token to the wrapper protocol, while the
+                // wrapped delta will be used to wrap an EXACT_OUT amount of wrapped tokens (mint).
+                if (bufferUnderlyingToRebalance != 0) {
+                    // The amount of underlying tokens to deposit is the necessary amount to fulfill the trade
```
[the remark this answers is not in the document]
EndymionJkb: Re: symmetry, here's what I think would be the most simple/symmetric (removing comments for illustration, and not doing any renaming). Sometimes variables get overwritten multiple times along the path, but that's partly because of type conversion, and I think it's ok.

In `_wrapWithBuffer`:
```
int256 bufferUnderlyingToRebalance = bufferBalances.getBufferUnderlyingSurplus(wrappedToken);
int256 bufferWrappedToRebalance;

uint256 vaultUnderlyingDeltaHint = amountInUnderlying;
uint256 vaultWrappedDeltaHint = amountOutWrapped;

if (kind == SwapKind.EXACT_IN) {
    vaultUnderlyingDeltaHint = (amountInUnderlying.toInt256() + bufferUnderlyingToRebalance).toUint256();
    underlyingToken.forceApprove(address(wrappedToken), vaultUnderlyingDeltaHint);
    vaultWrappedDeltaHint = wrappedToken.deposit(vaultUnderlyingDeltaHint, address(this));
} else {
    if (bufferUnderlyingToRebalance != 0) {
        vaultUnderlyingDeltaHint = (amountInUnderlying.toInt256() + bufferUnderlyingToRebalance)
            .toUint256();
        vaultWrappedDeltaHint = wrappedToken.previewDeposit(vaultUnderlyingDeltaHint);
    }

    underlyingToken.forceApprove(address(wrappedToken), vaultUnderlyingDeltaHint);
    vaultUnderlyingDeltaHint = wrappedToken.mint(vaultWrappedDeltaHint, address(this));
}

underlyingToken.forceApprove(address(wrappedToken), 0);
_settleWrap(underlyingToken, IERC20(wrappedToken), vaultUnderlyingDeltaHint, vaultWrappedDeltaHint);
```

In `_unwrapWithBuffer`:
```
int256 bufferWrappedToRebalance = bufferBalances.getBufferWrappedSurplus(wrappedToken);
int256 bufferUnderlyingToRebalance;

uint256 vaultUnderlyingDeltaHint = amountOutUnderlying;
uint256 vaultWrappedDeltaHint = amountInWrapped;

if (kind == SwapKind.EXACT_IN) {
    vaultWrappedDeltaHint = (amountInWrapped.toInt256() + bufferWrappedToRebalance).toUint256();
    vaultUnderlyingDeltaHint = wrappedToken.redeem(vaultWrappedDeltaHint, address(this), address(this));
} else {
    if (bufferWrappedToRebalance != 0) {
        vaultWrappedDeltaHint = (amountInWrapped.toInt256() + bufferWrappedToRebalance).toUint256();
        vaultUnderlyingDeltaHint = wrappedToken.previewRedeem(vaultWrappedDeltaHint);
    }

    vaultWrappedDeltaHint = wrappedToken.withdraw(vaultUnderlyingDeltaHint, address(this), address(this));
}

_settleUnwrap(underlyingToken, IERC20(wrappedToken), vaultUnderlyingDeltaHint, vaultWrappedDeltaHint);
```

EndymionJkb (commented):
First pass for symmetry/renaming. Shaping up well!

pkg/vault/test/foundry/BufferVaultPrimitive.t.sol
joaobrunoah: We are already checking if buffer is rebalanced at the end (line 1058). Also, line 1051 checks if the deposit operation was the same for all deposit cases, since `previewDeposit(_wrapAmount)` has the same value across all tests.

pkg/vault/contracts/Vault.sol
joaobrunoah: So, I need to approve the transfer of underlying (line 1238) and mint ExactOut using wrapped amount (line 1241). The wrap is assymetrical in relation to _unwrap, because in the unwrap operation I don't need approvals to burn wrapped tokens (wToken contract has allowance to burn its own tokens, but does not have allowance to move underlying from sender to wToken contract).

pkg/vault/contracts/Vault.sol
joaobrunoah: unwrap and wrap are assymetrical, due to the approval. If we assigned vaultWrappedDeltaHint here, it wouldn't be used anywhere (differently from the wrap function, where we need both values).

EndymionJkb: See related comment below (https://github.com/balancer/balancer-v3-monorepo/pull/975#discussion_r1764175974). It is asymmetric, but assigning the hints at the top (while in one case it wouldn't be used), might be a bit cleaner, and avoids the "else" clauses.

joaobrunoah: I made it a bit more symmetrical, following @EndymionJkb suggestion.

pkg/vault/contracts/Vault.sol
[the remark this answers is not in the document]
joaobrunoah: Done!

pkg/vault/contracts/Vault.sol
joaobrunoah: @jubeira @EndymionJkb , I needed to mess the symmetry of wrap and unwrap, but gained 80 bytes and, in the real world, potentially 10k of gas by removing a previewRedeem from unwrap ExactOut.

pkg/vault/contracts/Vault.sol
```diff
@@ -1234,39 +1247,30 @@ contract Vault is IVaultMain, VaultCommon, Proxy {
...
+                // that the buffer had more wrapped than underlying. If that's the case, an amount of underlying must
+                // be added to the buffer, so `bufferUnderlyingImbalance` must be negative (note that the buffer
+                // balance update below already removes bufferUnderlyingImbalance from the buffer, so if this number
+                // is negative it'll actually be added to the buffer balances).
+                bufferUnderlyingImbalance = vaultUnderlyingDeltaHint.toInt256() - amountInUnderlying.toInt256();
+                // A wrap should remove underlying from the sender and buffer, and return wrapped tokens to the sender
+                // and buffer. However, `vaultWrappedDeltaHint` may be smaller than `amountOutWrapped`, meaning
+                // that the buffer had more wrapped than underlying. If that's the case, an amount of wrapped must
+                // be removed from the buffer, so `bufferWrappedImbalance` must be negative.
+                bufferWrappedImbalance = vaultWrappedDeltaHint.toInt256() - amountOutWrapped.toInt256();
```
jubeira, line 1263: As stated above, this is the only real place where we're computing the wrapped imbalance (the one before `_settleWrap` can be computed without calling `getBufferWrappedImbalance`.

Then, this can be a local inside this `if`.

jubeira (commented):
Another row of comments :)

I like the iterative approach, but now that we're polishing it and changing how we deal with things I think we can polish it a bit more even.
I've just sent comments for `wrap`, but same applies for `unwrap` (should end up being symmetrical after all).

pkg/vault/contracts/Vault.sol
```diff
@@ -1185,43 +1187,54 @@ contract Vault is IVaultMain, VaultCommon, Proxy {
...
-            // surplus of underlying tokens.
+            // We wrap the user's tokens via an external call and additionally rebalance the buffer if it has an
+            // imbalance of underlying tokens.
 
-            // Gets the amount of underlying to wrap in order to rebalance the buffer.
-            uint256 bufferUnderlyingSurplus = bufferBalances.getBufferUnderlyingSurplus(wrappedToken);
-            uint256 bufferWrappedSurplus;
+            // Gets the imbalance of underlying in the buffer. If negative, the buffer has more wrapped than underlying.
+            int256 bufferUnderlyingImbalance = bufferBalances.getBufferUnderlyingImbalance(wrappedToken);
+            int256 bufferWrappedImbalance = bufferBalances.getBufferWrappedImbalance(wrappedToken);
```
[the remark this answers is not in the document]
joaobrunoah: So, this is a long thread of thought:
* When we are wrapping ExactOut, we need to make sure that we have enough amountOutWrapped to give to the user at the end. 
* In the previous approach, I was using `vaultWrappedDeltaHint = previewDeposit(amountInUnderlying + bufferUnderlyingImbalance)`, which works but is not 100% accurate. I can't ensure that I'll have enough amountOutWrapped and that the buffer will be balanced, because I'm not using the ExactOut amounts to calculate the mint argument. 
* Calculating both imbalances does not increase gas costs and reduces the bytecode (especially in the unwrap part, where I don't need to approve anything and saved one external call).

pkg/vault/contracts/Vault.sol
```diff
@@ -1185,43 +1187,54 @@ contract Vault is IVaultMain, VaultCommon, Proxy {
...
-                    // A wrap operation withdraws underlying tokens from the buffer and mints wrapped tokens in return.
-                    // For the buffer to rebalance during a wrap operation, it must have an excess (surplus) of
-                    // underlying tokens. Since this is an EXACT_OUT wrap (mint), we need to calculate the amount of
-                    // wrapped tokens the buffer should receive to rebalance, based on the underlying token surplus.
-                    // The deposit function is used here, as it performs the inverse of a mint operation.
-                    bufferWrappedSurplus = wrappedToken.previewDeposit(bufferUnderlyingSurplus);
+                // We need to calculate both `vaultUnderlyingDeltaHint` and `vaultWrappedDeltaHint`. The underlying
+                // delta will be used to approve transfer from underlying token to the wrapper protocol, while the
+                // wrapped delta will be used to wrap an EXACT_OUT amount of wrapped tokens (mint).
+                if (vaultWrappedDeltaHint != 0) {
```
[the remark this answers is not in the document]
joaobrunoah: This was a typo (and a very ugly one). I should be checking for the imbalance, not for the delta hint. The tests because it doesn't make a difference in terms of logic, but makes in terms of gas.

pkg/vault/contracts/Vault.sol
```diff
@@ -1185,43 +1187,54 @@ contract Vault is IVaultMain, VaultCommon, Proxy {
...
+            int256 bufferWrappedImbalance = bufferBalances.getBufferWrappedImbalance(wrappedToken);
 
-            // Expected amount of underlying deposited into the wrapper protocol.
-            uint256 vaultUnderlyingDeltaHint;
-            // Expected amount of wrapped minted by the wrapper protocol.
-            uint256 vaultWrappedDeltaHint;
+            // Expected amount of underlying deposited into the wrapper protocol if buffer is balanced.
+            uint256 vaultUnderlyingDeltaHint = amountInUnderlying;
+            // Expected amount of wrapped minted by the wrapper protocol if buffer is balanced.
+            uint256 vaultWrappedDeltaHint = amountOutWrapped;
```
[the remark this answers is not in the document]
joaobrunoah: Actually, it doesn't. For ExactOut, the if had a typo, it should be checking the imbalance and not the delta hint. So, for ExactOut, if there's no imbalance in the buffer, these values will be used to wrap.

pkg/vault/contracts/Vault.sol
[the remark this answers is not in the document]
joaobrunoah: After some more thought, I reverted and did as you said. Calculating the buffer imbalance already calls an external call, so it won't make a difference, and the 80 bytes of saving was not due to this change, but related to the typo in the `if` clause of the _wrapWithBuffer function that I described below.

pkg/vault/contracts/Vault.sol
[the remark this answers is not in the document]
jubeira: hm, got it.

I have mixed feelings about that last `if (bufferUnderlyingImbalance != 0)`. It does optimize things for buffers that have very little liquidity. Perhaps in some L2s buffers will be initialized just for the system to work, but without any meaningful liquidity. In that case, there's no need to rebalance it ever, as any trade will always just go to the lending protocol, and the buffer will always stay balanced. More concretely, it saves one `preview` function which can be interesting (the sum doesn't move the needle). I understand this case is tested in `testMintBufferBalancedNotEnoughLiquidity` under `BufferVaultPrimitive`.

On the other hand, it's pretty deep already (if not enough liquidity, if exact out, if rebalance is needed), and it introduces an asymmetry with exact in (which to be fair happens rather often with exact outs). And also the fact that we need to initialize `vaultUnderlyingDeltaHint` and `vaultWrappedDeltaHint` just in case we don't enter the most inner `if`, but at the same time we overwrite those two almost every time (i.e. unless it's perfectly balanced) is a bit tricky.

Without the last `if`, we have:
- if no liquidity --> creates delta hints, no assignment needed
  - exact in --> always sets delta hints 
  - exact out --> always sets delta hints 

and the worst part is that extra `preview` call. The result should be the same if we do the math right? i.e. for `bufferUnderlyingImbalance == 0` we get `vaultUnderlyingDeltaHint == amountInUnderlying` (same as now) and `vaultWrappedDeltaHint == previewDeposit(amountInUnderlying)` --> is this the same as `amountOutWrapped` or does that have a rounding error?

I'd love to hear somebody else's thoughts about this. This smells like an optimization that bites you back; I'm not sure it's really worth it.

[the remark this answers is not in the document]
jubeira: > `previewMint(previewDeposit(x)) == x`

This is not the case. That's actually a small source of rounding errors (which actually will happen almost always, since whenever we go to the lending protocol for liquidity chances are that the buffer is not going to be balanced).

pkg/vault/contracts/Vault.sol
```diff
@@ -1234,40 +1247,29 @@ contract Vault is IVaultMain, VaultCommon, Proxy {
...
+            // be added to the buffer, so `bufferUnderlyingImbalance` must be negative (note that the buffer
+            // balance update below already removes bufferUnderlyingImbalance from the buffer, so if this number
+            // is negative it'll actually be added to the buffer balances).
+            int256 bufferUnderlyingImbalance = vaultUnderlyingDeltaHint.toInt256() - amountInUnderlying.toInt256();
+            // A wrap should remove underlying from the sender and buffer, and return wrapped tokens to the sender
+            // and buffer. However, `vaultWrappedDeltaHint` may be smaller than `amountOutWrapped`, meaning
+            // that the buffer had more wrapped than underlying. If that's the case, an amount of wrapped must
+            // be removed from the buffer, so `bufferWrappedImbalance` must be negative.
+
+            int256 bufferWrappedImbalance = vaultWrappedDeltaHint.toInt256() - amountOutWrapped.toInt256();
```
jubeira, line 1262: last minute megasilly comment, but it's something that helps me follow the code better: the comments in the code are good. But more than "imbalance" this is some sort of `netBufferUnderlyingDelta` / `netBufferWrappedDelta`, measured according to the real balances entering and exiting the vault. Or maybe `bufferOutgoingUnderlyingAmount`, `bufferIncomingWrappedAmount` if you want to preserve the signs when we adjust the balance.

Bottom line, this is not the buffer imbalance that we are calculating; we already do that before wrapping. We're calculating how much to take out from buffer underlying balances (i.e. whatever the vault used to wrap `vaultUnderlyingDeltaHint` minus whatever the user paid `amountInUnderlying`) and how much to add to buffer wrapped balances (i.e. whatever the vault received from the lending protocol `vaultWrappedDeltaHint` minus whatever the vault pays the user `amountOutWrapped`).

The buffer ends up rebalanced because of the math we did before wrapping; now we're just accounting for those results.

jubeira (commented):
Great job! Hopefully this is the last round of questions and feedback.

I like the new structure much better than the previous one. It's much simpler to follow overall, with less edge cases and optimizations that are probably not that relevant in practice.

I'll do one last pass after this.

pkg/vault/contracts/Vault.sol
joaobrunoah: Ok, I'll change the naming

pkg/vault/contracts/Vault.sol
```diff
@@ -1200,28 +1204,37 @@ contract Vault is IVaultMain, VaultCommon, Proxy {
...
-
-                // The mint operation returns exactly `vaultWrappedDelta` shares. To do so, it withdraws underlying
+                // The amount of wrapped tokens to mint is the necessary amount to fulfill the trade
+                // (amountOutWrapped), less the excess amount of wrapped tokens in the buffer (bufferImbalance).
+                // `bufferImbalance` may be positive if buffer has an excess of wrapped assets or negative if the
+                // buffer has an excess of underlying assets. `vaultWrappedDeltaHint` will always be a positive
+                // number, because if `abs(bufferImbalance) > amountOutWrapped` and `bufferImbalance > 0`,
+                // it means the buffer has enough liquidity to fulfill the trade
+                // (i.e. `bufferBalances.getBalanceDerived() >= amountOutWrapped`).
+                bufferImbalance = bufferBalances.getBufferWrappedImbalance(wrappedToken);
```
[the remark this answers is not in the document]
joaobrunoah: So, just for documentation, amountIn and amountOut are calculated in the beginning of _wrapWithBuffer and _unwrapWithBuffer. The amount that we will return to the user is not affected by the precision of buffer imbalance or the buffer balances. That said, if this operation is not as accurate as it should be, the buffer will not be perfectly balanced at the end (by 1 wei, it's just rounding), but that's fine.

pkg/vault/contracts/Vault.sol
```diff
@@ -1185,43 +1187,48 @@ contract Vault is IVaultMain, VaultCommon, Proxy {
...
             // Expected amount of wrapped minted by the wrapper protocol.
             uint256 vaultWrappedDeltaHint;
 
             if (kind == SwapKind.EXACT_IN) {
+                // EXACT_IN requires the exact amount of underlying tokens to be deposited, so we call deposit.
                 // The amount of underlying tokens to deposit is the necessary amount to fulfill the trade
                 // (amountInUnderlying), plus the amount needed to leave the buffer rebalanced 50/50 at the end
-                // (bufferUnderlyingSurplus).
-                vaultUnderlyingDeltaHint = amountInUnderlying + bufferUnderlyingSurplus;
+                // (bufferUnderlyingImbalance). `bufferUnderlyingImbalance` may be positive if buffer has an excess of
```
EndymionJkb, line 1202:
```suggestion
                // (bufferUnderlyingImbalance). `bufferUnderlyingImbalance` may be positive if the buffer has an excess of
```

pkg/vault/contracts/Vault.sol
```diff
@@ -1185,43 +1187,48 @@ contract Vault is IVaultMain, VaultCommon, Proxy {
...
-                    // For the buffer to rebalance during a wrap operation, it must have an excess (surplus) of
-                    // underlying tokens. Since this is an EXACT_OUT wrap (mint), we need to calculate the amount of
-                    // wrapped tokens the buffer should receive to rebalance, based on the underlying token surplus.
-                    // The deposit function is used here, as it performs the inverse of a mint operation.
-                    bufferWrappedSurplus = wrappedToken.previewDeposit(bufferUnderlyingSurplus);
-                }
-
-                // The mint operation returns exactly `vaultWrappedDelta` shares. To do so, it withdraws underlying
+                // EXACT_OUT requires the exact amount of wrapped tokens to be minted, so we call mint.
+                // The amount of wrapped tokens to mint is the necessary amount to fulfill the trade
```
EndymionJkb, line 1213:
```suggestion
                // The amount of wrapped tokens to mint is the amount necessary to fulfill the trade
```

pkg/vault/contracts/Vault.sol
```diff
@@ -1185,43 +1187,48 @@ contract Vault is IVaultMain, VaultCommon, Proxy {
...
+                // (amountOutWrapped), less the excess amount of wrapped tokens in the buffer (bufferWrappedImbalance).
+                // `bufferWrappedImbalance` may be positive if buffer has an excess of wrapped assets or negative if
+                // the buffer has an excess of underlying assets. `vaultWrappedDeltaHint` will always be a positive
+                // number, because if `abs(bufferWrappedImbalance) > amountOutWrapped` and `bufferWrappedImbalance > 0`,
+                // it means the buffer has enough liquidity to fulfill the trade
+                // (i.e. `bufferBalances.getBalanceDerived() >= amountOutWrapped`).
+                int256 bufferWrappedImbalance = bufferBalances.getBufferWrappedImbalance(wrappedToken);
+                vaultWrappedDeltaHint = (amountOutWrapped.toInt256() - bufferWrappedImbalance).toUint256();
+
+                // For Wrap ExactOut, we need to also calculate `vaultUnderlyingDeltaHint` before the mint operation,
```
EndymionJkb, line 1223:
```suggestion
                // For Wrap ExactOut, we also need to calculate `vaultUnderlyingDeltaHint` before the mint operation,
```

pkg/vault/contracts/Vault.sol
```diff
@@ -1185,43 +1187,48 @@ contract Vault is IVaultMain, VaultCommon, Proxy {
...
+                // to approve the transfer of underlying tokens to the wrapper protocol.
+                vaultUnderlyingDeltaHint = wrappedToken.previewMint(vaultWrappedDeltaHint);
+
+                // The mint operation returns exactly `vaultWrappedDeltaHint` shares. To do so, it withdraws underlying
                 // from the Vault and returns the shares. So, the vault needs to approve the transfer of underlying
```
EndymionJkb, line 1228:
```suggestion
                // tokens from the Vault and returns the shares. So, the vault needs to approve the transfer of underlying
```

pkg/vault/contracts/Vault.sol
```diff
@@ -1320,80 +1303,55 @@ contract Vault is IVaultMain, VaultCommon, Proxy {
...
             if (kind == SwapKind.EXACT_IN) {
-                // EXACT_IN requires the exact amount of wrapped tokens to be unwrapped, so we call redeem.
-                // The amount of wrapped tokens to redeem is the amount necessary to fulfill the trade
-                // (amountInWrapped), plus the amount needed to leave the buffer rebalanced 50/50 at the end
-                // (bufferWrappedSurplus).
-                vaultWrappedDeltaHint = amountInWrapped + bufferWrappedSurplus;
+                // EXACT_IN requires the exact amount of wrapped tokens to be unwrapped, so we call redeem. The amount
+                // of wrapped tokens to redeem is the amount necessary to fulfill the trade (amountInWrapped), plus the
+                // amount needed to leave the buffer rebalanced 50/50 at the end (bufferWrappedImbalance).
+                // `bufferWrappedImbalance` may be positive if buffer has an excess of wrapped, or negative if the
```
EndymionJkb, line 1318:
```suggestion
                // `bufferWrappedImbalance` may be positive if the buffer has an excess of wrapped, or negative if the
```

pkg/vault/contracts/Vault.sol
```diff
@@ -1320,80 +1303,55 @@ contract Vault is IVaultMain, VaultCommon, Proxy {
...
-                if (bufferWrappedSurplus > 0) {
-                    // An unwrap operation withdraws wrapped tokens from the buffer and returns underlying tokens.
-                    // For the buffer to rebalance during an unwrap operation, it must have an excess (surplus) of
-                    // wrapped tokens. Since this is an EXACT_OUT unwrap (withdraw), we need to calculate the amount of
-                    // underlying tokens the buffer should receive to rebalance, based on the wrapped token surplus.
-                    // The redeem function is used here, as it performs the inverse of a withdraw operation.
-                    bufferUnderlyingSurplus = wrappedToken.previewRedeem(bufferWrappedSurplus);
-                }
-                vaultUnderlyingDeltaHint = amountOutUnderlying + bufferUnderlyingSurplus;
+                // The amount of underlying tokens to withdraw is the necessary amount to fulfill the trade
```
EndymionJkb, line 1328:
```suggestion
                // The amount of underlying tokens to withdraw is the amount necessary to fulfill the trade
```

pkg/vault/contracts/Vault.sol
```diff
@@ -1320,80 +1303,55 @@ contract Vault is IVaultMain, VaultCommon, Proxy {
...
+                // (amountOutUnderlying), less the excess amount of underlying assets in the buffer
+                // (bufferUnderlyingImbalance). `bufferUnderlyingImbalance` may be positive if buffer has an excess of
```
EndymionJkb, line 1330:
```suggestion
                // (bufferUnderlyingImbalance). `bufferUnderlyingImbalance` may be positive if the buffer has an excess of
```

pkg/vault/contracts/Vault.sol
```diff
@@ -1512,8 +1470,8 @@ contract Vault is IVaultMain, VaultCommon, Proxy {
                 wrappedBalancesAfter
             );
         }
-        // Update the Vault's wrapped reserves, discarding any unexpected surplus of tokens (difference between Vault's
-        // actual and expected balances).
+        // Update the Vault's wrapped reserves, discarding any unexpected surplus of tokens (difference between
+        // Vault's actual and expected balances).
```
EndymionJkb, line 1474:
```suggestion
        // the Vault's actual and expected balances).
```

EndymionJkb (commented):
Some superficial things, as I see this is moving fast. I'm planning to write some documentation on this to make sure I understand it.

Even though it's "simpler" in a sense, there are so many details and cases, it's hard to wrap my head around it - or maybe I need to unwrap, I'm not sure... I definitely need to mint some more brain cells to correct the imbalance and redeem my sanity; the preview operation was not enough :)

pkg/vault/contracts/Vault.sol
```diff
@@ -1185,43 +1187,48 @@ contract Vault is IVaultMain, VaultCommon, Proxy {
...
-                    bufferWrappedSurplus = wrappedToken.previewDeposit(bufferUnderlyingSurplus);
-                }
-
-                // The mint operation returns exactly `vaultWrappedDelta` shares. To do so, it withdraws underlying
-                // from the Vault and returns the shares. So, the vault needs to approve the transfer of underlying
-                // tokens to the wrapper.
-                underlyingToken.forceApprove(address(wrappedToken), amountInUnderlying + bufferUnderlyingSurplus);
+                // EXACT_OUT requires the exact amount of wrapped tokens to be minted, so we call mint.
+                // The amount of wrapped tokens to mint is the amount necessary to fulfill the trade
+                // (amountOutWrapped), less the excess amount of wrapped tokens in the buffer (bufferWrappedImbalance).
```
jubeira, line 1214:
```suggestion
                // (amountOutWrapped), minus the excess amount of wrapped tokens in the buffer (bufferWrappedImbalance).
```

pkg/vault/contracts/Vault.sol
```diff
@@ -1234,40 +1241,16 @@ contract Vault is IVaultMain, VaultCommon, Proxy {
...
-            }
+            // In a wrap operation, the buffer underlying balance increases by `amountInUnderlying` (the amount that
+            // the caller deposited into the buffer) and decreases by `vaultUnderlyingDeltaHint` (the amount of
+            // underlying deposited by the buffer into the wrapper protocol). Conversely, the buffer wrapped balance
+            // decreases by `amountOutWrapped` (the amount of wrapped tokens that the buffer returned to the caller)
+            // and increases by `vaultWrappedDeltaHint` (the amount of wrapped tokens minted by the wrapper protocol).
+            bufferBalances = PackedTokenBalance.toPackedBalance(
+                bufferBalances.getBalanceRaw() + amountInUnderlying - vaultUnderlyingDeltaHint,
+                bufferBalances.getBalanceDerived() + vaultWrappedDeltaHint - amountOutWrapped
+            );
```
jubeira, line 1252: :100: :fire:

pkg/vault/contracts/Vault.sol
```diff
@@ -1320,80 +1303,55 @@ contract Vault is IVaultMain, VaultCommon, Proxy {
...
-                if (bufferWrappedSurplus > 0) {
-                    // An unwrap operation withdraws wrapped tokens from the buffer and returns underlying tokens.
-                    // For the buffer to rebalance during an unwrap operation, it must have an excess (surplus) of
-                    // wrapped tokens. Since this is an EXACT_OUT unwrap (withdraw), we need to calculate the amount of
-                    // underlying tokens the buffer should receive to rebalance, based on the wrapped token surplus.
-                    // The redeem function is used here, as it performs the inverse of a withdraw operation.
-                    bufferUnderlyingSurplus = wrappedToken.previewRedeem(bufferWrappedSurplus);
-                }
-                vaultUnderlyingDeltaHint = amountOutUnderlying + bufferUnderlyingSurplus;
+                // (amountOutUnderlying), less the excess amount of underlying assets in the buffer
```
jubeira, line 1329:
```suggestion
                // (amountOutUnderlying), minus the excess amount of underlying assets in the buffer
```

pkg/vault/test/foundry/BufferVaultPrimitive.t.sol
```diff
@@ -206,12 +269,14 @@ contract BufferVaultPrimitiveTest is BaseVaultTest {
                 })
             )
         );
+
+        assertTrue(dai.allowance(address(vault), address(waDAI)) == 0, "Wrong allowance");
```
jubeira, line 273:
```suggestion
        assertEq(dai.allowance(address(vault), address(waDAI)), 0, "Leftover allowance between vault and wrapper");
```

pkg/vault/test/foundry/BufferVaultPrimitive.t.sol
```diff
@@ -148,14 +148,77 @@ contract BufferVaultPrimitiveTest is BaseVaultTest {
...
+        uint256 totalUnderlyingInBuffer = _wrapAmount / 2;
+        vm.startPrank(lp);
+        router.initializeBuffer(
+            IERC4626(address(waDAI)),
+            totalUnderlyingInBuffer / 3,
+            waDAI.previewDeposit((2 * totalUnderlyingInBuffer) / 3)
+        );
+        vm.stopPrank();
+
+        IBatchRouter.SwapPathExactAmountIn[] memory paths = _exactInWrapUnwrapPath(
```
jubeira, line 172: Late comment about this one: we're using the same method to wrap and unwrap, depending on token from and token to.
While this seems to be correct, you need to remember the order of the arguments to make it right. Here we're testing deposits for example, which means we are going to wrap (tokenFrom is underlying, tokenTo is wrapped).

I'd suggest splitting the method, having one for wrap and another one for unwrap, and you only specify the wrapper contract. That way it's harder to misuse the test helper, and it's immediately clear to the reader whether we're creating a path to wrap or unwrap in order to compare it to the test we're in.

pkg/vault/test/foundry/fork/YieldBearingPoolSwapBase.t.sol
```diff
@@ -572,27 +573,17 @@ abstract contract YieldBearingPoolSwapBase is BaseVaultTest {
...
             // Rounding issues are very small.
-            assertApproxEqAbs(paths[0], vars.expectedUnderlyingDeltaTokenIn, ROUNDING_TOLERANCE, "Wrong path count");
-            assertApproxEqAbs(
-                amounts[0],
-                vars.expectedUnderlyingDeltaTokenIn,
-                ROUNDING_TOLERANCE,
-                "Wrong amounts count"
-            );
+            assertEq(paths[0], vars.expectedUnderlyingDeltaTokenIn, "Wrong path count");
+            assertEq(amounts[0], vars.expectedUnderlyingDeltaTokenIn, "Wrong amounts count");
```
jubeira, line 586: changing `assertApproxEqAbs` for `assertEq` is the happiest moment of the week.

pkg/vault/test/foundry/fork/YieldBearingPoolSwapBase.t.sol
```diff
@@ -917,118 +912,105 @@ abstract contract YieldBearingPoolSwapBase is BaseVaultTest {
...
-            vm.revertTo(snapshotId);
-
-            amountInUnderlying = vaultUnderlyingDeltaHint - bufferUnderlyingSurplus;
+            // Mint converts underlying to wrapped. The rebalance logic can introduce rounding issues, so we
+            // should consider it in our result preview. The logic below reproduces Vault's rebalance logic.
+            int256 bufferImbalance = bufferBalances.getBufferWrappedImbalance(wToken);
+            uint256 vaultWrappedDeltaHint = uint256(int256(amountOutWrapped) - bufferImbalance);
+            uint256 vaultUnderlyingDeltaHint = wToken.previewMint(vaultWrappedDeltaHint);
+
+            if (bufferImbalance != 0) {
```
jubeira, line 984: do these `if`s still matter now?

jubeira (approved):
LGTM :raised_hands: wonderful job.

Thanks for all the iterations :face_holding_back_tears:

pkg/vault/test/foundry/fork/YieldBearingPoolSwapBase.t.sol
joaobrunoah: I left them to make sure the behavior of the buffer is the same as before, even removing the last `if` from the vault code.

PR #978, opened 2024-09-14
Safe cast stragglers

### LINKED ISSUE #181: Review unsafe casts
See comments in Vault regarding casts in events:
```
// TODO No they aren't anymore (stored as uint112)! Review this.
```

EndymionJkb:
# Description

Looking through the issues, was about to close 181 (since the comment it references is long gone), and thought I'd check for any unsafe casts. We'd recently gone through and added SafeCasts as a result of auditor comments. There were a couple very obscure cases still uncovered (outside tests).

## Type of change

- [ ] Bug fix 
- [ ] New feature 
- [ ] Breaking change 
- [ ] Dependency changes
- [X] Code refactor / cleanup
- [ ] Other

## Checklist:

## Issue Resolution

Closes #181


pkg/solidity-utils/contracts/helpers/FactoryWidePauseWindow.sol
```diff
@@ -14,6 +16,8 @@ pragma solidity ^0.8.24;
  * When the buffer period expires, it will unpause automatically, and remain permissionless forever after.
  */
 contract FactoryWidePauseWindow {
+    using SafeCast for *;
```
jubeira, line 19: Hmm where are we using this?

jubeira (commented):
Thanks @EndymionJkb; just one question, otherwise LGTM.

pkg/solidity-utils/contracts/helpers/FactoryWidePauseWindow.sol
EndymionJkb: Removed... it *was* used to case the timestamp in the `getNewPoolPauseWindowEndTime` function, but then I realized that would revert when it overflowed (in 2106, but still...), but the inequality would still work, so as it is now it will return 0 after 2106 instead of reverting, as it should. No longer needed the cast, but forgot to remove it.

jubeira (approved):
LGTM!

PR #982, opened 2024-09-16
Restrict `addLiquidityToBuffer` to proportional

### LINKED ISSUE #977: Consider restricting buffer addLiquidity to proportional
Per ToB comment, so that we don't lose it with everything else going on.

We aren't checking against the buffer minimum in that case; we can either to add those checks, or eliminate the issue by only supporting proportional add liquidity for buffers.

If it's really required to add just one or the other token, this can be done externally, by calling wrap/unwrap, deposit/withdraw, etc.

joaobrunoah:
# Description

Per issue #977, the current implementation of `addLiquidityToBuffer` does not implement the buffer minimum check. This check is not necessary if we only support proportional add liquidity.

## Type of change

- [ ] Bug fix 
- [ ] New feature 
- [ ] Breaking change 
- [ ] Dependency changes
- [x] Code refactor / cleanup
- [ ] Other

## Checklist:

## Issue Resolution

Closes #977


pkg/vault/contracts/Router.sol
```diff
@@ -654,66 +654,83 @@ contract Router is IRouter, RouterCommon, ReentrancyGuardTransient {
...
+     * @dev Can only be called by the Vault.
+     * @param wrappedToken Address of the wrapped token that implements IERC4626
+     * @param amountUnderlyingRaw Amount of underlying tokens that will be deposited into the buffer
+     * @param amountWrappedRaw Amount of wrapped tokens that will be deposited into the buffer
+     * @param sharesOwner Address that will own the deposited liquidity. Only this address will be able to
+     * remove liquidity from the buffer
+     * @return issuedShares the amount of tokens sharesOwner has in the buffer, expressed in underlying token amounts.
+     * (This is the BPT of an internal ERC4626 buffer)
+     */
+    function initializeBufferHook(
```
EndymionJkb, line 681: Separating the initialize and add hooks is an improvement by itself (though it costs bytecode). It's clearer, more closely parallels the vault functions, and avoids passing flag arguments.

pkg/vault/contracts/Router.sol
```diff
@@ -654,66 +654,83 @@ contract Router is IRouter, RouterCommon, ReentrancyGuardTransient {
...
                         )
                     )
                 ),
                 (uint256)
             );
     }
 
+    /**
+     * @notice Hook for initializing a vault buffer.
+     * @dev Can only be called by the Vault.
```
EndymionJkb, line 672:
```suggestion
     * @dev Can only be called by the Vault. Buffers must be initialized before use.
```

EndymionJkb (commented):
First pass looks good, with a few minor suggestions. Haven't gone through the tests in detail yet

pkg/vault/contracts/VaultAdmin.sol
```diff
@@ -498,30 +497,27 @@ contract VaultAdmin is IVaultAdmin, VaultCommon, Authentication, VaultGuard {
...
-            wrappedToken.previewRedeem(bufferBalances.getBalanceDerived());
+        // To proportionally add liquidity to buffer, we need to calculate the buffer invariant ratio. It's calculated
+        // as the amount of buffer shares the sender wants to issue (which in practice is the value that the sender
+        // will add to the buffer, expressed in underlying token amounts), divided by the total shares of
+        // the buffer.
+        uint256 invariantRatio = exactSharesToIssue.divUp(_bufferTotalShares[wrappedToken]);
+        // Multiply the current buffer balance by the invariant ratio to calculate the amount of underlying and wrapped
+        // tokens to add, keeping the proportion of the buffer.
+        amountUnderlyingRaw = bufferBalances.getBalanceRaw().mulUp(invariantRatio);
+        amountWrappedRaw = bufferBalances.getBalanceDerived().mulUp(invariantRatio);
```
jubeira, line 516:
```suggestion
        // Multiply the current buffer balance by the invariant ratio to calculate the amount of underlying and wrapped
        // tokens to add, keeping the proportion of the buffer.
        uint256 totalShares = _bufferTotalShares[wrappedToken];
        amountUnderlyingRaw = bufferBalances.getBalanceRaw().mulDivUp(exactSharesToIssue, totalShares);
        amountWrappedRaw = bufferBalances.getBalanceDerived().mulDivUp(exactSharesToIssue, totalShares);
```

This should mimic more closely the logic for add liquidity proportional. It's more accurate to multiply first, and since we're multiplying and dividing we don't need the usual fixed point scaling / unscaling by FP(1).

pkg/vault/contracts/test/VaultAdminMock.sol
```diff
@@ -55,13 +60,33 @@ contract VaultAdminMock is IVaultAdminMock, VaultAdmin {
...
+        if (underlyingAmount > 0) {
+            IERC20(wrappedToken.asset()).transferFrom(msg.sender, address(this), underlyingAmount);
+            _reservesOf[IERC20(wrappedToken.asset())] += underlyingAmount;
+            _bufferTotalShares[wrappedToken] += underlyingAmount;
+        }
+        if (wrappedAmount > 0) {
+            IERC20(address(wrappedToken)).transferFrom(msg.sender, address(this), wrappedAmount);
+            _reservesOf[IERC20(address(wrappedToken))] += wrappedAmount;
+            _bufferTotalShares[wrappedToken] += wrappedToken.previewRedeem(wrappedAmount);
+        }
```
jubeira, line 75: we're not minting shares to the caller; can't we just mint the assets to the vault instead of using `transferFrom`?
Otherwise consider assigning shares to the caller.

jubeira (commented):
LGTM! Thanks for pushing this @joaobrunoah.

I do have one relatively important comment regarding the math. I'll do one final pass once that's addressed :)

pkg/vault/contracts/VaultAdmin.sol
joaobrunoah: Makes sense!

pkg/vault/contracts/test/VaultAdminMock.sol
joaobrunoah: Since this was just a test, I did in the quicker way posible, but I'll send the shares to the caller.

pkg/interfaces/contracts/vault/IRouter.sol
```diff
@@ -331,19 +331,20 @@ interface IRouter {
...
+     * @dev Requires the buffer to be initialized beforehand. Restricting adds to proportional simplifies the Vault
+     * code, avoiding rounding issues and minimum amount checks. It is possible to add unbalanced by interacting
+     * with the wrapper contract directly.
      * @param wrappedToken Address of the wrapped token that implements IERC4626
-     * @param amountUnderlyingRaw Amount of underlying tokens that will be deposited into the buffer
-     * @param amountWrappedRaw Amount of wrapped tokens that will be deposited into the buffer
-     * @return issuedShares the amount of tokens sharesOwner has in the buffer, denominated in underlying tokens
-     * (This is the BPT of the vault's internal ERC4626 buffers)
+     * @param exactSharesToIssue The value in underlying tokens that `sharesOwner` wants to add to the buffer,
+     * in native token decimals
```
jubeira, line 340: If underlying token and wrapped token don't have the same decimals, the difference is absorbed by `preview` right? I'd clarify whether this refers to the wrapped token or the underlying token decimals.

pkg/vault/contracts/test/VaultAdminMock.sol
```diff
@@ -55,13 +60,37 @@ contract VaultAdminMock is IVaultAdminMock, VaultAdmin {
...
+            _bufferTotalShares[wrappedToken] += underlyingAmount;
+            _bufferLpShares[wrappedToken][msg.sender] += underlyingAmount;
+        }
+        if (wrappedAmount > 0) {
+            IERC20(address(wrappedToken)).transferFrom(msg.sender, address(this), wrappedAmount);
+            _reservesOf[IERC20(address(wrappedToken))] += wrappedAmount;
+            uint256 issuedSharesAmount = wrappedToken.previewRedeem(wrappedAmount);
+            _bufferTotalShares[wrappedToken] += issuedSharesAmount;
+            _bufferLpShares[wrappedToken][msg.sender] += issuedSharesAmount;
+        }
```
jubeira, line 79: I don't think this math really adds up with the actual operation; is that intended? We're not considering total supply or rates when computing the shares. E.g. if we only add underlying amount, that will increase the shares without considering the existing liquidity, which doesn't add up.

Do we really need this method? If we do, I'd consider something even more 'obviously manual': just also send how many shares to mint to the sender as an argument, and add those to the total too.

EDIT: lol buffers are linear so this is actually fine. Feel free to leave like this.

pkg/vault/test/foundry/BufferVaultPrimitive.t.sol
```diff
@@ -449,143 +449,169 @@ contract BufferVaultPrimitiveTest is BaseVaultTest {
...
+            lpSharesBeforeAdd + lpSharesToAdd,
             vault.getBufferOwnerShares(IERC4626(address(waDAI)), lp),
             "LP Buffer shares is wrong"
         );
-        assertEq(lpSharesAdded, underlyingAmountIn + waDAI.previewRedeem(wrappedAmountIn), "Issued shares is wrong");
+
+        // Since underlying and wrapped amount in are rounded up, it can introduced rounding issues in favor of the
+        // vault (`bufferInvariantDelta - 2e6 < sharesToAdd < bufferInvariantDelta`).
+        uint256 bufferInvariantDelta = expectedUnderlyingAmountIn + waDAI.previewRedeem(expectedWrappedAmountIn);
+        assertApproxEqAbs(lpSharesToAdd, bufferInvariantDelta - 1e6, 1e6, "Issued shares is wrong");
```
jubeira, line 521: I'd consider using `Ge` / `Le` to show that the error only goes one way.

jubeira (approved):
LGTM! Great job.

pkg/vault/contracts/test/VaultAdminMock.sol
joaobrunoah: I just added this method because I need an easy way to add tokens to a buffer and leave it unbalanced so that I can test the rebalance. Since the `addLiquidityToBuffer` is proportional now, I don't have an easy way to do it (I could wrap/unwrap, but then I'd need to modify the tests, which wouldn't add any value to them).

I changed this method's name to `addLiquidityToBufferUnbalancedForTests`, which removes confusion about the actual operation and clarifies that it's only used for testing purposes.

pkg/vault/test/foundry/BufferVaultPrimitive.t.sol
joaobrunoah: I usually used approxEq and Ge or Le, but I found out that this test covers both cases. Is it a bad pattern?

PR #983, opened 2024-09-16
Swap fees equivalence (token in)

### LINKED ISSUE #994: Add swap fee logging to add/remove liquidity events
Per discussion, we have the raw swap fee amount and token given below - but we don't have the same for add/remove liquidity, yield, etc. (We still need the percentage, in case it is dynamic.) We have events for when fees are *collected* (and withdrawn), but don't have the block-by-block fee *charged* events we had in v2, since we are not computing the "split" in the Vault; as an optimization, the Vault only knows the aggregate percentage, and the `ProtocolFeeController` handles splitting. Moving this computation out of the Vault, and skipping the associated events, represents a very large gas savings.

We also have hooks in v3, which can also take fees; those would need to be tracked by the hook. Since fees are more complicated and not handled in the Vault, it should be sufficient to track them using the hook and "collected" events. (They will then lag a bit, but could be refreshed at any time, as collection is permissionless.)

After more discussion, it seems we *do* want to track all swap fees block by block after all. To accomplish this, we will need to add swap fees to add/remove liquidity operations (i.e., to the `PoolBalanceChanged` event.

---------

Old text, when we were planning on removing the events:

We do have getters that return the uncollected fees, so anyone who really needs completely up-to-date fees could read them onchain.

Given all this, maybe we don't need `swapFeeAmount` and `swapFeeToken` in the Swap event. The gas savings could be significant, as we are downscaling *only* to compute the fee for this event, so the benefit would be greater than just emitting less data in the event itself. (We might even be able to refactor further, given that we don't need to retain this value all the way to the end, don't need the swapFeeToken, etc.)

### LINKED ISSUE #998: Ensure ExactIn and ExactOut swaps are symmetrical
Update fee calculation mechanism so that ExactIn and ExactOut have the same result, no matter the pool type.
In other words, if swapExactIn(Ai) = Bo, then swapExactOut(Bo) = Ai when swapping A --> B.

Originally, we took fees on tokenOut, reasoning that since charging tokenIn or tokenOut would be equivalent, it would likely be better to charge fees on the token people "want," rather than the one they're selling.

Then we added dynamic swap fees, and noticed that charging fees this way meant the arguments to the "inner" swap function could reflect the adjusted amountIn: but the before/dynamic fee hook calls could not do this in principle, since we didn't know what the fee was yet, so we couldn't "inflate" the scaled value.

This was undesirable, and our solution was to instead always charge fees on the calculated amount, which allowed us to adjust it *after* the swap, and preserve the integrity of the input arguments to the hooks called before the main swap.

Unfortunately, this led to an asymmetry between ExactIn and ExactOut swaps (ExactOut would always be a bit better), since the fees weren't always going through the pool math, introducing error.

Some numbers from fuzz testing:
For a weighted pool 99/1, swap fee up to 10%, the discrepancy can be 1-1.5% between exact in and exact out (quite large).
For an 80/20 pool, swap fee up to 3%, the discrepancy is < 0.02%.

So... we need to go back to tokenIn or tokenOut, since breaking swap amount symmetry is worse than ensuring consistent arguments to hooks.

Since there are paths in batch swaps where we need to pay up front (in tokenIn), we might as well charge on tokenIn. As a bonus, this is consistent with v2 (and also Uniswap).

	•	If it's exact in, just charge the fee in token in (amount * feePercentage), and then proceed. The pool will do the math without the fee, resulting in less amount out. The amount in that the pool receives is amountGiven (1 - feePercentage)
	•	If it's exact out, compute the amount in. To match exact in case, this amount calculated equals amountGivenExactIn (1 - feePercentage), so we amplify the amount in calculated by 1 / (1 - feePercentage)` to get to the original amount given exact in.

jubeira:
# Description

Update fee calculation mechanism so that swaps exact out and exact in have the same result, no matter the pool type.
In other words, if `swapExactIn(Ai) = Bo`, then `swapExactOut(Bo) = Ai` when swapping A --> B.

It also adjusts `_computeAndChargeAggregateSwapFees` to compute both total fees raw and aggregate fees raw more accurately, preventing rounding issues for the total swap fees raw included in the swap event.

## Type of change

- [x] Bug fix 
- [ ] New feature 
- [ ] Breaking change 
- [ ] Dependency changes
- [ ] Code refactor / cleanup
- [ ] Other

## Checklist:

## Issue Resolution

Closes #994 
Closes #998


pkg/vault/test/foundry/E2eErc4626Swaps.t.sol
```diff
@@ -349,9 +349,13 @@ contract E2eErc4626SwapsTest is BaseERC4626BufferTest {
             waUSDC.previewRedeem(balancesBefore.balances.vaultTokens[balancesBefore.waUsdcIdx]);
         uint256 vaultTotalUsdcAfter = balancesAfter.balances.vaultTokens[balancesAfter.usdcIdx] +
             waUSDC.previewRedeem(balancesAfter.balances.vaultTokens[balancesAfter.waUsdcIdx]);
+        console.log("vault usdc after: ", vaultTotalUsdcAfter);
+        console.log("vault usdc before: ", vaultTotalUsdcBefore);
```
joaobrunoah, line 353: Remaining CL

pkg/vault/test/foundry/E2eErc4626Swaps.t.sol
```diff
@@ -349,9 +349,13 @@ contract E2eErc4626SwapsTest is BaseERC4626BufferTest {
...
+        // TODO: check sign; vaultTotalUsdcAfter - vaultTotalUsdcBefore = senderUsdcDelta, solve for usdcAfter
+        // to prevent underflows
```
joaobrunoah, line 355: Is this TODO resolved? The signs are right because, in a Do/Undo operation, the `vaultTotalUsdcAfter` will always be bigger than `vaultTotalUsdcBefore` (The user should not get USDC or DAI out of the vault, or else the vault would be drained).
It worried me a bit that `vaultTotalUsdcAfter - vaultTotalUsdcBefore` threw an underflow, since the vault should never lose USDC tokens, but I imagine that happened with a small fee and a very small swap?

pkg/vault/test/foundry/VaultSwap.t.sol
```diff
@@ -546,13 +544,17 @@ contract VaultSwapTest is BaseVaultTest {
...
+        // Pool fees are taken on the calculated end
         if (kind == SwapKind.EXACT_OUT) {
+            // Fees applied to token in (user pays more usdc)
             usdcFee = fee;
-            usdcProtocolFee = protocolFee;
         } else {
+            // Fees applied to token out (user gets less dai).
             daiFee = fee;
-            daiProtocolFee = protocolFee;
         }
```
joaobrunoah, line 554: Just to check if I understood this part: To the Vault, fees are always taken in terms of token in amounts, but it's transparent to the user. So, the user feels like it's paying fees with the calculated end. We can calculate the usdc and dai fees like this because the pool is linear and the rates are 1, so the token in fee and the token calculated fee would be the same. Is that it?

joaobrunoah (commented):
Just a few minor comments.

EndymionJkb (approved):
I'm getting bleary-eyed looking at all the tests (this is definitely a two-approval kind of thing), but it looks right to me. As discussed, we can make another PR on top that addresses simplifying the Swap event.

pkg/vault/contracts/Vault.sol
```diff
@@ -382,12 +389,9 @@ contract Vault is IVaultMain, VaultCommon, Proxy {
...
         if (vaultSwapParams.kind == SwapKind.EXACT_IN) {
-            // Swap fee is always a percentage of the amountCalculated. On ExactIn, subtract it from the calculated
-            // amountOut. Round up to avoid losses during precision loss.
-            locals.swapFeeAmountScaled18 = amountCalculatedScaled18.mulUp(swapState.swapFeePercentage);
-
-            // Need to update `amountCalculatedScaled18` for the onAfterSwap hook.
-            amountCalculatedScaled18 -= locals.swapFeeAmountScaled18;
+            // Restore the original input value; this function should not mutate memory inputs.
+            // At this point swap fee amounts have already been computed for EXACT_IN.
+            poolSwapParams.amountGivenScaled18 = swapState.amountGivenScaled18;
```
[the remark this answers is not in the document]
jubeira: > Since we've deleted other stuff from SwapLocals and freed up some space, consider not re-using amountGivenScaled18, so that we don't need to restore it? Could have amountGivenScaled18ForCalc/Mutable/PoolMath; something like that.

The problem is not about the locals, but about `poolSwapParams`. We need to send the whole struct to `onSwap`, and the only thing we're modifying is the `amountGivenScaled18`. So we either do something like this PR, or we copy the whole structure with a modified `amountGivenScaled18`.

pkg/pool-weighted/test/foundry/E2eSwap.t.sol
```diff
@@ -112,22 +112,38 @@ contract E2eSwapWeightedTest is E2eSwapTest {
...
-            exactAmountOutDo + feesTokenB,
+            exactAmountOutDo,
             0,
             MAX_UINT128,
             false,
             bytes("")
         );
-        uint256 feesTokenA = vault.getAggregateSwapFeeAmount(pool, tokenA);
+
+        uint256 feesTokenA = vault.getAggregateSwapFeeAmount(address(poolWithMutableWeights), tokenA);
```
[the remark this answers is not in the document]
joaobrunoah: It was a bug that accidentally worked.

pkg/pool-weighted/test/foundry/LiquidityApproximationWeighted.t.sol
```diff
@@ -170,11 +170,35 @@ contract LiquidityApproximationWeightedTest is LiquidityApproximationTest {
...
+    function testAddLiquidityProportionalAndRemoveExactInExtremeWeights__Fuzz(
+        uint256 exactBptAmount,
+        uint256 swapFeePercentage,
+        uint256 weightDai
     ) public {
         exactBptAmount = bound(exactBptAmount, minAmount, maxAmount / 2 - 1);
         swapFeePercentage = bound(swapFeePercentage, minSwapFeePercentage, maxSwapFeePercentage);
         weightDai = bound(weightDai, 1e16, 99e16);
 
+        // TODO: check why the difference grows so large in the extremes (high fees, high dai weight).
```
joaobrunoah, line 196: As discussed, RemoveLiquiditySingleTokenExactIn charges the fee in token out, which amplifies the error when compared with the Swap, that charges the fee in token in.
Maybe this is not the whole explanation, but if it is, we could remove this TODO and add this comment here (and wherever we compare RemoveLiquiditySingleTokenExactIn with Swap).

pkg/vault/contracts/Vault.sol
```diff
@@ -1031,39 +1018,40 @@ contract Vault is IVaultMain, VaultCommon, Proxy {
...
-                poolData.decimalScalingFactors[index],
-                poolData.tokenRates[index]
-            );
-
             // Both Swap and Yield fees are stored together in a PackedTokenBalance.
             // We have designated "Raw" the derived half for Swap fee storage.
             bytes32 currentPackedBalance = _aggregateFeeAmounts[pool][token];
             _aggregateFeeAmounts[pool][token] = currentPackedBalance.setBalanceRaw(
-                currentPackedBalance.getBalanceRaw() + totalFeesRaw
+                currentPackedBalance.getBalanceRaw() + aggregateSwapFeeAmountRaw
```
joaobrunoah, line 1054: Much clearer!

pkg/vault/test/.contract-sizes/Vault
```diff
@@ -1,2 +1,2 @@
-Bytecode	23.264
-InitCode	24.604
\ No newline at end of file
+Bytecode	23.062
+InitCode	24.402
```
joaobrunoah, line 2: Nice!

pkg/vault/test/foundry/unit/VaultUnit.t.sol
```diff
@@ -79,31 +79,32 @@ contract VaultUnitTest is BaseTest {
         uint256 tokenIndex = 0;
         vault.manualSetAggregateSwapFeeAmount(pool, dai, 0);
 
-        uint256 swapFeeAmountScaled18 = 1e18;
-        uint256 protocolSwapFeePercentage = 10e16;
+        uint256 totalSwapFeeAmountScaled18 = 1.123456789987654321e18;
+        uint256 aggregateSwapFeePercentage = 10.98765432112345e16;
```
joaobrunoah, line 83: I believe these values are random? Maybe it'd be good to comment on why these values were chosen.

pkg/vault/contracts/Vault.sol
```diff
@@ -1031,39 +1018,40 @@ contract Vault is IVaultMain, VaultCommon, Proxy {
      * Splitting the fees and event emission occur during fee collection.
      * Should only be called in a non-reentrant context.
      *
-     * @return totalFeesRaw Sum of protocol and pool creator fees raw
+     * @return totalSwapFeeAmountRaw Total swap fees raw (LP + aggregate protocol fees)
+     * @return aggregateSwapFeeAmountRaw Sum of protocol and pool creator fees raw
      */
     function _computeAndChargeAggregateSwapFees(
```
joaobrunoah, line 1024: I'd say the name of this function should change, since it's computing the total and aggregate swap fees. But, I understand that it only charge the aggregate swap fees, so probably it's better to leave it as is.

joaobrunoah (approved):
LGTM!

pkg/pool-weighted/test/foundry/E2eSwap.t.sol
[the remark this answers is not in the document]
EndymionJkb: Have to wonder how many more of those there are...

pkg/vault/test/foundry/unit/VaultUnit.t.sol
EndymionJkb: I made it a fuzz test and it passed. Could just do that.

```
    function testComputeAndChargeAggregateSwapFees__Fuzz(uint256 totalSwapFeeAmountScaled18, uint256 aggregateSwapFeePercentage) public {
        totalSwapFeeAmountScaled18 = bound(totalSwapFeeAmountScaled18, 0, MAX_UINT128);
        aggregateSwapFeePercentage = bound(aggregateSwapFeePercentage, 1e12, 50e16);
```

pkg/vault/contracts/Vault.sol
```diff
@@ -486,22 +476,14 @@ contract Vault is IVaultMain, VaultCommon, Proxy {
...
         emit Swap(
             vaultSwapParams.pool,
             vaultSwapParams.tokenIn,
             vaultSwapParams.tokenOut,
             amountInRaw,
             amountOutRaw,
             swapState.swapFeePercentage,
-            locals.swapFeeAmountRaw,
-            locals.swapFeeToken
+            locals.totalSwapFeeAmountRaw
```
EndymionJkb, line 486: Consider adding a comment defining `totalSwapFeeAmountRaw` as the total swap fee, which later gets divided between the protocol, pool creator, and LPs. It's potentially a bit confusing, as we usually report the aggregate fees (e.g., in the getters), and this is a change from what we did before.

pkg/vault/contracts/Vault.sol
```diff
@@ -629,12 +611,13 @@ contract Vault is IVaultMain, VaultCommon, Proxy {
         LiquidityLocals memory locals;
         locals.numTokens = poolData.tokens.length;
         amountsInRaw = new uint256[](locals.numTokens);
-        uint256[] memory swapFeeAmountsScaled18;
+        // swapFeeAmounts shall store scaled18 amounts first, and then it will be reused to store raw amounts.
```
EndymionJkb, line 614:
```suggestion
        // `swapFeeAmounts` stores scaled18 amounts first, and is then reused to store raw amounts.
```
Would discourage "shall"; sounds like the bible (though I guess that's "shalt").

pkg/vault/contracts/Vault.sol
```diff
@@ -883,12 +867,13 @@ contract Vault is IVaultMain, VaultCommon, Proxy {
     {
         LiquidityLocals memory locals;
         locals.numTokens = poolData.tokens.length;
-        uint256[] memory swapFeeAmountsScaled18;
         amountsOutRaw = new uint256[](locals.numTokens);
+        // swapFeeAmounts shall store scaled18 amounts first, and then it will be reused to store raw amounts.
```
EndymionJkb, line 871:
```suggestion
        // `swapFeeAmounts` stores scaled18 amounts first, and is then reused to store raw amounts.
```

pkg/vault/contracts/Vault.sol
EndymionJkb: Could just be `_computeAndChargeSwapFees`, though I think it's also fine as is.

There is a comment in the unchanged code saying we're not emitting events, though, which is now a bit misleading.

"Note that this computes the aggregate total of the protocol fees and stores it, without emitting any events."

Maybe just add "The total swap fees are reported in the `PoolBalancesChanged` event"

EndymionJkb (approved):
This seems to be changing, so I'll get some last comments in. Was going to comment on the failing test, but that seems to be fixed.

pkg/pool-weighted/test/foundry/LiquidityApproximationWeighted.t.sol
jubeira: These tests will get one more pass. For now I'm just fixing them and taking some notes; the main point is to get the code merged for now.

pkg/vault/contracts/Vault.sol
jubeira: that's already in the struct definition :)

PR #984, opened 2024-09-17
Minor `WeightedMath` rounding fix for `computeBalance`

jubeira:
# Description

`computeBalance` always needs to round the final result up (both when used in `add` and `remove` in `BasePoolMath`).
But the invariant ratio can be either greater than FP(1) or smaller than FP(1) depending on whether it's used inside `add` or `remove`.

`k^x` is [monotonically increasing for k > 1 and monotonically decreasing for k < 1](https://www.wolframalpha.com/input?i=plot+k%5Ex+for+k+%3D+%7B0.5%3B+1.5%7D), so the way we round the exponent depends on k. This PR fixes this.

## Type of change

- [x] Bug fix 
- [ ] New feature 
- [ ] Breaking change 
- [ ] Dependency changes
- [ ] Code refactor / cleanup
- [ ] Other

## Checklist:

## Issue Resolution

N/A


pkg/solidity-utils/contracts/math/WeightedMath.sol
```diff
@@ -121,10 +121,21 @@ library WeightedMath {
...
 
-        // Rounds result up overall.
+        // Rounds result up overall, rounding up the two individual steps:
+        // - balanceRatio = invariantRatio ^ (1 / weight)
+        // - newBalance = balance * balanceRatio
+        //
+        // Regarding `balanceRatio`, the exponent is always > FP(1), but the invariant ratio can be either greater or
+        // lower than FP(1) depending on whether this is solving an `add` or a `remove` operation.
+        // - For i > 1, we need to round the exponent up, as i^x is monotonically increasing for i > 1.
+        // - For i < 1, we need to round the exponent down, as as i^x is monotonically decreasing for i < 1.
```
EndymionJkb, line 131: Is it worth short-circuiting the case where the invariantRatio = 1; newBalance = currentBalance without doing anything? Did you find a numeric example in the tests showing the effect of the rounding? (If so, might be good to include it, but not important.)

EndymionJkb (approved):
Looks good to me! Our laser-focus on rounding is ironing out the last tiny effects that we never noticed before. Why would this make the gas go down so much? (Maybe the snaps were out-of-date?)

Rounding directions elsewhere look consistent:
In computeOutGivenExactIn, i <= 1, so we round the exponent down
In computeInGivenExactOut, i >= 1, so we round the exponent up

joaobrunoah (approved):
I agree with @EndymionJkb that would be good to have a test that fails for the current implementation of Weighted Math. Weird that we changed it and no tests failed. Anyway, the change in the code makes sense.

pkg/solidity-utils/contracts/math/WeightedMath.sol
jubeira: I don't think so. The ratio is never going to be 1 unless you are adding / removing exactly 0 BPT tokens.

PR #985, opened 2024-09-17
PoC - Stateful fuzzing tests

joaobrunoah:
# Description

This PR is currently a PoC of a stateful fuzzing test. The new test, starting with the word `invariant`, randomly picks a function inside the `OperationsHandler.sol` contract, and fuzz the input of the chosen function. It does that 500 times (`depth` variable inside foundry.toml) and then execute the invariant test to check if the pool invariant was respected (The new pool invariant should not decrease, since only swaps were executed).

It repeats the test 100 times (`runs` variable inside foundry.toml), resulting in a total of 50000 swap executions. It does not revert the test if a swap fails, since the idea is to check if it's possible to break the invariant, and not the swap operation.

## Type of change

- [ ] Bug fix 
- [x] New feature 
- [ ] Breaking change 
- [ ] Dependency changes
- [ ] Code refactor / cleanup
- [ ] Other

## Checklist:


EndymionJkb:
I seem to recall the auditors (ToB?) saying forge was really bad for this sort of thing. Are we still pursuing it?

joaobrunoah:
> I seem to recall the auditors (ToB?) saying forge was really bad for this sort of thing. Are we still pursuing it?

Hey @EndymionJkb . So, auditors use Echidna and Medusa, both are excellent tools but would require a bigger learning curve. Checking this Repo: https://github.com/devdacian/solidity-fuzzing-comparison/tree/main, looks like Forge is not that bad and could be a first step to understand stateful tests without investing too much time. After talking to @jubeira , we decided to try it out.

joaobrunoah:
Closed since we are adopting Stateful Fuzz testing with Medusa

PR #988, opened 2024-09-18
Payable multicall

### LINKED ISSUE #973: Router eth manipulation tweaks
- Make `multicall` payable.
- Check multiple steps with ETH. Right now `returnEth` will return everything, so if there's more than one step it won't work.

elshan-eth:
# Description

Currently, it is not possible to use multicall with ETH. This PR fixes this and makes multicall `payable` and modifies returnEth to prevent double returns.

## Type of change

- [X] Bug fix 
- [ ] New feature 
- [ ] Breaking change 
- [ ] Dependency changes
- [ ] Code refactor / cleanup
- [ ] Other

## Checklist:

## Issue Resolution

Closes #973


pkg/vault/contracts/RouterCommon.sol
```diff
@@ -83,6 +87,15 @@ abstract contract RouterCommon is IRouterCommon, VaultGuard {
...
     }
 
+    /// @notice Locks the return of excess ETH to the sender until the end of the function.
+    /// @dev This modifier must be used after `saveSender` to ensure that the sender is saved.
+    modifier useMulticall() {
+        _isReturnEthLockedSlot().tstore(true);
+        _;
+        _isReturnEthLockedSlot().tstore(false);
+        _returnEth(_getSenderSlot().tload());
+    }
```
EndymionJkb, line 97: Seems error-prone to require two modifiers. Can we just combine them? Repeating "saveSender" in this one's name so that people don't try to use both anyway.

```suggestion
    /**
     * @notice Locks the return of excess ETH to the sender until the end of the function.
     * @dev This also encompasses the `saveSender` functionality.
     */
    modifier saveSenderAndManageEth() {
        bool isExternalSender = _saveSender();

        // Lock the return of ETH during execution
        _isReturnEthLockedSlot().tstore(true);
        _;
        _isReturnEthLockedSlot().tstore(false);

        _returnEth(_getSenderSlot().tload());
        _discardSenderIfRequired(isExternalSender);
    }
```

pkg/vault/contracts/test/RouterCommonMock.sol
```diff
@@ -55,4 +55,12 @@ contract RouterCommonMock is RouterCommon {
     function mockSendTokenOut(address sender, IERC20 tokenOut, uint256 amountOut, bool wethIsEth) external {
         return _sendTokenOut(sender, tokenOut, amountOut, wethIsEth);
     }
+
+    function manualReturnETH() public payable {
+        //   _returnEth(msg.sender);
+    }
+
+    function assertETHBalancerMoreZero() public payable {
```
EndymionJkb, line 63:
```suggestion
    function assertNonZeroETHBalance() public payable {
```

EndymionJkb (commented):
Some suggestions for naming and simplification.

EndymionJkb:
Could also do something like this, to manually alter the amount of ETH in the middle:

```
function testUseMulticallModifierWithMultipleFunctions() public {
    uint256 extraAmount = 0.5 ether;

    uint256 balanceBefore = alice.balance;

    bytes[] memory calls = new bytes[](3);
    calls[0] = abi.encodeWithSelector(RouterCommonMock.sendExtraEth.selector, alice, extraAmount);
    calls[1] = abi.encodeWithSelector(RouterCommonMock.manualReturnETH.selector);
    calls[2] = abi.encodeWithSelector(RouterCommonMock.assertETHBalance.selector, balanceBefore - extraAmount);

    vm.prank(alice);
    routerMock.multicall{ value: 1 ether }(calls);

    uint256 balanceAfter = alice.balance;
    assertEq(balanceAfter, balanceBefore, "Value wasn't returned");
}

RouterCommonMock:
function sendExtraEth(address recipient, uint256 amount) public payable {
    payable(recipient).sendValue(amount);
}

function assertETHBalance(uint256 amount) public payable {
    require(address(msg.sender).balance == amount, "Wrong ETH balance");
}
```

EndymionJkb:
Note this auditor comment; I think we need to add payable a couple more places

"Consequently, every (batch) router method which should be multicallable needs the payable modifier.
It is currently missing on removeLiquidityCustom and removeLiquidityRecovery."

EndymionJkb (commented):
Looks good! See the related auditor comment though; I think we need to add payable to the custom removes (even though not used, so we don't revert in the middle of a multicall that contains payable operations).

"...every (batch) router method which should be multicallable needs the payable modifier. It is currently missing on removeLiquidityCustom and removeLiquidityRecovery."

pkg/vault/test/foundry/RouterCommon.t.sol
```diff
@@ -156,6 +156,31 @@ contract RouterCommonTest is BaseVaultTest {
...
+    }
+
+    function testSaveSenderAndManageEthModifierWithMultipleFunctions() public {
+        uint256 extraAmount = 0.5 ether;
+        uint256 balanceBefore = alice.balance;
+
+        bytes[] memory calls = new bytes[](3);
+        calls[0] = abi.encodeWithSelector(RouterCommonMock.sendExtraEth.selector, alice, extraAmount);
+        calls[1] = abi.encodeWithSelector(RouterCommonMock.manualReturnETH.selector);
+        calls[2] = abi.encodeWithSelector(RouterCommonMock.assertETHBalance.selector, balanceBefore - extraAmount);
```
jubeira, line 175: Can we add a comment here explaining very briefly each step for reference?

pkg/vault/contracts/RouterCommon.sol
```diff
@@ -191,8 +213,11 @@ abstract contract RouterCommon is IRouterCommon, VaultGuard {
      * returned ETH.
      */
     function _returnEth(address sender) internal {
-        uint256 excess = address(this).balance;
+        if (_isReturnEthLockedSlot().tload()) {
```
jubeira, line 216: Let's add a small comment explaining why we need this.
Something along the line `If the slot is locked, this is part of a larger operation and ETH needs to be returned at the very end.`

jubeira (approved):
Nice @elshan-eth; that was very smart. I like the pattern.

We can add the remaining payable modifiers in another PR.

joaobrunoah (approved):
LGTM!

PR #991, opened 2024-09-18
Stricter trade size guardrail for swaps

### LINKED ISSUE #990: Additional guardrail for swaps
Per auditor comments, the "minimum trade amount" guardrail ensures valid amounts when adding and removing liquidity. Since in single token cases zero is a valid amount, this check allows zero amounts (relying on other checks to ensure we don't have all zero or multiple non-zero amounts).

We're applying the same check on swaps, when it's less clear than zero amounts are valid. A zero-for-zero swap, or even a non-zero-for-zero (i.e., donation), wouldn't hurt anything - especially since it's always possible to "donate," by repeatedly doing reciprocal swaps and paying fees.

What's not valid is a zero (or small) amount in, and a non-zero amount out. This would drain value, and is most likely an attack. Change the guardrail for swaps to check once at the end (once the calculated amount is known), to ensure that if either swap amount is below the minimum, the amount out is zero. We want to allow for donation, but explicitly block small swaps that extract value.

EndymionJkb:
# Description

Per the associated issue, this changes the guardrail for swaps to check once at the end (after the calculated amount is known), to ensure that if either swap amount is below the minimum, the amount out must be zero. We want to allow for donation, but explicitly block small swaps that extract value.

The previous checked the given and calculated amounts independently, allowing zero, which allows the case where the amountIn is 0, and the amountOut is non-zero (and > minimum). This would be a pathological case that would be hard to achieve, but there's no reason to allow it.

## Type of change

- [ ] Bug fix 
- [ ] New feature 
- [ ] Breaking change 
- [ ] Dependency changes
- [X] Code refactor / cleanup
- [ ] Other

## Checklist:

## Issue Resolution

Closes #990


joaobrunoah (dismissed):
LGTM, just a small comment about the test.

pkg/vault/test/foundry/unit/VaultUnit.t.sol
```diff
@@ -249,4 +252,57 @@ contract VaultUnitTest is BaseTest {
...
+
+        vm.expectRevert(IVaultErrors.TradeAmountTooSmall.selector);
+        vault.ensureValidTradeAmount(tradeAmount);
+
+        // Should succeed when it's the minimum
+        vault.ensureValidTradeAmount(vault.getMinimumTradeAmount());
+    }
+
+    // Disallow too small in, allow zero out
+    function testMinimumSwapAmounts() public {
```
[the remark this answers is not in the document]
EndymionJkb: Will separate!

pkg/vault/contracts/Vault.sol
```diff
@@ -1475,14 +1473,28 @@ contract Vault is IVaultMain, VaultCommon, Proxy {
...
+    // donation), but small trades that extract value are explicitly blocked (e.g., zero in, non-zero out).
+    function _ensureValidSwapAmounts(uint256 amountGiven, uint256 amountCalculated, SwapKind kind) internal view {
+        (uint256 amountIn, uint256 amountOut) = kind == SwapKind.EXACT_IN
+            ? (amountGiven, amountCalculated)
+            : (amountCalculated, amountGiven);
+
+        if ((amountIn < _MINIMUM_TRADE_AMOUNT || amountOut < _MINIMUM_TRADE_AMOUNT) && amountOut != 0) {
+            revert TradeAmountTooSmall();
+        }
+    }
```
jubeira, line 1496: Do we want donations through swaps for small amounts though?
If a pool wants donations, devs can enable it already at registration time. Donating dust on the other hand doesn't seem a valid use case, ever. You can donate through swaps by swapping back and forth many times larger amounts anyways; there's no way to prevent that because that's how pools are supposed to work.

I'd just simplify and say 

```suggestion
    function _ensureValidSwapAmounts(uint256 amountGiven, uint256 amountCalculated) internal view {
        if (amountIn < _MINIMUM_TRADE_AMOUNT || amountOut < _MINIMUM_TRADE_AMOUNT) {
            revert TradeAmountTooSmall();
        }
    }
```

jubeira (commented):
Thanks @EndymionJkb !

I think we were using `_ensureValidTradeAmount` just for convenience, but the point described in the ticket is valid.
I'm not against the change, but I'd keep it even simpler to read and even a bit more restrictive. 

Also saves one `if kind == EXACT_IN` which we ask too many times in the critical path already :)

jubeira (approved):
Other than that LGTM!

pkg/vault/contracts/Vault.sol
EndymionJkb: Ok; well, this is closer to what it was before, just more restrictive (including the main purpose, preventing 0 in -> non-zero out. This just prohibits all dust trades, but also disallows the 100% fee cases where we legitimately get 0 out.

I have a version that explicitly allows 0 token out, doesn't break any tests, and isn't that bad (saves a little bytecode and gas vs. this version).

pkg/vault/contracts/Vault.sol
```diff
@@ -258,7 +260,10 @@ contract Vault is IVaultMain, VaultCommon, Proxy {
             poolSwapParams
         );
 
-        _ensureValidTradeAmount(amountCalculatedScaled18);
+        // Prohibit "dust trades" generally, but explicitly allow 0 tokens out.
+        if (vaultSwapParams.kind == SwapKind.EXACT_IN && amountCalculatedScaled18 != 0) {
+            _ensureValidSwapAmount(amountCalculatedScaled18);
+        }
```
jubeira, line 266: Hm but if this is exact out, amount in can be dust?

jubeira (commented):
I'm not 100% sold tbh. Let's discuss it a bit more tomorrow; shouldn't be a big change anyways.

EndymionJkb:
You can donate (including dust) as long as there's zero going out.

We could do your original version (block all dust), but we're relying on
the router to revert on zero in, or else we could theoretically have zero
in and non-zero out

EndymionJkb:
(On phone, so can't edit.) I mean the original allowed zero in (relying on
the router). We could block all dust, including zero out, but it breaks a
test (100% fee) that has zero out.

We could block all dust as you suggest, exclude the zero out case, but then
we'd need to also drop the max fee (which we were contemplating doing
anyway), so maybe it's not that bad. If we think zero out shouldn't be
supported, we could block it like you suggest, which would also create an
effective max fee that could be calculated

pkg/vault/contracts/Vault.sol
```diff
@@ -376,6 +372,8 @@ contract Vault is IVaultMain, VaultCommon, Proxy {
             poolSwapParams.amountGivenScaled18 -= locals.totalSwapFeeAmountScaled18;
         }
 
+        _ensureValidSwapAmount(poolSwapParams.amountGivenScaled18);
```
jubeira, line 375: Should we apply this check before the tax? 

The dynamic swap fee of 100% is ill-defined anyways; you can execute it exact-in like this, but exact-out will end up dividing by 0 [here](https://github.com/balancer/balancer-v3-monorepo/pull/991/files#diff-f03125e354a92ba020d9db1ee551c1b9ba5034e99ae369a139e02e8679002709R409).

EndymionJkb: I had it that way in the beginning - but we're doing amountOut after adjustment, so I'm not sure. The "after tax" amount is what's going into the pool math (i.e., a high fee could take a small amount given and make it near zero, which we're trying to avoid).

Similarly, maybe we should do amountOut before adjustment (right after the pool math). So we'd just bracket the pool math:

_ensureValidSwapAmount(poolSwapParams.amountGivenScaled18);

amountCalculatedScaled18 = IBasePool(vaultSwapParams.pool).onSwap(poolSwapParams);

_ensureValidSwapAmount(amountCalculatedScaled18);

jubeira (approved):
I think this sounds reasonable.
In principle I don't see this breaking reasonable use cases. And as you point out it's the pool math that needs to be protected, so the new place where we're applying the restriction makes more sense than before.

/cc @danielmkm

PR #996, opened 2024-09-19
Remove swap fee amount raw from swap event

jubeira:
# Description

On top of #983.
We're not emitting raw swap fee amounts for add / remove, so it's inconsistent to do it just for swaps. Moreover, there might be after swap fees with hooks, etc.

Swap fees can always be queried onchain as they are stored in the vault, or via events when they are collected.

## Type of change

- [ ] Bug fix 
- [ ] New feature 
- [ ] Breaking change 
- [ ] Dependency changes
- [x] Code refactor / cleanup
- [ ] Other

## Checklist:

## Issue Resolution

N/A


pkg/vault/contracts/Vault.sol
```diff
@@ -344,7 +344,6 @@ contract Vault is IVaultMain, VaultCommon, Proxy {
      */
     struct SwapInternalLocals {
         uint256 totalSwapFeeAmountScaled18;
-        uint256 totalSwapFeeAmountRaw;
```
EndymionJkb, line 347: I tried it - we can get rid of this struct now without blowing up the stack. Can define `totalSwapFeeAmountScaled18` on top, and `aggregateFeeAmountRaw` down where it's first used.

jubeira:
We'll need the raw swap fee amounts in the end.

PR #1000, opened 2024-09-22
Disable BPT getRate for WeightedPools

### LINKED ISSUE #997: Address Weighted Pool rate
Nesting weighted pools with rates (using the BPT as the rate) is not entirely safe (especially with very unbalanced weights and high fees). We already round correctly with invariants, and handle rounding rates with a maximum of 1 wei error. However, weighted pool rates have ~10k wei error, because of the weighted math.

We've considered various methods of addressing this:

1) Modifying IRateProvider to take a rounding direction / order - this was considered too disruptive (would require rewriting rate providers)
2) Preventing the rate from going down within a block, by caching and reverting - this would stop all possible "rate" attacks, even ones we haven't thought of, but would require transient storage and extra logic
3) Replacing the WeightedMath library with a pow function with less error (looks like there might not be one though)
4) Blocking it explicitly for our Weighted Pools (i.e., revert on getRate; don't expose in dynamic data), noting that it would affect any pool type with error in the invariant/rate > 1 wei.

The consensus is to do number 4 (plus documentation of the issue). And maybe also number 3 if it works out. We don't see a valid use case for doing this, so it's not worth adding complexity to the Vault to avoid it.

EndymionJkb:
# Description

The BPT rate of Weighted Pools is dependent on the invariant, which has a potentially large (and non-linear) error that increases with the number of tokens. Short of adding a rounding direction to the rate provider interface (similar to the invariant rounding direction parameter) - which we decided would be too disruptive -- the next best thing is to simply have `getRate` revert for WeightedPools, so that they cannot be used in nested pools as WITH_RATE tokens, using themselves as the rate provider, as this is manipulable and unsafe.

## Type of change

- [ ] Bug fix 
- [ ] New feature 
- [X] Breaking change 
- [ ] Dependency changes
- [ ] Code refactor / cleanup
- [ ] Other

## Checklist:

## Issue Resolution

Closes #997


jubeira (approved):
LGTM!

PR #1001, opened 2024-09-22
Add LBPool

gerrrg:
# Description
tl;dr: Create a weight shifting LBPool for V3.

### Note
This PR replaces and obsoletes https://github.com/balancer/balancer-v3-monorepo/pull/971 by turning it into a PR from the add-lbpool branch from w/in this repo rather than my fork.

## What
* Port over the main functionality of LBPool from V2 to V3
* Port over the gradual value change library
* ~~Create a WeightValidation library for improved portability (could use this in standard WP if desired)~~
* Adhere to events for V2 LBP to improve the ease of integration/compatibility
* Make the pool contract its own hook contract and do beforeJoin hook for onlyOwner LP
* Store info for only a single weight for higher precision/lower gas since weights must sum to `FP.ONE`
* Add relevant tests
* Modify some inherited tests (`BasePoolTest.sol`) to be more generalizable/override-able

## Why
Modernize a highly popular feature from v2

## Type of change

- [ ] Bug fix 
- [x] New feature 
- [ ] Breaking change 
- [ ] Dependency changes
- [ ] Code refactor / cleanup
- [ ] Other

## Checklist:

- [tbd] Tests have 100% code coverage

## Issue Resolution
n/a


gerrrg:
Copying over summary comment from old PR
---------------------------------------------
I've now incorporated all the changes that @EndymionJkb made in https://github.com/balancer/balancer-v3-monorepo/pull/974 and included them in this PR. I believe we can now close that one; I'll summarize the notes/comments from that review here for convenience to note anything known outstanding.

TODOs from PR https://github.com/balancer/balancer-v3-monorepo/pull/974
- [x] Consider moving LBPool to `pkg/pool-weighted/lbp` or similar
- [x] Consider checking for `swapsEnabled` in a minimally overridden `onSwap` rather than in an `onBeforeSwap` hook to reduce gas costs (using this hook results in a double storage read for balances)
- [ ] Port over v2 LBP tests
- [ ] Consider changing the fn name for `updateWeightsGradually` since it describes the action as "update" rather than "schedule".
  - pro: it's more clear 👓 
  - con: it breaks previous version convention (do we care? 🤷)
- [ ] Consider renaming `NormalizedWeightInvariant` in WeightedPool (inherited by LBPool) to be more descriptive

Other TODOs
- [x] LBPFactory test(s)
- [ ] LBPool test coverage

Misc comments/down-the-line TODOs
- Seedless version
- `{,I}TrustedRouterProvider.sol`

gerrrg:
@EndymionJkb I'm not sure if this is a bug in `lcov`, or if I'm actually doing something wrong here.

For `LBPoolFactory.sol`, `lcov` is claiming that the two SLOC in the constructor (setting `_poolVersion` and `_TRUSTED_ROUTER`) are not covered, which appears to be incorrect. The constructor _has_ to be invoked in order for the rest of the tests to run, and other tests (like `getPoolVersion` and the LBP's trusted router test) would undoubtedly fail if these lines were not run.

Other than that, I think I'm at 100% coverage now -- just a bit puzzled by that result

EndymionJkb:
> @EndymionJkb I'm not sure if this is a bug in `lcov`, or if I'm actually doing something wrong here.
> Other than that, I think I'm at 100% coverage now -- just a bit puzzled by that result

We're aware of issues with coverage. It behaves differently / has holes on both hardhat and forge. Modifiers aren't handled very well, for instance (though that's not what's going on here), which is why they can be run either independently or merged.

It's good to run coverage to make sure you aren't missing anything, but some anomalies are expected, so don't worry about it if it flags something you know actually does run.

pkg/pool-weighted/contracts/lbp/LBPool.sol
```diff
@@ -0,0 +1,294 @@
...
+
+        // The SafeCast ensures `endTime` can't overflow.
+        _startGradualWeightChange(startTime.toUint32(), endTime.toUint32(), _getNormalizedWeights(), endWeights);
+    }
+
+    /// @inheritdoc WeightedPool
+    function onSwap(PoolSwapParams memory request) public view override onlyVault returns (uint256) {
+        if (!_getPoolSwapEnabled()) {
+            revert SwapsDisabled();
+        }
```
jubeira, line 179: Why not just pause the pool instead?

EndymionJkb: Pausing would also prevent the owner from adding liquidity. Often these are deployed with swaps disabled, funded at a later point (to set initial prices), the update is scheduled, and then swaps are enabled to start the sale

pkg/pool-weighted/test/LBPool.test.ts
```diff
@@ -0,0 +1,349 @@
...
+      ).to.be.revertedWithCustomError(pool, 'MinWeight');
+
+      // Valid weight update
+      await expect(pool.connect(bob).updateWeightsGradually(startTime, endTime, [fp(0.01), fp(0.99)])).to.not.be
+        .reverted;
+    });
+
+    it('should always sum weights to 1', async () => {
+      const currentTime = await time.latest();
+      const startTime = currentTime + 60; // Set startTime 60 seconds in the future
```
elshan-eth, line 250: `const startTime = currentTime + MINUTE;`

elshan-eth (approved):

pkg/pool-weighted/contracts/lbp/LBPool.sol
```diff
@@ -0,0 +1,294 @@
...
+import { GradualValueChange } from "../lib/GradualValueChange.sol";
+import { WeightedPool } from "../WeightedPool.sol";
+
+/**
+ * @notice Weighted Pool with mutable weights, designed to support v3 Liquidity Bootstrapping.
+ * @dev Inheriting from WeightedPool is only slightly wasteful (setting 2 immutable weights and `_totalTokens`,
+ * which will not be used later), and it is tremendously helpful for pool validation and any potential future
+ * base contract changes.
+ */
+contract LBPool is WeightedPool, Ownable, BaseHooks {
```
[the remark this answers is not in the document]
gerrrg: I'm totally down with O2S. I had originally used that but remembered that the OG LBPool just had a simple immutable `owner`, so I figured using `Ownable` was splitting the difference. 

Happy to go with O2S.

pkg/pool-weighted/contracts/lbp/LBPool.sol
```diff
@@ -0,0 +1,294 @@
...
+     * @return success True if the hook allowed the registration, false otherwise
+     */
+    function onRegister(
+        address,
+        address pool,
+        TokenConfig[] memory,
+        LiquidityManagement calldata
+    ) public view override onlyVault returns (bool) {
+        // Since in this case the pool is the hook, we don't need to check anything else.
+        // We *could* check that it's two tokens, but better to let that be caught later, as it will fail with a more
```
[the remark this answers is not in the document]
gerrrg: What specifically are you asking for here? The check here verifies that this contract is both the pool and the hook

pkg/pool-weighted/contracts/lbp/LBPool.sol
```diff
@@ -0,0 +1,294 @@
...
+        AddLiquidityKind,
+        uint256[] memory,
+        uint256,
+        uint256[] memory,
+        bytes memory
+    ) public view override onlyVault returns (bool) {
+        if (router == _TRUSTED_ROUTER) {
+            return IRouterCommon(router).getSender() == owner();
+        }
+        revert RouterNotTrusted();
```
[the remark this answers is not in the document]
gerrrg: I actually recently made this change in a branch, so yes, I completely agree

pkg/pool-weighted/contracts/lbp/LBPool.sol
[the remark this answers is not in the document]
elshan-eth: I mean checking that the pool contains only 2 tokens

pkg/pool-weighted/test/foundry/LBPool.t.sol
```diff
@@ -0,0 +1,699 @@
...
+
+        uint256 newBptAmount = IERC20(pool).balanceOf(bob);
+
+        // should mint correct amount of BPT poolTokens
+        assertApproxEqAbs(newBptAmount - oldBptAmount, bptAmountOut, DELTA, "LP: Wrong bptAmountOut");
+        assertApproxEqAbs(bptAmountOut, expectedAddLiquidityBptAmountOut, DELTA, "Wrong bptAmountOut");
+    }
+
+    // overriding b/c bob has swap fee authority, not governance
+    // TODO: why does this test need to change swap fee anyway?
```
jubeira, line 189: This shouldn't be needed really.

pkg/pool-weighted/contracts/lbp/LBPool.sol
```diff
@@ -0,0 +1,295 @@
...
+    }
+
+    /**
+     * @notice Check that the caller who initiated the add liquidity operation is the owner.
+     * @dev We first ensure the caller is the standard router, so that we know we can trust the value it returns
+     * from `getSender`.
+     *
+     * @param router The address (usually a router contract) that initiated the add liquidity operation
+     */
+    function onBeforeAddLiquidity(
```
jubeira, line 219: What about initialize?

pkg/pool-weighted/test/foundry/LBPool.t.sol
```diff
@@ -0,0 +1,699 @@
...
+        uint256 bptAmountIn = newBptAmount - oldBptAmount;
+
+        uint256[] memory minAmountsOut = new uint256[](poolTokens.length);
+        for (uint256 i = 0; i < poolTokens.length; ++i) {
+            minAmountsOut[i] = less(tokenAmounts[i], 1e4);
+        }
+
+        // Prevent roundtrip fee
+        vault.manualSetAddLiquidityCalledFlag(pool, false);
+
```
jubeira, line 218:
```suggestion
```

This is no longer needed.

pkg/pool-weighted/test/foundry/LBPool.t.sol
```diff
@@ -0,0 +1,699 @@
...
+        );
+
+        vm.stopPrank();
+
+        (, , uint256[] memory balances, ) = vault.getPoolTokenInfo(address(pool));
+
+        for (uint256 i = 0; i < poolTokens.length; ++i) {
+            // Tokens are transferred to Bob
+            assertApproxEqAbs(
+                poolTokens[i].balanceOf(bob) + TOKEN_AMOUNT, //add TOKEN_AMOUNT to account for init join
```
jubeira, line 234:
```suggestion
                poolTokens[i].balanceOf(bob) + TOKEN_AMOUNT, // add TOKEN_AMOUNT to account for init join
```

pkg/pool-weighted/test/foundry/LBPool.t.sol
```diff
@@ -0,0 +1,699 @@
...
+            tokenConfig1,
+            [uint256(1e18)].toMemoryArray(),
+            DEFAULT_SWAP_FEE,
+            bob,
+            true,
+            ZERO_BYTES32
+        );
+
+        // Attempt to create a pool with 3 tokens
+        // Doesn't throw InputHelpers.InputLengthMismatch.selector b/c create3 intercepts error
```
jubeira, line 538:
```suggestion
        // Doesn't throw InputHelpers.InputLengthMismatch.selector b/c create2 intercepts error
```

jubeira: I'd consider adding a check at the factory tbh.

Create2 errors are super painful to debug.

pkg/vault/test/foundry/utils/BasePoolTest.sol
```diff
@@ -66,7 +67,7 @@ abstract contract BasePoolTest is BaseVaultTest {
         assertEq(pauseManager, address(0), "Pause manager should be 0");
     }
 
-    function testInitialize() public view {
+    function testInitialize() public view virtual {
```
jubeira, line 70: I'd double check these.

jubeira (commented):
Just a few minor comments, but otherwise looks great and makes a lot of sense.

Thanks @gerrrg ! And happy holidays btw :)
I'll take it from here.

pkg/pool-weighted/contracts/lbp/LBPool.sol
```diff
@@ -0,0 +1,295 @@
...
+    // LBPs are constrained to two tokens.
+    uint256 private constant _NUM_TOKENS = 2;
+
+    // LBPools are deployed with the Balancer standard router address, which we know reliably reports the true
+    // originating account on operations. This is important for liquidity operations, as these are permissioned
+    // operations that can only be performed by the owner of the pool. Without this check, a malicious router
+    // could spoof the address of the owner, allowing anyone to call permissioned functions.
+
+    // solhint-disable-next-line var-name-mixedcase
+    address private immutable _TRUSTED_ROUTER;
```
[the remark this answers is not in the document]
EndymionJkb: The other reason is Daniel said we did want to support at least two from launch: basic and batch. If we have the registry (see #897), we could use that. It would be equivalent to gerrg's original "provider" idea, and the maintenance would be done in the registry, so we wouldn't need to worry about it here.

pkg/pool-weighted/contracts/lbp/LBPoolFactory.sol
```diff
@@ -0,0 +1,96 @@
...
+                _TRUSTED_ROUTER
+            ),
+            salt
+        );
+
+        _registerPoolWithVault(
+            pool,
+            tokens,
+            swapFeePercentage,
+            true, // protocol fee exempt
```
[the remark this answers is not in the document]
EndymionJkb: We couldn't charge protocol fees in the old LBP because we were using the "invariant delta" method that was sensitive to weight changes. In the v3 design, I think it's ok to charge them, especially if this is out in the wild, and not part of a third party partner we have an off-chain agreement with.

There's no argument here for "start out zero and raise them later," as with an LBP there's no "later."

pkg/pool-weighted/contracts/lib/GradualValueChange.sol
```diff
@@ -0,0 +1,97 @@
...
+    }
+
+    function resolveStartTime(uint256 startTime, uint256 endTime) internal view returns (uint256 resolvedStartTime) {
+        // If the start time is in the past, "fast forward" to start now
+        // This avoids discontinuities in the value curve. Otherwise, if you set the start/end times with
+        // only 10% of the period in the future, the value would immediately jump 90%
+        resolvedStartTime = Math.max(block.timestamp, startTime);
+
+        if (resolvedStartTime > endTime) {
+            revert GradualUpdateTimeTravel();
```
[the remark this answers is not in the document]
EndymionJkb: Doesn't hurt; it should be a very rare thing, but we don't care about bytecode or gas here. v2 of course only had a "require," so no opportunity to parameterize it.

pkg/pool-weighted/test/foundry/LBPool.t.sol
EndymionJkb: Agreed. And wouldn't use the "b/c" abbreviation (or ".selector" on the error), as non-native speakers might not know what that means.

pkg/pool-weighted/contracts/lbp/LBPool.sol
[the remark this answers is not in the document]
EndymionJkb: It would look like:

```solidity
IBalancerContractRegistry private immutable _contractRegistry;

constructor(
    NewPoolParams memory params,
    IVault vault,
    address owner,
    bool swapEnabledOnStart,
    IBalancerContractRegistry contractRegistry
) WeightedPool(params, vault) Ownable(owner) {
    _contractRegistry = contractRegistry;
}

onBefore hooks:
{
    if (_contractRegistry.isActiveBalancerContract(ContractType.ROUTER, router) == false) {
        revert RouterNotTrusted();
    }

    return IRouterCommon(router).getSender() == owner();
}
```

EndymionJkb (commented):
Agree with the suggestions; ideally would review and merge the registry, then update this one to use it as a source of trusted routers (or we can add a follow-on PR).

pkg/pool-weighted/contracts/lbp/LBPool.sol
jubeira: Well initialize doesn't have the router argument, so we can't do the same approach.

But we can create and initialize in the same tx. I'll add a method in the factory.

pkg/pool-weighted/contracts/lbp/LBPool.sol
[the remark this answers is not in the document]
EndymionJkb: Right... so really the only benefit would be not having to redeploy if we update the basic router (which should be rare compared to adding a new router or updating something more relay-like such as the CLR router)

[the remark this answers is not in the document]
gerrrg: @jubeira good point that the router only affects liquidity provision. Definitely makes it _less_ of an issue, though I do still think it would be best to allow any "approved" router

For more context, the singular trusted router setup that is in the contract was 100% meant to be a placeholder until there was a router provider/contact registry. Simplifying it to be a single router made it possible for me to get unblocked in the meantime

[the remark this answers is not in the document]
jubeira: thanks, I'll double check.
I think this is a good starting point anyways.

pkg/vault/test/foundry/utils/BasePoolTest.sol
jubeira: Alright it's fine.

pkg/pool-weighted/contracts/lbp/LBPool.sol
```diff
@@ -0,0 +1,299 @@
...
+        _TRUSTED_ROUTER = trustedRouter;
+
+        // solhint-disable-next-line not-rely-on-time
+        uint32 currentTime = block.timestamp.toUint32();
+        _startGradualWeightChange(currentTime, currentTime, params.normalizedWeights, params.normalizedWeights);
+        _setSwapEnabled(swapEnabledOnStart);
+    }
+
+    /// @notice Returns trusted router, which is the gateway to add liquidity to the pool.
+    function getTrustedRouter() external view returns (address) {
```
EndymionJkb, line 104: We have this in two places (pool and factory). We could also just store the factory, and call `getTrustedRouter` on the factory. Though then we'd need a getter for the factory anyway, so I guess we're not saving anything (and would need an interface to call anything on the factory).

So I'd say it's fine as is unless we have another reason to keep the factory address. Maybe for the "factory as router" initialization?

Also there aren't tests for the getter in either the pool or the factory.

pkg/pool-weighted/contracts/lbp/LBPool.sol
```diff
@@ -0,0 +1,299 @@
...
+    // LBPs are constrained to two tokens.
+    uint256 private constant _NUM_TOKENS = 2;
+
+    // LBPools are deployed with the Balancer standard router address, which we know reliably reports the true
+    // originating account on operations. This is important for liquidity operations, as these are permissioned
+    // operations that can only be performed by the owner of the pool. Without this check, a malicious router
+    // could spoof the address of the owner, allowing anyone to call permissioned functions.
+
+    // solhint-disable-next-line var-name-mixedcase
+    address private immutable _TRUSTED_ROUTER;
```
EndymionJkb, line 52:
```suggestion
    address private immutable _trustedRouter;
```
Consider making this the same as in the factory (and consistent with the rest of the codebase, since it's not an actual constant).

EndymionJkb (commented):
Looks better now; seems the initialization method is still pending? I suppose we could do both: keep the factory method, and also add the onBeforeInitialize hook and check `_trustedRouter.getSender()` in it.

pkg/pool-weighted/contracts/lbp/LBPool.sol
```diff
@@ -0,0 +1,299 @@
...
+        bool swapEnabled;
+    }
+
+    // LBPs are constrained to two tokens.
+    uint256 private constant _NUM_TOKENS = 2;
+
+    // LBPools are deployed with the Balancer standard router address, which we know reliably reports the true
+    // originating account on operations. This is important for liquidity operations, as these are permissioned
+    // operations that can only be performed by the owner of the pool. Without this check, a malicious router
+    // could spoof the address of the owner, allowing anyone to call permissioned functions.
```
EndymionJkb, line 49:
```suggestion
    // could spoof the address of the owner, allowing anyone to call permissioned functions.
    //
    // Since the initialization mechanism does not allow verification of the router, it is technically possible
    // to front-run `initialize`. This should not be a concern in the typical LBP use case of a new token launch,
    // where there is no existing liquidity. In the unlikely event it is a concern, `LBPoolFactory` provides the
    // `createAndInitialize` function, which does both operations in a single step.
```
Consider adding additional explanation here based on our offline discussion here, in addition to the documentation in the factory itself.

pkg/pool-weighted/contracts/lbp/LBPoolFactory.sol
```diff
@@ -0,0 +1,197 @@
...
+        return
+            _create(name, symbol, tokenConfig, normalizedWeights, swapFeePercentage, owner, swapEnabledOnStart, salt);
+    }
+
+    /**
+     * @notice Deploys a new `LBPool` and seeds it with initial liquidity in the same tx.
+     * @dev Tokens must be sorted for pool registration.
+     * Use this method in case pool initialization frontrunning is an issue.
+     * If the owner is the only address with liquidity of one of the tokens, this should not be necessary.
+     * This method does not support native ETH management; WETH needs to be used instead.
```
EndymionJkb, line 101:
```suggestion
     * This method does not support native ETH management; WETH needs to be used instead.

```
Blank line after multi-line dev block

EndymionJkb (approved):
Looks good! I'm sure we'll continue to iterate on this in follow-ons, but I think it's a fine starting point.

Scutua (approved):

PR #1009, opened 2024-09-25
Add function to revoke permit on BPT

### LINKED ISSUE #1007: Add ability to revoke BPT permit
Re: TOB-BALV3-19

"The BalancerPoolToken contract does not implement a way for permit signatures to be revoked, resulting in these signatures being valid until the deadline."

EndymionJkb:
# Description

Per ToB issue:
"The BalancerPoolToken contract implements the permit function, allowing users to sign a payload in order to allow another user or contract to spend their tokens. However, once a user signs a payload and sends it to another user, they have no way of revoking this signature. Since only one nonce is valid until it is consumed, this would prevent the original user from submitting permit signatures with a different nonce."

This adds a `revokePermit` function that increments the nonce.

## Type of change

- [ ] Bug fix 
- [X] New feature 
- [ ] Breaking change 
- [ ] Dependency changes
- [ ] Code refactor / cleanup
- [ ] Other

## Checklist:

## Issue Resolution

Resolves #1007


jubeira (approved):
LGTM! The fix is correct; I'd be a bit more specific with the comments, otherwise it gives the impression that it can revoke permits that were executed already which is not the case. 

Thanks @EndymionJkb !

PR #1010, opened 2024-09-25
Roundrip fee on proportional remove

jubeira:
# Description

Proposal: tax proportional exits whenever liquidity was added in the same transaction.
Every other remove kind charges fees already as it's unbalanced, and there's no way to remove with more than one token close to the existing proportion to avoid fees (unlike add liquidity unbalanced).

This should make roundrips more difficult, and make swap sandwiches economically unfeasible.
For now I'll keep the liquidity approximation tests as they are, effectively clearing up the flag so as not to apply the tax when calling `removeLiquidityProportional`. In practice, the fee will make indirect swap paths worse than our current fuzz tests, and that's a desirable outcome.

## Type of change

- [ ] Bug fix 
- [x] New feature 
- [ ] Breaking change 
- [ ] Dependency changes
- [ ] Code refactor / cleanup
- [ ] Other

## Checklist:

## Issue Resolution

N/A


pkg/vault/contracts/Vault.sol
```diff
@@ -872,6 +873,13 @@ contract Vault is IVaultMain, VaultCommon, Proxy {
                 _totalSupply(params.pool),
                 bptAmountIn
             );
+
+            if (_addLiquidityCalled().tGet(params.pool)) {
```
jubeira, line 877: Note: if we clear this flag here, it's easy to bypass.

You add a huge amount, make a decoy proportional exit, and then a big exit.
Whenever the flag is set, it'll be set for the rest of the tx. A more correct approach would be to clean it when `transient` modifier ends, but we would need a transient set for that and I really don't think it's worth the added complexity.

pkg/interfaces/contracts/vault/IVaultExtension.sol
```diff
@@ -67,6 +67,12 @@ interface IVaultExtension {
      */
     function getReservesOf(IERC20 token) external view returns (uint256);
 
+    /**
+     * @notice Returns `true` if `addLiquidity` was called for the given pool in this transaction.
+     * @param pool Address of the pool to check.
```
EndymionJkb, line 72:
```suggestion
     * @notice This flag is used to detect and tax "round trip" transactions (adding/removing liquidity in the same pool)
     * @param pool Address of the pool to check
     * @return liquidityAdded True if liquidity has been added to this pool in the current transaction
```

pkg/vault/contracts/Vault.sol
EndymionJkb: Agree that's probably not worth it for an edge case. Although... maybe aggregators might bundle lots of transactions and come across that, since it just checks the pool, not the pool + user, so this could also get triggered on bundled transactions with multiple unrelated users doing liquidity operations on a popular pool. Maybe worth investigating if that's possible? If one person added and another person removed, the second one would get unexpectedly taxed.

Though even storing the user wouldn't be foolproof, as someone could craft a transaction where the add and remove steps were done from separate addresses controlled by the same account.

pkg/interfaces/contracts/vault/IVaultExtension.sol
EndymionJkb: Could also add a dev section here with more detail on why this is desirable / improves security, and documenting limitations.

EndymionJkb (approved):
Looks correct; one more slice of cheese. Just wonder if it might cause issues with aggregators (see comment).

pkg/vault/contracts/Vault.sol
jubeira: Double checked with @danielmkm and this should not be a problem in general.

PR #1008, opened 2024-09-25
Main snaps out-of-date

EndymionJkb:
# Description

This is causing large diffs on all new PRs. Just update the gas to current main code.

## Type of change

- [ ] Bug fix 
- [ ] New feature 
- [ ] Breaking change 
- [ ] Dependency changes
- [ ] Code refactor / cleanup
- [ ] Optimization: [ ] gas / [ ] bytecode
- [ ] Documentation or wording changes
- [X] Other

## Checklist:

- [X] The diff is legible and has no extraneous changes
- [N/A] Complex code has been commented, including external interfaces
- [N/A] Tests have 100% code coverage
- [X] The base branch is either `main`, or there's a description of how to merge

PR #1012, opened 2024-09-25
Payable Router operations

### LINKED ISSUE #1011: Make remaining Router operations payable
Per TOB-BALV3-18, while `permitBatchAndCall` is payable, not *all* Router operations are payable, and mixing payable and non-payable operations will revert. For instance, an "add liquidity" operation followed by a "remove custom liquidity" would currently fail. Recovery mode would also revert in a batch, if combined with other payable operations.

EndymionJkb:
# Description

Make the remaining non-query Router functions payable (custom remove and recovery withdrawal). Probably people wouldn't be doing recovery mode withdrawals in a batch, but I don't see a reason to prevent it.

Not checking and reverting with value saves a small amount of bytecode.

The diff would be a lot smaller if we merged #1008.
## Type of change

- [X] Bug fix 
- [ ] New feature 
- [ ] Breaking change 
- [ ] Dependency changes
- [ ] Code refactor / cleanup
- [ ] Other

## Checklist:

## Issue Resolution

Resolves #1011


elshan-eth (approved):
LGTM!

pkg/vault/contracts/Router.sol
```diff
@@ -449,7 +449,7 @@ contract Router is IRouter, RouterCommon, ReentrancyGuardTransient {
     function removeLiquidityRecovery(
         address pool,
         uint256 exactBptAmountIn
-    ) external returns (uint256[] memory amountsOut) {
+    ) external payable returns (uint256[] memory amountsOut) {
```
jubeira, line 452: Not a problem specific to this PR, but if I just call this function with ETH it will get stuck in the router until somebody else calls a function that returns it. Doesn't sound like the best design idea.

EndymionJkb: Calling a remove with ETH by itself isn't something that should ever happen - but this effect is certainly unintended. I don't think it hurts anything to add it.

jubeira (approved):
Fair enough; LGTM!

Viva Las Vegas

PR #1014, opened 2024-09-26
Avoid zero value transfers

### LINKED ISSUE #993: Handle tokens that prevent zero transfer
Per auditor comment (and we know this from v1/v2), some tokens can't handle zero value transfers. While most of the time this is checked (often as a gas optimization), there are places where it's not. We don't need to do it for WETH or BPT, but any time there is an unknown "pool token," we can ensure transfers are non-zero.

Hopefully it will be mostly a wash as far as gas. The checks take some time, but we're also avoiding zero transfers.

EndymionJkb:
# Description

Per auditor comment, support tokens that prohibit zero transfers (and save a little gas, at the cost of a bit of bytecode), by checking for 0 amounts and refraining from token or ETH transfers of zero value.

## Type of change

- [X] Bug fix 
- [ ] New feature 
- [ ] Breaking change 
- [ ] Dependency changes
- [ ] Code refactor / cleanup
- [ ] Other

## Checklist:

## Issue Resolution

Resolves #993.


pkg/vault/contracts/Router.sol
```diff
@@ -97,7 +97,7 @@ contract Router is IRouter, RouterCommon, ReentrancyGuardTransient {
                 // Transfer WETH from the Router to the Vault.
                 _weth.transfer(address(_vault), amountIn);
                 _vault.settle(_weth, amountIn);
-            } else {
+            } else if (amountIn > 0) {
```
jubeira, line 100: how about `if (amountIn == 0) break;` above instead? Would also apply to eth wraps.

pkg/vault/contracts/Router.sol
```diff
@@ -298,7 +298,7 @@ contract Router is IRouter, RouterCommon, ReentrancyGuardTransient {
                 _weth.deposit{ value: amountIn }();
                 _weth.transfer(address(_vault), amountIn);
                 _vault.settle(_weth, amountIn);
-            } else {
+            } else if (amountIn > 0) {
```
jubeira, line 301: (same as above)

pkg/vault/contracts/RouterCommon.sol
```diff
@@ -263,12 +263,16 @@ abstract contract RouterCommon is IRouterCommon, VaultGuard {
     function _sendTokenOut(address sender, IERC20 tokenOut, uint256 amountOut, bool wethIsEth) internal {
         // If the tokenOut is ETH, then unwrap `amountOut` into ETH.
         if (wethIsEth && tokenOut == _weth) {
-            // Receive the WETH amountOut.
+            // Receive the WETH amountOut. We want to call `sendTo` regardless for reentrancy protection;
+            // it has an internal check for zero amounts.
```
jubeira, line 267: Hmm could you explain why this is an issue? A zero value `sendTo` has no effect anyways, as it shouldn't modify the state at all.

jubeira: maybe just return early if `amountOut` is 0?

pkg/vault/contracts/Vault.sol
```diff
@@ -140,10 +140,12 @@ contract Vault is IVaultMain, VaultCommon, Proxy {
 
     /// @inheritdoc IVaultMain
     function sendTo(IERC20 token, address to, uint256 amount) external nonReentrant onlyWhenUnlocked {
-        _takeDebt(token, amount);
-        _reservesOf[token] -= amount;
+        if (amount > 0) {
```
jubeira, line 143: I'd remove this from here and prefer avoiding the calls from the router whenever possible. Just to keep it extra simple.

`_takeDebt` already does nothing for `amount == 0` anyways.

jubeira (commented):
thanks @EndymionJkb, looks good!
I have a few comments and minor suggestions; I think I'd just modify the routers and keep the vault as is.

pkg/vault/contracts/RouterCommon.sol
EndymionJkb: `sendTo` does a transfer; if 0, it would revert on tokens that don't allow it.

It breaks a non-reentrancy test, but I can try to avoid that another way I guess.

pkg/vault/contracts/Vault.sol
EndymionJkb: Ok; will try. Having it here reduces the changes to the Router. I agree it's cleaner to have it 100% in the router (though it imposes this burden on future routers), as long as it doesn't exceed the bytecode.

EndymionJkb: I'm not sure this is better. Not checking in `sendTo` adds a lot of code to the Routers - and I can't seem to avoid having to change VaultAdmin anyway.

pkg/vault/contracts/RouterCommon.sol
EndymionJkb: I have to be careful to always call settle, as that updates reserves and handles DoS dust, etc. I can short-circuit other places to avoid calling sendTo.

jubeira:
```
    function settle(IERC20 token, uint256 amountHint) external nonReentrant onlyWhenUnlocked returns (uint256 credit) {
        uint256 reservesBefore = _reservesOf[token];
        uint256 currentReserves = token.balanceOf(address(this));
        _reservesOf[token] = currentReserves;
        credit = currentReserves - reservesBefore;

        // If the given hint is equal or greater to the reserve difference, we just take the actual reserve difference
        // as the paid amount; the actual balance of the tokens in the vault is what matters here.
        if (credit > amountHint) {
            // If the difference in reserves is higher than the amount claimed to be paid by the caller, there was some
            // leftover that had been sent to the vault beforehand, which was not incorporated into the reserves.
            // In that case, we simply discard the leftover by considering the given hint as the amount paid.
            // In turn, this gives the caller credit for the given amount hint, which is what the caller is expecting.
            credit = amountHint;
        }

        _supplyCredit(token, credit);
    }
```

I don't think calling settle with 0 makes any difference. If anything, it will just add the dust to the reserves, which is irrelevant to the current operation.
- If credit is 0, it will do `_supplyCredit(token, 0)`. `_supplyCredit` calls `_accountDelta`, which returns early for a `delta == 0` (i.e. does nothing).
- If credit is > 0, `credit > amountHint --> true`. So the `amountHint` will be assigned to the credit, and then `_supplyCredit(token, 0)` will be called.

In the latter case the reserves are updated, incorporating the dust to the reserves, but that can also be done on the next operation. It shouldn't affect debit / credit settlement nor prevent this operation from finishing.

jubeira: In other words, I think breaking loops early is just fine; we don't need to call `settle` with 0 amounts.

pkg/pool-hooks/contracts/MinimalRouter.sol
```diff
@@ -222,26 +226,26 @@ abstract contract MinimalRouter is RouterCommon, ReentrancyGuardTransient {
...
             IERC20 token = tokens[i];
 
             // There can be only one WETH token in the pool.
             if (params.wethIsEth && address(token) == address(_weth)) {
                 // Send WETH here and unwrap to native ETH.
                 _vault.sendTo(_weth, address(this), amountOut);
                 _weth.withdraw(amountOut);
-                ethAmountOut = amountOut;
+                // Send ETH to receiver.
+                payable(params.receiver).sendValue(amountOut);
```
jubeira, line 244: :+1: good idea.

pkg/vault/contracts/BatchRouter.sol
```diff
@@ -526,8 +528,10 @@ contract BatchRouter is IBatchRouter, BatchRouterCommon, ReentrancyGuardTransien
                         // Refund unused portion of BPT flashloan to the Vault.
                         if (bptAmountIn < stepMaxAmountIn) {
                             uint256 refundAmount = stepMaxAmountIn - bptAmountIn;
-                            stepTokenIn.safeTransfer(address(_vault), refundAmount);
-                            _vault.settle(stepTokenIn, refundAmount);
+                            if (refundAmount > 0) {
```
jubeira, line 531: This is always the case, as we're already checking that `stepMaxAmountIn > bptAmountIn` above. We can remove this `if` safely.

pkg/vault/contracts/test/PoolHooksMock.sol
```diff
@@ -181,26 +181,34 @@ contract PoolHooksMock is BaseHooks, VaultGuard {
...
-            } else {
-                hookAdjustedAmountCalculatedRaw -= hookDiscount;
-                if (shouldSettleDiscount) {
-                    params.tokenIn.transfer(address(_vault), hookDiscount);
-                    _vault.settle(params.tokenIn, hookDiscount);
+            if (hookDiscount > 0) {
+                if (params.kind == SwapKind.EXACT_IN) {
+                    hookAdjustedAmountCalculatedRaw += hookDiscount;
+                    if (shouldSettleDiscount) {
+                        if (hookDiscount > 0) {
```
jubeira, line 199: We've already checked for this in L195

pkg/vault/contracts/test/PoolHooksMock.sol
```diff
@@ -181,26 +181,34 @@ contract PoolHooksMock is BaseHooks, VaultGuard {
...
+                            params.tokenOut.transfer(address(_vault), hookDiscount);
+                            _vault.settle(params.tokenOut, hookDiscount);
+                        }
+                    }
+                } else {
+                    hookAdjustedAmountCalculatedRaw -= hookDiscount;
+                    if (shouldSettleDiscount) {
+                        if (hookDiscount > 0) {
```
jubeira, line 207: Same as above; I think we can remove this one.

jubeira (approved):
I think this looks good; just a few minor comments. Thanks for pushing this @EndymionJkb !

Do you still think it'd make sense to protect `sendTo` against zero value inside the vault? I was initially against it, and in general it's better to avoid the external call from the router by checking outside `sendTo`. I'm leaning towards keeping it like this (i.e. no protection inside `sendTo`, as you should avoid those calls anyways), but I'm open to other opinions.

pkg/vault/contracts/BatchRouter.sol
EndymionJkb: Ooof; I just blindly did some of these without checking whether I needed to...

EndymionJkb:
> Do you still think it'd make sense to protect `sendTo` against zero value inside the vault? I was initially against it, and in general it's better to avoid the external call from the router by checking outside `sendTo`. I'm leaning towards keeping it like this (i.e. no protection inside `sendTo`, as you should avoid those calls anyways), but I'm open to other opinions.

I'd rather not change it all back again at this point. It's probably a decent trade-off to have a little extra router code to avoid pointless external calls. It's an edge case anyway. Ideally we wouldn't have had to change the Vault at all - but there is that one place in VaultAdmin that calls sendTo directly.

PR #1027, opened 2024-10-01
Update events

EndymionJkb:
# Description

Per SDK request, add events for aggregate fee updates, and add the total supply (after the operation) to PoolBalanceChanged. The ProtocolFeeController will still emit its own events whenever a constituent swap or yield fee changes (i.e., protocol or pool creator fee), but now the Vault will also emit an event with the final resulting aggregate percentage, which is what gets applied and segregated from the pool balances in the Vault.

## Type of change

- [ ] Bug fix 
- [X] New feature 
- [ ] Breaking change 
- [ ] Dependency changes
- [ ] Code refactor / cleanup
- [ ] Other

## Checklist:


pkg/vault/contracts/VaultAdmin.sol
```diff
@@ -308,6 +308,8 @@ contract VaultAdmin is IVaultAdmin, VaultCommon, Authentication, VaultGuard {
         onlyProtocolFeeController
     {
         _poolConfigBits[pool] = _poolConfigBits[pool].setAggregateSwapFeePercentage(newAggregateSwapFeePercentage);
+
+        emit AggregateSwapFeePercentageChanged(pool, newAggregateSwapFeePercentage);
```
jubeira, line 312: Not completely related to this PR, but the vault doesn't validate the fee precision (only the protocol fee controller does).

Should we validate it here? If the protocol fee controller doesn't validate it well, the value that gets set in storage won't really match the emitted event. Super minor, but still a possibility I guess?

jubeira (approved):
LGTM, clear and concise!

PR #1033, opened 2024-10-08
Optimize scaling

jubeira:
# Description

We're getting decimal scaling factors as FP numbers, and using FP math to apply them to raw values (and to live values to unscale).

However, these are just numbers with a ton of zeroes, and are always a multiple of FP(1). If we store the decimal scaling factor as a "raw" value (i.e. what it is today divided by FP(1)), we can save a few operations scaling and downscaling by switching FP math with raw math. The result should be exactly the same, as there are no rounding issues in there.

## Type of change

- [ ] Bug fix 
- [ ] New feature 
- [ ] Breaking change 
- [ ] Dependency changes
- [ ] Code refactor / cleanup
- [ ] Other

## Checklist:


pkg/vault/test/foundry/unit/PoolConfigLib.t.sol
```diff
@@ -362,8 +362,8 @@ contract PoolConfigLibTest is Test {
 
         uint256[] memory scalingFactors = config.getDecimalScalingFactors(2);
 
-        assertEq(scalingFactors[0], 1e23, "scalingFactors[0] mismatch");
-        assertEq(scalingFactors[1], 1e38, "scalingFactors[1] mismatch");
+        assertEq(scalingFactors[0], 1e5, "scalingFactors[0] mismatch"); // 1e23 / FP(1)
+        assertEq(scalingFactors[1], 1e20, "scalingFactors[1] mismatch"); // 1e38 / FP(1)
```
joaobrunoah, line 366:
```suggestion
        assertEq(scalingFactors[0], 1e5, "scalingFactors[0] mismatch");
        assertEq(scalingFactors[1], 1e20, "scalingFactors[1] mismatch");
```

These comments don't make sense, because 1e23 and 1e38 do not exist anymore. The value stored in PoolConfigLib is used directly, without adding and subtracting 18.

pkg/vault/test/foundry/unit/Scaling.t.sol
```diff
@@ -0,0 +1,135 @@
+// SPDX-License-Identifier: GPL-3.0-or-later
```
joaobrunoah, line 1: Nice tests!

pkg/vault/test/foundry/unit/VaultUnitLiquidity.t.sol
```diff
@@ -807,7 +807,7 @@ contract VaultUnitLiquidityTest is BaseTest, VaultContractsDeployer {
 
         for (uint256 i = 0; i < tokens.length; i++) {
             poolData.tokens[i] = tokens[i];
-            poolData.decimalScalingFactors[i] = 1e18;
+            poolData.decimalScalingFactors[i] = 1; // 18 decimals
```
joaobrunoah, line 810:
```suggestion
            poolData.decimalScalingFactors[i] = 1; // 18 decimals, so 10^(18-18) = 1
```

I'd explain it a bit better, just to have in the whole code that the scaling factor is 10^(18-tokenDecimals)

pkg/vault/test/foundry/unit/VaultUnitSwap.t.sol
```diff
@@ -41,7 +41,7 @@ contract VaultUnitSwapTest is BaseTest, VaultContractsDeployer {
 
     IERC20[] swapTokens;
     uint256[] initialBalances = [uint256(10e18), 10e18];
-    uint256[] decimalScalingFactors = [uint256(1e18), 1e18];
+    uint256[] decimalScalingFactors = [uint256(1), 1]; // 18 decimals
```
joaobrunoah, line 44:
```suggestion
    uint256[] decimalScalingFactors = [uint256(1), 1]; // 18 decimals, so 10^(18-18) = 1
```

I'd explain it a bit better, just to have in the whole code that the scaling factor is 10^(18-tokenDecimals)

pkg/solidity-utils/test/foundry/utils/BaseTest.sol
```diff
@@ -208,7 +208,7 @@ abstract contract BaseTest is Test, GasSnapshot {
     function getDecimalScalingFactor(uint8 decimals) internal pure returns (uint256 scalingFactor) {
         require(decimals <= 18, "Decimals must be between 0 and 18");
         uint256 decimalDiff = 18 - decimals;
-        scalingFactor = 1e18; // FP1
+        scalingFactor = 1;
```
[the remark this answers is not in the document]
jubeira: Just adjusted the loop below.

PR #1034, opened 2024-10-08
Fix rounding for computeAddLiquiditySingleTokenExactOut and StableMath

elshan-eth:
# Description

In the computeAddLiquiditySingleTokenExactOut function, we are rounding values incorrectly when calculating amountIn. Currently, the calculation looks like this:
`uint256 amountIn = newBalance (ROUND_DOWN) - currentBalances[tokenInIndex] (ROUND_UP);`

This means that we are taking less from the user than we could if we rounded in the opposite direction. A fairer approach for the Vault would be to use the following calculation:
`uint256 amountIn = newBalance (ROUND_UP) - currentBalances[tokenInIndex] (ROUND_DOWN);`

As for the computeRemoveLiquiditySingleTokenExactIn function, the amountOut value should be rounded down. Thus, the correct calculation would be:
uint256 amountOut = currentBalances[tokenOutIndex] (ROUND_DOWN) - newBalance (ROUND_UP);

## Type of change

- [X] Bug fix 
- [ ] New feature 
- [ ] Breaking change 
- [ ] Dependency changes
- [ ] Code refactor / cleanup
- [ ] Other

## Checklist:


pkg/vault/contracts/BasePoolMath.sol
```diff
@@ -238,22 +238,34 @@ library BasePoolMath {
...
 
-        uint256 newBalance = pool.computeBalance(currentBalances, tokenInIndex, invariantRatio);
+        uint256 amountIn = 0;
+        uint256 fee = 0;
+        {
+            // We remove round up for current balance to make `amountIn` go up.
+            uint256[] memory currentBalancesRoundedDown = new uint256[](currentBalances.length);
+            for (uint256 i = 0; i < currentBalances.length; ++i) {
+                currentBalancesRoundedDown[i] = currentBalances[i] - 1;
+            }
```
jubeira, line 248: So, this sounds strictly correct but a bit too much effort for it to be justified. We're allocating a whole new array just to subtract 1 wei which feels like a lot.

For methods that use `computeInvariant` it makes more sense because everything depends on a ratio, and we need a new balance array anyways. In this case all of that is an extra, and there's no known edge case that breaks with the current implementation on `main`.

In other words, while this effectively pushes the rounding implicit fee even more in favor of the vault, it feels a bit too much given all the other rounding that we're applying. Moreover, these methods always charge a fee because they involve single tokens, so the taxable amount should always be positive for valid use cases (if it's not, chances are that `newBalances` will be below current balances and checked math will revert anyways). 

According to the snaps this ends up adding 110 bytes and ~1k gas for add single token exact out. I'd rather save that for another battle.

jubeira (commented):
Thanks @elshan-eth !

From this one, I'd only keep the change in `StablePool`. The changes in `BasePoolMath` are correct, but they are too expensive in terms of complexity / bytecode / gas for very little gain. There's no edge case in sight that has a problem with these methods, and all the existing rounding + minimum swap fees should be enough.

pkg/solidity-utils/test/foundry/StableMath.t.sol
```diff
@@ -228,15 +238,61 @@ contract StableMathTest is Test {
...
 
-        uint256 invariant = stableMathMock.computeInvariant(amp, balances);
+        uint256 invariant = stableMathMock.computeInvariant(amp, balances, Rounding.ROUND_DOWN);
         uint256 balanceOne = stableMathMock.computeBalance(amp, balances, invariant, tokenIndex);
         uint256 balanceTwo = stableMathMock.computeBalance(amp, balances, invariant + invariantDiff, tokenIndex);
 
         assertGe(balanceTwo, balanceOne, "The balance should be greater or eq when the invariant is greater.");
     }
+
+    function testComputeBalanceRounding__Fuzz(
```
elshan-eth, line 255: Nice!

joaobrunoah (approved):
I believe snaps are out of date, but it looks good.

jubeira:
> I believe snaps are out of date, but it looks good.

This only modifies stable pool, so they should be fine.

PR #1035, opened 2024-10-08
Fix computeBalance in stable pool math

elshan-eth:
# Description

The `computeBalance` function with a higher invariant can give a result smaller than that of `computeBalance` with a lower invariant. The issue arises due to precision loss during the computation of `computeBalance`.

## Type of change

- [X] Bug fix 
- [ ] New feature 
- [ ] Breaking change 
- [ ] Dependency changes
- [ ] Code refactor / cleanup
- [ ] Other

## Checklist:

### DIFF
--- a/pkg/solidity-utils/test/foundry/StableMath.t.sol
+++ b/pkg/solidity-utils/test/foundry/StableMath.t.sol
@@ -68,6 +68,10 @@ contract StableMathTest is Test {
         amp = boundAmp(amp);
         uint256[] memory balances = boundBalances(rawBalances);
 
+        try stableMathMock.computeInvariant(amp, balances) returns (uint256) {} catch {
+            vm.assume(false);
+        }
+
         stableMathMock.computeInvariant(amp, balances);
     }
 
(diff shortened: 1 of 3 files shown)


pkg/solidity-utils/test/foundry/StableMath.t.sol
```diff
@@ -68,6 +68,10 @@ contract StableMathTest is Test {
         amp = boundAmp(amp);
         uint256[] memory balances = boundBalances(rawBalances);
 
+        try stableMathMock.computeInvariant(amp, balances) returns (uint256) {} catch {
```
jubeira, line 71: Maybe a bit redundant, but perhaps we can check that the error was that the invariant didn't converge here and elsewhere.

jubeira (approved):
This sounds very reasonable; great finding!

The precision is 1e3, so I don't see overflow risk anyways.
On top of this, should we round up inside StablePool's `computeBalance` when computing the invariant?

joaobrunoah (approved):
LGTM!

PR #1036, opened 2024-10-08
 Adjust stable math & live balances rounding (option 2)

jubeira:
# Description

Same as #1020, but with an approach that only undoes rounding up if the result is not precise.
This results in a lower average implicit fee for add and remove kinds that use `computeInvariant`, but it doesn't break so many tests and still addresses edge cases in stable math rounding.

## Type of change

- [x] Bug fix 
- [ ] New feature 
- [ ] Breaking change 
- [ ] Dependency changes
- [ ] Code refactor / cleanup
- [ ] Other

## Checklist:

## Issue Resolution

N/A


jubeira:
Thinking it further, the only objective advantage of this approach over #1020 is that it breaks less tests.

In practice, live balances will _not_ be multiples of FP(1) almost always, so in practice we'll end up rounding down as in #1020 subtracting 1 wei to every token balance anyways, with some extra operations in the middle. This still breaks some tests, as it's not possible to use perfectly round amounts everywhere (composite liquidity router is a great example). So perhaps it's just better to go with the simpler approach in #1020 and just embrace that unbalanced operations will always have this implicit tax, no matter how round the balances are.

jubeira:
#1020 does the job and is simpler.

PR #1038, opened 2024-10-09
Compiler settings adjustment

jubeira:
# Description

Use 500 optimizer runs in the Vault again. This seems to be the best bang-for-buck setting. Bytecode size increases a bit, but gas costs go down significantly.

I've tested other settings (700, 800, 900, 1000 runs) and the inflection point is at 900 runs (below that number, the result is the same as with 500). With 900, gas goes down by very little (mostly 100, perhaps a few hundred tops), but bytecode size grows 1.6k above the limit, so it's not a good tradeoff and we wouldn't be able to deploy it anyways. 

## Type of change

- [ ] Bug fix 
- [ ] New feature 
- [ ] Breaking change 
- [ ] Dependency changes
- [ ] Code refactor / cleanup
- [x] Optimization: [x] gas / [ ] bytecode
- [ ] Other

## Checklist:

## Issue Resolution

N/A


joaobrunoah (approved):
LGTM!

PR #1043, opened 2024-10-10
add sender for queries and save sender

### LINKED ISSUE #1037: Expose sender in queries
We need to make the sender a parameter for queries, and make `saveSender` take it.

elshan-eth:
# Description

The SDK team asked that we add `sender` as an argument to query methods. This PR corrects the behavior of the `saveSender` modifier.

## Type of change

- [ ] Bug fix 
- [ ] New feature 
- [ ] Breaking change 
- [ ] Dependency changes
- [ ] Code refactor / cleanup
- [ ] Other

## Checklist:

## Issue Resolution

Closes #1037


pkg/interfaces/contracts/vault/IBatchRouter.sol
```diff
@@ -100,27 +100,31 @@ interface IBatchRouter {
      * @notice Queries a swap operation involving multiple paths (steps), specifying exact input token amounts.
      * @dev Min amounts out specified in the paths are ignored.
      * @param paths Swap paths from token in to token out, specifying exact amounts in
+     * @param sender Address of the real account that called the swap function
```
EndymionJkb, line 103: My understanding is we're doing this so that aggregators or other third parties can simulate calls from particular user accounts (e.g., so that user-sensitive hooks will return correct results), in which case the sender would not necessarily be the real caller (as it is with the non-query functions).

So consider something for all these like "Address passed to the operation as the sender; may influence results (e.g., with user-dependent hooks)"

Also, they all say "swap function," even if the actual operation is something else. Addressed above.

EndymionJkb (approved):
Looks consistent - sender argument added to all queries that used `saveSender` (just not recovery mode remove, as that by definition can't be sensitive to the sender), and saved in `saveSender`. And `saveSender(msg.sender)` for all non-queries.

Just have a suggestion for how to document the parameter, but otherwise looks good.

jubeira (approved):
Looks pretty good to me! Thanks @elshan-eth :)

Can we please address the comments so that they don't say `swap` for every operation? Other than that this is good to :ship:

PR #1046, opened 2024-10-12
Add full amplification update data

### LINKED ISSUE #1045: Query amplification update status on-chain
If the "start" event for a stable pool amplification update is missed in the subgraph, there's no way to compute the current amp factor on-chain, if it's changing. Add the start/end date on-chain, possibly in two places (the dynamic data, and maybe also `getAmplificationParameter`).

EndymionJkb:
# Description

The SDK team noted that if the subgraph missed a start or stop amplification event, it would be impossible to calculate the current amp parameter during an update from the data currently available on-chain (i.e., just the `isUpdating` flag and current value).

Since the `getAmplificationParameter` function was inherited from v2, and there may be lots of code using it, this PR leaves that function alone and adds a new function to return the low-level state.

The same information was also added to the Stable Pool dynamic data.

Added tests for Immutable and Dynamic data, which were missing. (For both Weighted and Stable.)

## Type of change

- [ ] Bug fix 
- [X] New feature 
- [ ] Breaking change 
- [ ] Dependency changes
- [ ] Code refactor / cleanup
- [ ] Other

## Checklist:

## Issue Resolution

Resolves #1045


pkg/pool-stable/test/foundry/LiquidityApproximationStable.t.sol
```diff
@@ -160,11 +161,21 @@ contract LiquidityApproximationStableTest is LiquidityApproximationTest, StableP
...
+        (AmplificationState memory ampState, uint256 precision) = IStablePool(pool).getAmplificationState();
+        assertEq(ampState.startTime, block.timestamp, "Wrong amp update start time");
+        assertEq(ampState.endTime, endTime, "Wrong amp update end time");
+        assertEq(ampState.startValue, value, "Wrong amp update start value");
+        assertEq(ampState.endValue, newAmplificationParameter * precision, "Wrong amp update end value");
+
         vm.warp(block.timestamp + updateInterval + 1);
 
-        (uint256 value, bool isUpdating, uint256 precision) = StablePool(pool).getAmplificationParameter();
+        (value, isUpdating, ) = IStablePool(pool).getAmplificationParameter();
```
jubeira, line 178: this is a kinda weird place for this test :sweat_smile: 
I'd test the getter elsewhere, preferably in `StablePool.t.sol`

jubeira (approved):
The change looks good, thanks @EndymionJkb.

Can we move the test for the new function to `StablePool.t.sol`? Also, can we add a test for `getStablePoolDynamicData` now that we are at it?

pkg/pool-stable/test/foundry/LiquidityApproximationStable.t.sol
EndymionJkb: It was just convenient, since this is where we're testing the amp update, and the main stable pool test is pretty empty. I can move it.

joaobrunoah (approved):
LGTM!

PR #1049, opened 2024-10-14
fix coverage flag

elshan-eth:
# Description

The COVERAGE flag is interpreted as a string, so it will always be true. Because of this, when we want to compile with viaIr we will get the size of contracts without viaIr.

## Type of change

- [ ] Bug fix 
- [ ] New feature 
- [ ] Breaking change 
- [ ] Dependency changes
- [ ] Code refactor / cleanup
- [ ] Other

## Checklist:

### DIFF
--- a/pvt/common/hardhat-base-config.ts
+++ b/pvt/common/hardhat-base-config.ts
@@ -19,7 +19,7 @@ type SolcConfig = {
 };
 
 // The coverage report doesn't work well with via-ir flags, so we disable it
-const viaIR = !process.env.COVERAGE;
+const viaIR = !(process.env.COVERAGE === 'true' ? true : false);
 const optimizerSteps =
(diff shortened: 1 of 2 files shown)


pvt/common/hardhat-base-config.ts
```diff
@@ -70,12 +70,12 @@ const contractSettings: ContractSettings = {
...
     runs: 500,
-    viaIR: false,
+    viaIR,
   },
   '@balancer-labs/v3-vault/contracts/test/CompositeLiquidityRouterMock.sol': {
     version: '0.8.26',
     runs: 9999,
-    viaIR: false,
+    viaIR,
   },
```
jubeira, line 79: Let's just delete these two blocks if the router does compile with viaIR

jubeira (approved):
Good catch! Although I think we were still compiling with via IR as the variable is usually `undefined`.

If the composite liquidity router now compiles with via IR it means that the original problem was gone already. Not sure exactly where, but if it works I won't complain :)

PR #1052, opened 2024-10-15
Graceful failure on unsupported decimals

### LINKED ISSUE #1050: Explicit error for high-decimal tokens
Per auditor suggestion, revert with a specific custom error vs. numeric underflow when attempting to register a pool with a token that has more than 18 decimals. (Also improve documentation around both high- and low-decimal tokens.)

EndymionJkb:
# Description

Per auditor comment, there are some tokens out there (e.g., using ray math) that have > 18 decimals. We say these are unsupported, but if you try to register them, it will fail with a numeric underflow, which isn't very informative.

This defines a maximum number of decimals, and fails gracefully if it is exceeded. Also removes a function that we no longer use.

## Type of change

- [ ] Bug fix 
- [X] New feature 
- [ ] Breaking change 
- [ ] Dependency changes
- [ ] Code refactor / cleanup
- [ ] Other

## Checklist:

## Issue Resolution

Resolves #1050

### DIFF
--- a/pkg/vault/contracts/VaultExtension.sol
+++ b/pkg/vault/contracts/VaultExtension.sol
@@ -232,7 +232,15 @@ contract VaultExtension is IVaultExtension, VaultCommon, Proxy {
                 revert InvalidTokenType();
             }
 
-            tokenDecimalDiffs[i] = uint8(18) - IERC20Metadata(address(token)).decimals();
+            uint8 tokenDecimals = IERC20Metadata(address(token)).decimals();
+
+            if (tokenDecimals > _MAX_TOKEN_DECIMALS) {
+                revert InvalidTokenDecimals();
+            } else {
+                unchecked {
+                    tokenDecimalDiffs[i] = _MAX_TOKEN_DECIMALS - tokenDecimals;
+                }
+            }
 
             // Store token and seed the next iteration.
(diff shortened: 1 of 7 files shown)


pkg/vault/contracts/VaultExtension.sol
```diff
@@ -232,7 +232,16 @@ contract VaultExtension is IVaultExtension, VaultCommon, Proxy {
...
             }
 
-            tokenDecimalDiffs[i] = uint8(18) - IERC20Metadata(address(token)).decimals();
+            // Store the token decimal conversion factor as a delta from the maximum supported value.
+            uint8 tokenDecimals = IERC20Metadata(address(token)).decimals();
+
+            if (tokenDecimals > _MAX_TOKEN_DECIMALS) {
+                revert InvalidTokenDecimals();
+            } else {
+                unchecked {
```
jubeira, line 241: probably a bit overkill given that this is not in the critical path :grimacing: but OK

jubeira (approved):
LGTM! Clean and concise :)

pkg/vault/contracts/VaultExtension.sol
EndymionJkb: Saves 30 bytes :)

PR #1054, opened 2024-10-16
Safer swap fee bounds

### LINKED ISSUE #887: Consider limiting pool creator fee to 99%
Instead of 100%, which would make LP fees 0, it might be a good idea to leave some wiggle room.

### LINKED ISSUE #1053: Clearer revert reason for 100% fees
Per auditor comment, the swapFeePercentage could be set to 100% for pools with dynamic fee hooks. In such a scenario, EXACT_OUT swaps will revert from ZeroDivision() as the complement() would be 0.

EndymionJkb:
# Description

We check that the swap fee can never be > 100% (to keep the math well-behaved), but if it is exactly 100%, ExactOut swaps will revert with an arithmetic error, which is not ideal.

It is also confusing that we have something called MAX_FEE_PERCENTAGE that is set to 1e18 - when the actual maximum fee is less than that. Also, it is used interchangeably with 1e18/FixedPoint.ONE, which makes an invalid assumption.

This sets the values that should be 1e18 to FixedPoint.ONE, and sets MAX_FEE_PERCENTAGE to a value less than 1e18. Since we also have precision constraints, it is reasonable to set it to the maximum value under 100% that meets the precision.

The same fix applies to the pool creator fee (i.e., in case all other fees are 0, to keep the LP fees > 0).

## Type of change

- [X] Bug fix 
- [ ] New feature 
- [ ] Breaking change 
- [ ] Dependency changes
- [ ] Code refactor / cleanup
- [ ] Other

## Checklist:

## Issue Resolution

Resolves #1053 
Resolves #887


joaobrunoah (approved):
LGTM!

pkg/vault/contracts/ProtocolFeeController.sol
```diff
@@ -132,15 +135,15 @@ contract ProtocolFeeController is
...
+        if (newYieldFeePercentage > MAX_PROTOCOL_YIELD_FEE_PERCENTAGE) {
             revert ProtocolYieldFeePercentageTooHigh();
         }
         _ensureValidPrecision(newYieldFeePercentage);
         _;
     }
 
     modifier withValidPoolCreatorFee(uint256 newPoolCreatorFeePercentage) {
-        if (newPoolCreatorFeePercentage > FixedPoint.ONE) {
+        if (newPoolCreatorFeePercentage > MAX_CREATOR_FEE_PERCENTAGE) {
```
jubeira, line 146: This is below the max fee percentage, as the absolute max fee should accommodate max pool creator with max protocol fee percentage. Added test below for that specific condition.

PR #1058, opened 2024-10-17
Document effect of charging on calculated amounts in FeeTakingHook

### LINKED ISSUE #1030: Review tax in `FeeTakingHookExample`
We're now charging fees always on exact in, but that's only possible if we modify amounts given before going into the pool math.

In the after swap hook, we can modify token amounts, but we can only alter the calculated end. This means that the tax will be charged in different tokens depending on whether the swap is exact in / out, potentially breaking the equivalence.

Should also round up the fee.

EndymionJkb:
# Description

Document the potential effect of charging the tax on different tokens depending on the direction in the fee taking hook, per the associated issue.

## Type of change

- [ ] Bug fix 
- [ ] New feature 
- [ ] Breaking change 
- [ ] Dependency changes
- [ ] Code refactor / cleanup
- [ ] Other

## Checklist:

## Issue Resolution

Closes #1030


jubeira (approved):
LGTM!

Can we also round up `hookFee` now that we are at it just to illustrate the point that fees should be rounded up?

joaobrunoah (approved):
LGTM!

PR #1059, opened 2024-10-17
Fungibility tests to make sure invariant increases linearly to amount added unbalanced

joaobrunoah:
# Description

We need to make sure that the pool invariant increases linearly when tokens are added, and proportional to the supply that the pool had before. This ensures that the BPT is fungible.

To test this property, we can add liquidity unbalanced with proportional amounts, because this function will call computeInvariant to calculate the amount of BPT tokens to mint to the user.

## Type of change

- [ ] Bug fix 
- [ ] New feature 
- [ ] Breaking change 
- [ ] Dependency changes
- [x] Code refactor / cleanup
- [ ] Other

## Checklist:


pkg/vault/test/foundry/Fungibility.t.sol
```diff
@@ -0,0 +1,33 @@
...
+        for (uint256 i = 0; i < balancesRaw.length; i++) {
+            exactAmountsIn[i] = balancesRaw[i].mulDown(proportion);
+        }
+
+        uint256 totalSupplyBefore = BalancerPoolToken(pool).totalSupply();
+
+        vm.prank(lp);
+        uint256 bptAmountOut = router.addLiquidityUnbalanced(pool, exactAmountsIn, 0, false, bytes(""));
+
+        // Ensure minted BPT is close 0.0001% from what it ideally should be.
```
EndymionJkb, line 30:
```suggestion
        // Ensure minted BPT is within 0.0001% of what it ideally should be.
```

pkg/pool-weighted/test/foundry/FungibilityWeighted.t.sol
```diff
@@ -0,0 +1,73 @@
...
+
+import { WeightedPoolFactory } from "../../contracts/WeightedPoolFactory.sol";
+import { WeightedPool } from "../../contracts/WeightedPool.sol";
+import { WeightedPoolContractsDeployer } from "./utils/WeightedPoolContractsDeployer.sol";
+
+contract FungibilityWeightedTest is WeightedPoolContractsDeployer, FungibilityTest {
+    using ArrayHelpers for *;
+    using CastingHelpers for address[];
+
+    uint256 internal constant DEFAULT_SWAP_FEE = 1e16; // 1%
```
EndymionJkb, line 25: Could move DEFAULT_SWAP_FEE up to the parent Test

EndymionJkb (approved):
Good sanity check. I've also added an explanation in the docs, in the amm with custom invariant section, and will also add a docs PR to add it to the comments in IBasePool, where the function is defined.

pkg/pool-weighted/test/foundry/FungibilityWeighted.t.sol
joaobrunoah: Good point, I'll do it

pkg/pool-weighted/test/foundry/FungibilityWeighted.t.sol
```diff
@@ -0,0 +1,71 @@
...
+                bytes32(poolCreationNonce++)
+            )
+        );
+        vm.label(address(newPool), label);
+
+        // Cannot set the pool creator directly on a standard Balancer weighted pool factory.
+        vault.manualSetPoolCreator(address(newPool), lp);
+
+        ProtocolFeeControllerMock feeController = ProtocolFeeControllerMock(address(vault.getProtocolFeeController()));
+        feeController.manualSetPoolCreator(address(newPool), lp);
```
jubeira, line 67:
```suggestion
```
I think we can skip this

pkg/pool-stable/test/foundry/FungibilityStable.t.sol
```diff
@@ -0,0 +1,57 @@
...
+            )
+        );
+        vm.label(address(newPool), label);
+
+        // Cannot set the pool creator directly on a standard Balancer stable pool factory.
+        vault.manualSetPoolCreator(address(newPool), lp);
+
+        ProtocolFeeControllerMock feeController = ProtocolFeeControllerMock(address(vault.getProtocolFeeController()));
+        feeController.manualSetPoolCreator(address(newPool), lp);
+
```
jubeira, line 54:
```suggestion
```

pkg/vault/test/foundry/Fungibility.t.sol
```diff
@@ -0,0 +1,33 @@
...
+
+        uint256[] memory exactAmountsIn = new uint256[](balancesRaw.length);
+        for (uint256 i = 0; i < balancesRaw.length; i++) {
+            exactAmountsIn[i] = balancesRaw[i].mulDown(proportion);
+        }
+
+        uint256 totalSupplyBefore = BalancerPoolToken(pool).totalSupply();
+
+        vm.prank(lp);
+        uint256 bptAmountOut = router.addLiquidityUnbalanced(pool, exactAmountsIn, 0, false, bytes(""));
```
jubeira, line 28: Let's add a small comment saying that we need to use unbalanced because it's the codepath that actually uses the invariant; proportional will always work.

pkg/pool-weighted/test/foundry/FungibilityWeighted.t.sol
```diff
@@ -0,0 +1,71 @@
...
+contract FungibilityWeightedTest is WeightedPoolContractsDeployer, FungibilityTest {
+    using ArrayHelpers for *;
+    using CastingHelpers for address[];
+
+    uint256 internal poolCreationNonce;
+
+    function setUp() public override {
+        super.setUp();
+
+        vault.manuallySetSwapFee(pool, 0);
```
jubeira, line 30: I'd do this step in the base `FungibilityTest` instead.

pkg/pool-stable/test/foundry/FungibilityStable.t.sol
```diff
@@ -0,0 +1,57 @@
...
+        PoolHooksMock(poolHooksContract).allowFactory(address(factory));
+
+        StablePool newPool = StablePool(
+            factory.create(
+                "Stable Pool",
+                "STABLE",
+                vault.buildTokenConfig(tokens.asIERC20()),
+                DEFAULT_AMP_FACTOR,
+                roleAccounts,
+                swapFeePercentage, // 1% swap fee, but test will override it
```
jubeira, line 40: I think in this case it doesn't :) (see comment above)

It's probably not a big deal because the taxable amount will be minimal anyways, but the test should just use 0 fees.

pkg/pool-weighted/test/foundry/FungibilityWeighted.t.sol
```diff
@@ -0,0 +1,71 @@
...
+
+        // Allow pools created by `factory` to use poolHooksMock hooks.
+        PoolHooksMock(poolHooksContract).allowFactory(address(factory));
+
+        WeightedPool newPool = WeightedPool(
+            factory.create(
+                "50/50 Weighted Pool",
+                "50_50WP",
+                vault.buildTokenConfig(tokens.asIERC20()),
+                [uint256(50e16), uint256(50e16)].toMemoryArray(),
```
jubeira, line 51: Let's use 20 / 80 or 75/25 to make it a bit more spicy.

jubeira (approved):
LGTM, thanks for putting this together!
This makes a lot of sense; just a few minor comments and observations.

Side-comment: I think the approach of comparing BPT amount out directly as in the proposed change is fine. Do you think it's worth comparing it against a proportional add? It should always be a little less, and proportional add should be closer to the perfect proportion, but I'm not sure we've tested that elsewhere directly (although liquidity approximation tests and e2e tests do check that indirectly).

PR #1064, opened 2024-10-18
Add aux event from vault.

johngrantuk:
# Description

In an event based state system (e.g. as used by Paraswap) pool state, etc is tracked using events. It's very convenient to be able to listen for all relevant events from the Vault (which is already a known contract), rather than have to track/add pools.

As an example - For Stable pool amp updates the pool itself emits the event. Introducing a function that allows a registered pool to emit an event via the Vault could be a useful alternative. 

This is an initial PoC to see what people think.

## Type of change

- [ ] Bug fix 
- [X] New feature 
- [ ] Breaking change 
- [ ] Dependency changes
- [ ] Code refactor / cleanup
- [ ] Other

## Checklist:


jubeira (approved):
Cleaned up a bit. LGTM!

danielmkm (approved):
This looks good to me, would just change the name of the first arg in the event from `from` to `pool`

PR #1073, opened 2024-10-22
Adopt Proxy Pattern in the Router

### LINKED ISSUE #1071: Consider proxy pattern for routers
Routers (particularly the main `Router`) have a lot of queries which use up a lot of bytecode.

We don't care about gas here, so it might be a good idea to push them to an extension.

joaobrunoah:
# Description

Adopts Proxy pattern in the Router, so we can move query functions to a "RouterExtension" file.

## Type of change

- [ ] Bug fix 
- [x] New feature 
- [ ] Breaking change 
- [ ] Dependency changes
- [ ] Code refactor / cleanup
- [ ] Other

## Checklist:

## Issue Resolution

Closes #1071


EndymionJkb (commented):
Looks like this is definitely the answer! Made a follow-on to reduce the diff (see #1074), which also has these comments mostly incorporated.

pkg/vault/contracts/Router.sol
```diff
@@ -65,12 +76,7 @@ contract Router is IRouter, RouterCommon, ReentrancyGuardTransient {
             );
     }
 
-    /**
-     * @notice Hook for initialization.
-     * @dev Can only be called by the Vault.
-     * @param params Initialization parameters (see IRouter for struct definition)
-     * @return bptAmountOut BPT amount minted in exchange for the input tokens
-     */
+    /// @inheritdoc IRouterMain
```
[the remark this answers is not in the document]
joaobrunoah: Now that we have router and router extension with a proxy, I had issues with tests when trying to call the hook directly. If we want to use IRouterMock as the type of router, the external functions must be defined in the interface, or we will need to cast Router and RouterExtension in all tests that call the hook directly.

pkg/pool-weighted/test/WeightedPool.test.ts
```diff
@@ -64,7 +66,8 @@ describe('WeightedPool', function () {
 
     const WETH = await deploy('v3-solidity-utils/WETHTestToken');
     permit2 = await deployPermit2();
-    router = await deploy('v3-vault/Router', { args: [vault, WETH, permit2] });
+    routerExtension = await deploy('v3-vault/RouterExtension', { args: [vault, WETH, permit2] });
+    router = await deploy('v3-vault/Router', { args: [vault, WETH, permit2, routerExtension] });
```
jubeira, line 70: I'd create a TS utility to deploy this, same as we did with the vault.

jubeira: Ah I see there is one already; can we use it here?

pkg/pool-weighted/test/gas/.hardhat-snapshots/[WeightedPool - Standard] add liquidity proportional
```diff
@@ -1 +1 @@
-179.3k
\ No newline at end of file
+179.8k
```
Comment (jubeira) [inline pkg/pool-weighted/test/gas/.hardhat-snapshots/[WeightedPool - Standard] add liquidity proportional:1]: :confused: I wasn't expecting this to happen; quite the opposite actually.
We shouldn't be loading the router extension here, and the router has less external methods now. Why is this going up?

pkg/vault/contracts/RouterExtension.sol
```diff
@@ -0,0 +1,414 @@
...
+import "@balancer-labs/v3-interfaces/contracts/vault/VaultTypes.sol";
+
+import {
+    ReentrancyGuardTransient
+} from "@balancer-labs/v3-solidity-utils/contracts/openzeppelin/ReentrancyGuardTransient.sol";
+
+import { Router } from "./Router.sol";
+import { RouterCommon } from "./RouterCommon.sol";
+
+contract RouterExtension is RouterCommon, ReentrancyGuardTransient, IRouterExtension {
```
jubeira, line 20: This one should have a fallback that reverts if it receives ETH.

jubeira (commented):
Thanks @joaobrunoah ! This makes a lot of sense.

The bytecode size went down by a lot which is great. But I wasn't expecting the gas to go up in this case; I was expecting it to go down actually. Do we know why this happens? I'm a bit reluctant to proceed the change if it makes gas worse unless we absolutely need that bytecode.

There's one more thing I'd like to test though. Do queries work if you target the extension instead of the router?

pkg/vault/contracts/RouterExtension.sol
EndymionJkb: Should be the same situation as the VaultAdmin (end of the Proxy line) - receive that reverts unconditionally (no payables means any calls directly to query methods with ETH would fail too), and the same fallback as VaultAdmin

joaobrunoah:
Abandoning this PR since a router proxy increases the gas in some operations. Alternatively, we will separate buffer operations in another router, which should save some bytes in the main router.

PR #907, opened 2024-08-20
Replace encodeWithSelector with encodeCall

EndymionJkb:
# Description

Replace `encodeWithSelector` with the more type-safe/less error-prone `encodeCall`, per auditor suggestion.

## Type of change

- [ ] Bug fix 
- [ ] New feature 
- [ ] Breaking change 
- [ ] Dependency changes
- [X] Code refactor / cleanup
- [ ] Optimization: [ ] gas / [ ] bytecode
- [ ] Documentation or wording changes
- [ ] Other

## Checklist:

- [X] The diff is legible and has no extraneous changes
- [N/A] Complex code has been commented, including external interfaces
- [N/A] Tests have 100% code coverage
- [X] The base branch is either `main`, or there's a description of how to merge

## Issue Resolution

Resolves #904

PR #1077, opened 2024-10-23
Universal encode call

EndymionJkb:
# Description

I already went through and did this once (see [this PR](https://github.com/balancer/balancer-v3-monorepo/pull/907), and two follow-ons), but a couple have slipped back in after the purge.

## Type of change

- [ ] Bug fix 
- [ ] New feature 
- [ ] Breaking change 
- [ ] Dependency changes
- [X] Code refactor / cleanup
- [ ] Other

## Checklist:


joaobrunoah (approved):
Nice to standardize this. So, If I got correctly, we should avoid using encodeWithSelector and use encodeCall for everything, right?

I didn't get why stable pool gas went up, probably main is out of date?

EndymionJkb:
> Nice to standardize this. So, If I got correctly, we should avoid using encodeWithSelector and use encodeCall for everything, right?

Yes, which was the policy (I think it was an early auditor comment?); we just forgot to enforce it on the latest code. Note that you still have to use encodeWithSelector in some places (notably with errors in `expectRevert`, which is the most comment usage, in the RevertCodec, etc.)

> I didn't get why stable pool gas went up, probably main is out of date?
Not sure about the gas; either that or the phase of the moon.

jubeira (approved):
LGTM!

PR #1084, opened 2024-10-28
Adjust aggregate fee precision in edge cases

EndymionJkb:
# Description

Protocol fee percentages are limited to 24-bit precision for performance reasons (i.e., to fit all the fees in a single slot), and because high precision is not needed. Generally we expect protocol fees set by governance to be simple integers.

However, the pool creator fee is entirely controlled by the pool creator, and it is possible to craft a valid pool creator fee percentage that would cause the aggregate fee percentage to fail the precision check. This case should be rare, so we ensure this can't happen by truncating the final value.

## Type of change

- [ ] Bug fix 
- [ ] New feature 
- [ ] Breaking change 
- [ ] Dependency changes
- [X] Code refactor / cleanup
- [ ] Other

## Checklist:


jubeira:
We're not proceeding with this for now.

PR #1085, opened 2024-10-29
Stable Surge hook (v2)

EndymionJkb:
# Description

Draft of a stable surge hook, according to the new algorithm that has only a threshold, and can be used for n-token pools.

The main idea is to "surge" (i.e., increase) the dynamic swap fee of a stable pool according to whether the end state of the trade (computed using the standard static swap fee percentage): 1) leaves the pool more unbalanced than before the swap; and 2) the degree of imbalance is greater than a threshold.

The fee is then increased linearly between the current static fee percentage and a hard-coded maximum surge fee (e.g., 95%).

Since pools can behave very differently, each pool can have its own threshold. There is a default threshold set on deployment of the hook, and all new pools registered with this hook are initially set to this value. A pool's individual threshold can be changed by the swap manager of the pool. (Question: Is this correct? They could just set it to 100%, and their pool would never get surged. Maybe governance can override it? Or the hook could have an owner, but that also seems as dangerous as letting the swap manager set it.)

This is a draft; there are no tests yet.

## Type of change

- [ ] Bug fix 
- [X] New feature 
- [ ] Breaking change 
- [ ] Dependency changes
- [ ] Code refactor / cleanup
- [ ] Other

## Checklist:

### DIFF
--- a/pkg/pool-hooks/contracts/StableSurgeHookExample.sol
+++ b/pkg/pool-hooks/contracts/StableSurgeHookExample.sol
@@ -1,0 +1,303 @@
+// SPDX-License-Identifier: GPL-3.0-or-later
+
+pragma solidity ^0.8.24;
+
+import { IBasePoolFactory } from "@balancer-labs/v3-interfaces/contracts/vault/IBasePoolFactory.sol";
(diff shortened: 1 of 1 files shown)


jubeira:
Superseded by #1101

PR #1086, opened 2024-10-29
Nested pool benchmark

elshan-eth:
# Description

Refactoring for Benchmark + nested pool tests.

## Type of change

- [ ] Bug fix 
- [ ] New feature 
- [ ] Breaking change 
- [ ] Dependency changes
- [X] Code refactor / cleanup
- [ ] Other

## Checklist:


pvt/benchmarks/src/PoolBenchmark.behavior.ts
```diff
@@ -651,105 +850,209 @@ export class Benchmark {
...
+        await permit2.connect(alice).approve(pool, batchRouter, MAX_UINT160, MAX_UINT48);
+
+        await deployAndInitializePool(PoolTag.WithBPT, sortAddresses([await pool.getAddress(), tokenCAddress]), false, [
+          bptBalance,
+          bptBalance,
+        ]);
+      });
+
+      context('swap', () => {
+        itTestsSwap(PoolTag.Nested, {
```
jubeira, line 981: Note that we want something a bit different from this one. I believe what we're measuring here is a swap internal to the parent pool; we also need the measurement end to end.

In other words, suppose we have:
- Pool A: waUsdc / waUsdt / waDai
- Pool B: Pool A / GHO

If I'm not mistaken this test measures a swap of Pool A for GHO right?

We also want to test DAI / GHO.
We must use the batch router for this, and we need initialized buffers (with enough liquidity). The batch router will figure all of this out given the correct swap path.

The swap path should be:
- DAI - waDAI (buffer wrap)
- waDAI - pool A (add liquidity)
- pool A - GHO (swap)

And also vice versa, GHO to DAI:
- GHO - Pool A (swap)
- Pool A - waDAI (exit)
- waDAI - DAI (unwrap)

Again, the batch router should figure everything out as long as the swap path is correctly set.

jubeira (commented):
Great job, thanks for pushing this @elshan-eth !

We need to dig a bit deeper in the nested case to get a measurement end to end. There are a lot of things happening at the batch router right now, so we need a swap path that actually goes through those steps to get a better picture.

Other than that, it's good to go!

pkg/vault/test/gas/.hardhat-snapshots/[PoolMock - WithNestedPool - BatchRouter] swap exact in - tokenA-tokenD
```diff
@@ -0,0 +1 @@
+514.8k
```
Comment (jubeira) [inline pkg/vault/test/gas/.hardhat-snapshots/[PoolMock - WithNestedPool - BatchRouter] swap exact in - tokenA-tokenD:1]: this makes much more sense; great!

pvt/benchmarks/src/PoolBenchmark.behavior.ts
```diff
@@ -1356,5 +1241,167 @@ export class Benchmark {
...
+      ],
+      "exactAmountIn": "50000000",
+      "minAmountOut": "0"
+    }
+  ],
+  "deadline": "115792089237316195423570985008687907853269984665640564039457584007913129639935",
+  "wethIsEth": false,
+  "userData": "0x"
+}
+      */
```
jubeira, line 1404: Is this needed?

jubeira (approved):
Awesome; exactly what we need for now! :ship: good job!

pvt/benchmarks/src/PoolBenchmark.behavior.ts
```diff
@@ -29,25 +25,59 @@ import {
...
+};
+
+export type TestSettings = {
+  offNestedPoolTests: boolean;
+};
+
 export class Benchmark {
   _testDirname: string;
   _poolType: string;
+  _settings: TestSettings = { offNestedPoolTests: false };
```
EndymionJkb, line 56: Just a little confused about the. naming here. It skips the nested pool tests if this is false, so it means "turn off nested pool tests?" As is, "off" as a prefix like that means something like "based upon" NestedPoolTests, like you're extending them, vs. turning them off.

Maybe call it `withNestedPool` for consistency with the PoolTag?

EndymionJkb (approved):
Looks good (hard to follow the diff in the behavior file, but I trust the result). Just a question about naming.

PR #1091, opened 2024-11-06
Truncate high precision aggregate fees

### LINKED ISSUE #1095: Truncate to required precision in aggregate fee calculation
It is possible for individually valid fee components to exceed the Vault's 24-bit precision when combined to form the aggregate. This is always avoidable, but is annoying, and eliminating the possibility of reverting is more important than exact fee precision.

EndymionJkb:
# Description

It is possible for the aggregate fee calculation to result in higher precision than the Vault supports, even if the individual components all have low enough precision. This can always be remedied, but is annoying. To avoid this, simply truncate the high precision bits in this edge case.

## Type of change

- [ ] Bug fix 
- [ ] New feature 
- [ ] Breaking change 
- [ ] Dependency changes
- [X] Code refactor / cleanup
- [ ] Other

## Checklist:

## Issue Resolution

Resolves #1095

### DIFF
--- a/pkg/vault/contracts/ProtocolFeeController.sol
+++ b/pkg/vault/contracts/ProtocolFeeController.sol
@@ -180,7 +180,9 @@ contract ProtocolFeeController is
      * @notice Settle fee credits from the Vault.
      * @dev This must be called after calling `collectAggregateFees` in the Vault. Note that since charging protocol
      * fees (i.e., distributing tokens between pool and fee balances) occurs in the Vault, but fee collection
-     * happens in the ProtocolFeeController, the swap fees reported here may encompass multiple operations.
+     * happens in the ProtocolFeeController, the swap fees reported here may encompass multiple operations. The Vault
+     * differentiates between swap and yield fees (since they can have different percentage values); the Controller
+     * combines swap and yield fees, then allocates the total between the protocol and pool creator.
      *
      * @param pool The address of the pool on which the swap fees were charged
      * @param swapFeeAmounts An array with the total swap fees collected, sorted in token registration order
@@ -231,8 +233,11 @@ contract ProtocolFeeController is
                 }
 
                 if (needToSplitFees) {
-                    uint256 totalVolume = feeAmounts[i].divUp(aggregateFeePercentage);
-                    uint256 protocolPortion = totalVolume.mulUp(protocolFeePercentage);
+                    // The Vault took a single "cut" for the aggregate total percentage (protocol + pool creator) for
+                    // this fee type (swap or yield). Now we need to "disaggregate" the total fee amount and divide
+                    // it between the protocol and pool creator, according to their individual percentages.
+                    uint256 totalAggregateFeeCollected = feeAmounts[i].divUp(aggregateFeePercentage);
+                    uint256 protocolPortion = totalAggregateFeeCollected.mulUp(protocolFeePercentage);
 
                     _protocolFeeAmounts[pool][token] += protocolPortion;
                     _poolCreatorFeeAmounts[pool][token] += feeAmounts[i] - protocolPortion;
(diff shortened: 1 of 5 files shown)


pkg/vault/contracts/ProtocolFeeController.sol
```diff
@@ -231,8 +233,11 @@ contract ProtocolFeeController is
                 }
 
                 if (needToSplitFees) {
-                    uint256 totalVolume = feeAmounts[i].divUp(aggregateFeePercentage);
-                    uint256 protocolPortion = totalVolume.mulUp(protocolFeePercentage);
+                    // The Vault took a single "cut" for the aggregate total percentage (protocol + pool creator) for
+                    // this fee type (swap or yield). Now we need to "disaggregate" the total fee amount and divide
+                    // it between the protocol and pool creator, according to their individual percentages.
+                    uint256 totalAggregateFeeCollected = feeAmounts[i].divUp(aggregateFeePercentage);
```
jubeira, line 239: Let's call it the same as in the Vault: https://github.com/balancer/balancer-v3-monorepo/blob/main/pkg/vault/contracts/Vault.sol#L1046 `totalFeeAmount`.

Let's also clarify that the first step is actually reconstructing the total fees, which are used to compute the protocol fees in the first place, and the remainder goes to the pool creator.

I think this is more or less clear elsewhere, but the nuances around this are kinda easy to forget, so the more comments we add, the better.

pkg/vault/test/foundry/ProtocolFeeController.t.sol
```diff
@@ -948,6 +948,62 @@ contract ProtocolFeeControllerTest is BaseVaultTest {
...
+        feeController.setPoolCreatorSwapFeePercentage(pool, FEE_SCALING_FACTOR);
+
+        vm.prank(admin);
+        feeController.setProtocolSwapFeePercentage(pool, CUSTOM_PROTOCOL_SWAP_FEE_PCT);
+
+        uint256 fullPrecisionAggregateFeePercentage = CUSTOM_PROTOCOL_SWAP_FEE_PCT +
+            CUSTOM_PROTOCOL_SWAP_FEE_PCT.complement().mulDown(FEE_SCALING_FACTOR);
+
+        // Full precision should not equal the nominal value
+        assertFalse(fullPrecisionAggregateFeePercentage == CUSTOM_PROTOCOL_SWAP_FEE_PCT, "Fee precision lost");
```
jubeira, line 968: `assertNotEq`?

pkg/vault/test/foundry/ProtocolFeeController.t.sol
```diff
@@ -948,6 +948,62 @@ contract ProtocolFeeControllerTest is BaseVaultTest {
...
+
+        uint256 correctedFeePercentage = (fullPrecisionAggregateFeePercentage / FEE_SCALING_FACTOR) *
+            FEE_SCALING_FACTOR;
+        assertTrue(correctedFeePercentage == CUSTOM_PROTOCOL_SWAP_FEE_PCT, "Fee precision not adjusted");
```
jubeira, line 972: `assertEq`?

pkg/vault/test/foundry/ProtocolFeeController.t.sol
```diff
@@ -948,6 +948,62 @@ contract ProtocolFeeControllerTest is BaseVaultTest {
...
+
+        // Retrieve it from the Vault - should be the same as we set.
+        PoolConfig memory config = vault.getPoolConfig(pool);
+        assertEq(config.aggregateSwapFeePercentage, CUSTOM_PROTOCOL_SWAP_FEE_PCT);
```
jubeira, line 976: error msg :)

pkg/vault/test/foundry/ProtocolFeeController.t.sol
```diff
@@ -948,6 +948,62 @@ contract ProtocolFeeControllerTest is BaseVaultTest {
...
+        // Full precision should not equal the nominal value
+        assertFalse(fullPrecisionAggregateFeePercentage == CUSTOM_PROTOCOL_YIELD_FEE_PCT, "Fee precision lost");
+
+        uint256 correctedFeePercentage = (fullPrecisionAggregateFeePercentage / FEE_SCALING_FACTOR) *
+            FEE_SCALING_FACTOR;
+        assertTrue(correctedFeePercentage == CUSTOM_PROTOCOL_YIELD_FEE_PCT, "Fee precision not adjusted");
+
+        // Retrieve it from the Vault - should be the same as we set.
+        PoolConfig memory config = vault.getPoolConfig(pool);
+        assertEq(config.aggregateYieldFeePercentage, CUSTOM_PROTOCOL_YIELD_FEE_PCT);
```
jubeira, line 1004: (same as above)

jubeira (approved):
LGTM! How about a small fuzz test that simply bombards both set protocol / pool creator fee percentage with arbitrary numbers using up to the min allowed precision?

joaobrunoah (approved):
LGTM!

pkg/vault/test/foundry/ProtocolFeeController.t.sol
EndymionJkb: Keep forgetting that exists...

PR #1098, opened 2024-11-07
Round trip fee guardrail in Recovery Mode

### LINKED ISSUE #1094: Consider roundtrip fees in recovery mode
This should not be necessary, but if it's easy to implement and there's no risk of reverting unnecessarily it might be ok.

EndymionJkb:
# Description

Apply the same roundtrip fee guardrail on recovery mode that we do with regular proportional withdrawals. If there was a preceding add liquidity operation (which there wouldn't normally be in a recovery mode withdrawal), charge the pool's static swap fee.

To keep recovery mode as simple as possible, this fee is charged to the user (i.e., they receive fewer tokens for their BPT), but the tokens simply remain in the pool, accruing value to the other LPs.

## Type of change

- [ ] Bug fix 
- [X] New feature 
- [ ] Breaking change 
- [ ] Dependency changes
- [ ] Code refactor / cleanup
- [ ] Other

## Checklist:

## Issue Resolution

Resolves #1094

### DIFF
--- a/pkg/vault/contracts/VaultExtension.sol
+++ b/pkg/vault/contracts/VaultExtension.sol
@@ -744,6 +744,14 @@ contract VaultExtension is IVaultExtension, VaultCommon, Proxy {
         return _isPoolInRecoveryMode(pool);
     }
 
+    // Needed to avoid stack-too-deep.apply
+    struct RecoveryLocals {
+        bool chargeRoundtripFee;
+        uint256[] swapFeeAmountsRaw;
+        uint256 swapFeePercentage;
+        uint256 numTokens;
+    }
+
     /// @inheritdoc IVaultExtension
     function removeLiquidityRecovery(
         address pool,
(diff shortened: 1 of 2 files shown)


pkg/vault/test/foundry/RecoveryMode.t.sol
```diff
@@ -49,6 +51,39 @@ contract RecoveryModeTest is BaseVaultTest {
...
+        uint256[] memory amountsIn = [uint256(defaultAmount), uint256(defaultAmount)].toMemoryArray();
+
+        vm.prank(alice);
+        (, uint256 bptAmountOut, ) = router.addLiquidityCustom(pool, amountsIn, bptAmount, false, bytes(""));
+
+        // Put pool in recovery mode.
+        vault.manualEnableRecoveryMode(pool);
+        // Set the fee, and flag to trigger collection of it.
+        vault.manualSetStaticSwapFeePercentage(pool, BASE_MAX_SWAP_FEE);
+        vault.manualSetAddLiquidityCalledFlag(pool, true);
```
jubeira, line 65: You don't need this one actually. The whole test runs in the same tx.

Maybe just leave a comment, and / or assert that the flag is true?

pkg/vault/test/foundry/RecoveryMode.t.sol
```diff
@@ -49,6 +51,39 @@ contract RecoveryModeTest is BaseVaultTest {
...
+            pool,
+            alice,
+            initialSupply - amountToRemove, // totalSupply after the operation
+            [-int256(amountOutAfterFee), -int256(amountOutAfterFee)].toMemoryArray(),
+            [feeAmount, feeAmount].toMemoryArray()
+        );
+
+        // Do a recovery withdrawal.
+        vm.prank(alice);
+        router.removeLiquidityRecovery(pool, amountToRemove);
```
jubeira, line 84: Can we please also check balances? I know it's too much, but we're at a very late stage now.

pkg/vault/contracts/VaultExtension.sol
```diff
@@ -760,21 +768,39 @@ contract VaultExtension is IVaultExtension, VaultCommon, Proxy {
...
+        // Normally we expect recovery mode withdrawals to be stand-alone operations. If there is a previous add
+        // operation in this transaction, this might be an attack, so we apply the same guardrail used for regular
+        // proportional withdrawals. To keep things simple, all we do is reduce the `amountsOut`, leaving the "fee"
+        // tokens in the pool.
+        locals.swapFeeAmountsRaw = new uint256[](locals.numTokens);
+        locals.chargeRoundtripFee = _addLiquidityCalled().tGet(pool);
+
+        // Don't make the call to retrieve the fee unless we have to.
+        if (locals.chargeRoundtripFee) {
+            locals.swapFeePercentage = _vault.getStaticSwapFeePercentage(pool);
```
jubeira, line 795: You don't need an external call here; we're at the vault :)
```suggestion
            locals.swapFeePercentage = _poolConfigBits[pool].getStaticSwapFeePercentage();
```

jubeira (commented):
LGTM!

This is super overkill, but I'm not against it. I'd love @danielmkm's thoughts on this one.

pkg/vault/contracts/VaultExtension.sol
```diff
@@ -744,6 +744,14 @@ contract VaultExtension is IVaultExtension, VaultCommon, Proxy {
         return _isPoolInRecoveryMode(pool);
     }
 
+    // Needed to avoid stack-too-deep.apply
```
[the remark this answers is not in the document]
EndymionJkb: I tried using a struct for the `removeLiquidityRecovery` function itself first, as we have elsewhere, but it wasn't enough.

pkg/vault/contracts/VaultExtension.sol
EndymionJkb: Wow; this is what happens when I work too late at night and my brain's already fried :)

pkg/vault/test/foundry/RecoveryMode.t.sol
EndymionJkb: I did this, but it doesn't seem to be testing the content of the events properly. If that flag is set, it should be charging fees in the original test too, but the expected event doesn't have them, yet it passes. I manually clear the flag in the one that's not supposed to have the roundtrip fee, and that works too.

Testing the balances verifies that the two cases behave differently, but I wonder where else events aren't getting tested.

jubeira (approved):
LGTM!

PR #1100, opened 2024-11-08
Support per-token fee collection

EndymionJkb:
# Description

In the unlikely event fee collection for a pool is blocked by a bricked or halted token, add a permissioned call to withdraw fees for individual pool tokens. I suppose it could be permissionless too (shouldn't hurt anything to partially collect fees), but it should be a rare / emergency operation, so it seemed best to assign it to governance. (Should there be a special event when it's called? Right now it just emits the normal fee collected events.)

## Type of change

- [ ] Bug fix 
- [X] New feature 
- [ ] Breaking change 
- [ ] Dependency changes
- [ ] Code refactor / cleanup
- [ ] Other

## Checklist:


pkg/vault/test/foundry/ProtocolFeeController.t.sol
```diff
@@ -870,6 +873,64 @@ contract ProtocolFeeControllerTest is BaseVaultTest {
         );
     }
 
+    function testFeeCollectionPerTokenWithoutPermission() public {
```
joaobrunoah, line 876: It'd be nice to test if the hook is only callable by the vault.

joaobrunoah (approved):
LGTM!

jubeira (commented):
Thanks @EndymionJkb.

I'm not fully convinced about this one. The change looks good, but still: is it really worth the complexity? Pool fees will be collected weekly or whatever. I don't think we're expecting decent tokens from valuable pools to break transfers very often, if ever.

It's not that much of an extra, but the added value here sounds minimal to me. Let me think it a bit further.

EndymionJkb:
> It's not that much of an extra, but the added value here sounds minimal to me. Let me think it a bit further.

They (= Maxis) are saying this does happen (they have a similar situation going on now).

pkg/vault/test/foundry/ProtocolFeeController.t.sol
```diff
@@ -870,6 +873,83 @@ contract ProtocolFeeControllerTest is BaseVaultTest {
...
+        );
+
+        authorizer.grantRole(
+            feeControllerAuth.getActionId(IProtocolFeeController.withdrawProtocolFeesPerToken.selector),
+            admin
+        );
+
+        // Call with a token that isn't in the pool.
+        vm.prank(admin);
+        feeController.collectAggregateFeesPerToken(pool, veBAL);
```
joaobrunoah, line 911: This should be another test, with this specific scenario.

pkg/vault/contracts/VaultAdmin.sol
```diff
@@ -286,12 +287,15 @@ contract VaultAdmin is IVaultAdmin, VaultCommon, Authentication, VaultGuard {
         for (uint256 i = 0; i < poolTokens.length; ++i) {
             IERC20 token = poolTokens[i];
 
-            (totalSwapFees[i], totalYieldFees[i]) = _aggregateFeeAmounts[pool][token].fromPackedBalance();
+            // If a feeToken is specified, all reported fees will be zero except for that token.
```
joaobrunoah, line 290: We should have a test for this change. I believe we're only testing the ProtocolFeeController.

joaobrunoah (approved):
LGTM!

pkg/vault/contracts/VaultAdmin.sol
```diff
@@ -286,12 +287,15 @@ contract VaultAdmin is IVaultAdmin, VaultCommon, Authentication, VaultGuard {
...
+            if (feeToken == IERC20(address(0)) || feeToken == token) {
```
jubeira, line 291: maybe do a boolean 
```
bool collectAllTokens = (feeToken == IERC20(address(0)));
```

outside the outer for loop and use it here for just a bit more clarity?

jubeira:
As per offthread discussion with @danielmkm, we're not moving forward with this.

The PR looks correct and all, but it's a bit hard to justify the added complexity. If fees are collected in a timely manner the loss should be capped shall this ever happen.

PR #1101, opened 2024-11-08
Stable surge v3 with median algorithm

elshan-eth:
# Description

This PR inherited of #1085. This PR has tests and a new algorithm for calculating pool unbalanced factors. 

Algorithm:
It’s important to define what constitutes a point of imbalance. A pool is balanced when it has equal balances for all tokens, which is point 0. A point of imbalance can be considered equal to 1 if, in order to stabilize the pool, changes need to be made to the balances (addition or removal) amounting to the total current balance.

1. For finding a balanced point we use a median of balances. The median represents the closest balanced point for stability.

2. After finding the median we need to calculate how much needs to be added or removed to reach the balanced point: `diff[n] = abs(balance[n] - median)`

3. Calculate total diff
`totalDiff = diff[0] + diff[1] + ... + diff[n]`

4. Calculate unbalanced factor:
`totalDiff/totalBalance`

Example: we have the balances in a stable pool `[100, 1000, 3000, 10000]`
```
(1000 + 3000) / 2 = 2000 (median)
diff[0] = abs(100 - 2000) = 1900
diff[1] = abs(1000 - 2000) = 1000
diff[2] = abs(3000 - 2000) = 1000
diff[3] = abs(10000 - 2000) = 8000

(1900 + 1000 + 1000 + 8000) / 14100 (current total balance) = 0.84397, which means that the pool imbalance is 84%.
```

## Type of change

- [ ] Bug fix 
- [X] New feature 
- [ ] Breaking change 
- [ ] Dependency changes
- [ ] Code refactor / cleanup
- [ ] Other

## Checklist:


pkg/pool-hooks/contracts/StableSurgeHookExample.sol
```diff
@@ -0,0 +1,243 @@
...
+    }
+
+    /// @inheritdoc IHooks
+    function onRegister(
+        address,
+        address pool,
+        TokenConfig[] memory,
+        LiquidityManagement calldata
+    ) public override onlyVault returns (bool) {
+        emit StableSurgeHookExampleRegistered(pool);
```
EndymionJkb, line 95: Check for StablePoolFactory?

pkg/pool-hooks/contracts/StableSurgeHookExample.sol
```diff
@@ -0,0 +1,243 @@
...
+
+    /// @dev Assumes the percentage value has been externally validated (e.g., with `_ensureValidPercentage`).
+    function _setSurgeThresholdPercentage(address pool, uint256 newSurgeThresholdPercentage) private {
+        _surgeThresholdPercentage[pool] = newSurgeThresholdPercentage;
+
+        emit ThresholdSurgePercentageChanged(pool, newSurgeThresholdPercentage);
+    }
+
+    function _ensureValidPercentage(uint256 percentage) private pure {
+        if (percentage > FixedPoint.ONE) {
```
EndymionJkb, line 239:
```suggestion
        if (percentage > MAX_SURGE_FEE_PERCENTAGE) {
```

pkg/pool-hooks/contracts/utils/StableSurgeMedianMath.sol
```diff
@@ -0,0 +1,58 @@
+// SPDX-License-Identifier: GPL-3.0-or-later
+
+pragma solidity ^0.8.24;
+
+library StableSurgeMedianMath {
+    function calculateImbalance(uint256[] memory balances) internal pure returns (uint256) {
```
EndymionJkb, line 6:
```suggestion
    function calculateImbalance(uint256[] memory balancesScaled18) internal pure returns (uint256) {
```
Would document that these have to be scaled (given the 1e18 factor below)

pkg/pool-hooks/contracts/utils/StableSurgeMedianMath.sol
```diff
@@ -0,0 +1,58 @@
...
+        uint256 mid = sortedBalances.length / 2;
+
+        if (sortedBalances.length % 2 == 0) {
+            return (sortedBalances[mid - 1] + sortedBalances[mid]) / 2;
+        } else {
+            return sortedBalances[mid];
+        }
+    }
+
+    // First implementation had quickselect algorithm, but it was removed in the final version
```
EndymionJkb, line 31: Note also the bubble sort in InputHelpers.sortTokens. How does that compare?

elshan-eth: This insert sort uses less gas than bubble sort. The worth case with 8-length arrays uses 15K gas. The average case is 10K.

pkg/pool-hooks/contracts/StableSurgeHookExample.sol
elshan-eth: I don't think we should limit factories. I mean custom factories should be able to use this hook. WDYT about it?

EndymionJkb: My understanding is this is an official Balancer hook that will be released in production. If that's the case, the straightforward thing would be to create a new StablePoolWithSurgeFactory that deploys this hook on deployment, then passes the hook address and its own address as the "allowedFactory" when creating new StablePools.

People wanting to use it with their own custom pools would do the same in their custom pool factory.

It really only works with StablePools. If we wanted to make it more generic, we could deploy it independently, and have an internal `mapping(address factory -> bool allowed)` with a permissioned call to allow governance to maintain a set of allowed factories. We'd add our StablePoolFactory first, and then others if they became available and were accepted by the community.

joaobrunoah: What's the harm of not checking that the pool using this hook is not from Balancer's stable pool factory? And, is it really necessary to make a permissioned call to register other allowed factories?

This hook only uses dynamic swap fee hook, it does not move tokens around, would there be a possible exploit in using this hook in a place that it shouldn't be used? We could probably document that this hook is supposed to be used in a StablePool (or weighted pools with the same weight, it'd also work), but overloading the governance with a permissioned call to allow factories in a hook that, at first sight, is not harmful, seems to be overkill. (I'm probably oversimplifying and missing some very important detail, I just want to know what detail it is)

pkg/pool-hooks/contracts/StableSurgeHookExample.sol
```diff
@@ -0,0 +1,243 @@
...
+    /// @inheritdoc IHooks
+    function onComputeDynamicSwapFeePercentage(
+        PoolSwapParams calldata params,
+        address pool,
+        uint256 staticSwapFeePercentage
+    ) public view override onlyVault returns (bool, uint256) {
+        uint256 amp;
+        (amp, , ) = IStablePool(pool).getAmplificationParameter();
+
+        // In order to calculate `weightAfterSwap` we need balances after swap, so we must compute the swap amount.
```
jubeira, line 112: Is this absolutely needed? Can't we do an approximation with the current balances and the amounts given?

Sounds a bit overkill to compute the whole swap just to compute the fee percentage, given that we'll need to do it all over again in the pool `onSwap` function right after this.

Intuitively, we know whether we start in a balanced / unbalanced state, and we can sort of guess whether the swap will push it further in the right direction or not. Although I get it might be more nuanced than that, as you might start in an imbalanced state, the direction might be correct but the magnitude of the swap pushes it into unbalanced territory but on the opposite side.

pkg/pool-hooks/contracts/StableSurgeHookExample.sol
jubeira: I'd say it's not dangerous, but in `onComputeDynamicSwapFeePercentage` we're treating the pool as a stable pool and calling functions that are not present elsewhere.

If the factory is not the right one, it'll always revert. That can be prevented if we block registrations by mistake here. I would add the check.

pkg/pool-hooks/contracts/StableSurgeHookExample.sol
jubeira: If there's really no way around it, why not just call `pool.onSwap` and get a result as opposed to reimplementing all the logic here?

pkg/pool-hooks/contracts/StableSurgeHookExample.sol
```diff
@@ -0,0 +1,243 @@
...
+
+    /**
+     * @notice Sets the hook threshold percentage.
+     * @dev This function must be permissioned. If the pool does not have a swap fee manager role set, the surge
+     * threshold will be effectively immutable, set to the default threshold for this hook contract.
+     */
+    function setSurgeThresholdPercentage(address pool, uint256 newSurgeThresholdPercentage) external {
+        if (_vault.getPoolRoleAccounts(pool).swapFeeManager != msg.sender) {
+            revert SenderNotAllowed();
+        }
```
jubeira, line 170: This should probably follow the same logic as the swap fee percentage: if there's no swap fee manager, default to governance.

pkg/pool-hooks/contracts/StableSurgeHookExample.sol
```diff
@@ -0,0 +1,243 @@
...
+            return staticFeePercentage;
+        }
+
+        uint256 oldTotalImbalance = StableSurgeMedianMath.calculateImbalance(params.balancesScaled18);
+
+        if (newTotalImbalance <= oldTotalImbalance || newTotalImbalance <= surgeThresholdPercentage) {
+            return staticFeePercentage;
+        }
+
+        // surgeFee = staticFee + (maxFee - staticFee) * (pctImbalance - pctThreshold) / (1 - pctThreshold).
```
jubeira, line 223: Can you expand this a bit further? E.g. explain how this behaves in the limits. When `pctImbalance --> 1`, `surgeFee = maxFee`, when `newTotalImbalance == surgeThresholdPercentage, surgeFee = staticFee` (case above).

pkg/pool-hooks/contracts/utils/StableSurgeMedianMath.sol
```diff
@@ -0,0 +1,58 @@
...
+    // because it seems to be overhead for small arrays (up to 8 elements)
+    // Complexity of this insertion sort is O(n^2) but it is not a problem for small arrays
```
jubeira, line 33: I'd reword this a bit, saying which is the selected algorithm and why (i.e. not explain what was used for previous versions).

pkg/pool-hooks/contracts/utils/StableSurgeMedianMath.sol
```diff
@@ -0,0 +1,58 @@
...
+                sortedBalances[j] = sortedBalances[j - 1];
+                j--;
+            }
+
+            sortedBalances[j] = value;
+        }
+    }
+
+    function absSub(uint256 a, uint256 b) internal pure returns (uint256) {
+        return a > b ? a - b : b - a;
```
jubeira, line 56: can be unchecked.

pkg/pool-hooks/contracts/utils/StableSurgeMedianMath.sol
```diff
@@ -0,0 +1,58 @@
...
+            return (sortedBalances[mid - 1] + sortedBalances[mid]) / 2;
+        } else {
+            return sortedBalances[mid];
+        }
+    }
+
+    // First implementation had quickselect algorithm, but it was removed in the final version
+    // because it seems to be overhead for small arrays (up to 8 elements)
+    // Complexity of this insertion sort is O(n^2) but it is not a problem for small arrays
+    function sort(uint256[] memory balances) internal pure returns (uint256[] memory sortedBalances) {
```
jubeira, line 34: Have you considered this? https://github.com/OpenZeppelin/openzeppelin-contracts/blob/master/contracts/utils/Arrays.sol#L43

Unless this implementation has a significant advantage, I'd say let's just use OZ for basic functionality.

pkg/pool-hooks/contracts/StableSurgeHookExample.sol
```diff
@@ -0,0 +1,243 @@
...
+import { StableMath } from "@balancer-labs/v3-solidity-utils/contracts/math/StableMath.sol";
+import { FixedPoint } from "@balancer-labs/v3-solidity-utils/contracts/math/FixedPoint.sol";
+
+import { StableSurgeMedianMath } from "./utils/StableSurgeMedianMath.sol";
+
+/**
+ * @notice Hook that charges a fee on trades that push a pool into an imbalanced state beyond a given threshold.
+ * @dev Uses the dynamic fee mechanism to apply a directional fee.
+ */
+contract StableSurgeHookExample is BaseHooks, VaultGuard {
```
jubeira, line 28: I'd rename to just `StableSurgeHook`. We want this for real :)

jubeira (commented):
First pass done! Great job figuring this out.

I have a few questions regarding the implementation (not specific to this PR; we've been carrying it since the first iteration). I wonder if we can avoid computing the swap result; if we don't I'd just use the actual pool implementation for simplicity. I'd definitely try to avoid re-implementing it here.

pkg/pool-hooks/contracts/StableSurgeHookExample.sol
elshan-eth: It seems more reasonable to do this after https://github.com/balancer/balancer-v3-monorepo/pulls

pkg/pool-hooks/test/foundry/StableSurgeHookTest.sol
```diff
@@ -0,0 +1,336 @@
...
+            surgeThresholdPercentage,
+            staticFeePercentage
+        );
+
+        uint256[] memory newBalances = new uint256[](numTokens);
+        for (uint256 i = 0; i < numTokens; ++i) {
+            newBalances[i] = balances[i];
+        }
+        newBalances[tokenIn] += amountGivenScaled18;
+        newBalances[tokenOut] -= amountGivenScaled18;
```
elshan-eth, line 220: This calculation is very rough since a stable pool uses a curve, and the input amount does not have a 1:1 ratio with the output amount. I don’t see any other way, except for a highly suboptimal approach where we replace these calculations with actual output amount calculations using `calculateInvariant` and `calculateBalances`. However, in terms of gas costs, this would be equivalent to performing a full swap calculation, which seems impractical. WDYT?

@jubeira @EndymionJkb

pkg/pool-hooks/contracts/StableSurgeHook.sol
```diff
@@ -0,0 +1,244 @@
...
+import { FixedPoint } from "@balancer-labs/v3-solidity-utils/contracts/math/FixedPoint.sol";
+import { Authentication } from "@balancer-labs/v3-solidity-utils/contracts/helpers/Authentication.sol";
+
+import { StableSurgeMedianMath } from "./utils/StableSurgeMedianMath.sol";
+
+/**
+ * @notice Hook that charges a fee on trades that push a pool into an imbalanced state beyond a given threshold.
+ * @dev Uses the dynamic fee mechanism to apply a directional fee.
+ */
+contract StableSurgeHook is BaseHooks, VaultGuard, Authentication {
```
EndymionJkb, line 30: Can use SingletonAuthentication here for simplicity (see #1125).

pkg/pool-hooks/contracts/StableSurgeHook.sol
```diff
@@ -0,0 +1,244 @@
...
+    using FixedPoint for uint256;
+
+    /**
+     * @notice A new `StableSurgeHookExample` contract has been registered successfully.
+     * @dev If the registration fails the call will revert, so there will be no event.
+     * @param pool The pool on which the hook was registered
+     * @param factory The factory that registered the pool
+     */
+    event StableSurgeHookExampleRegistered(address indexed pool, address indexed factory);
```
EndymionJkb, line 39:
```suggestion
    event StableSurgeHookRegistered(address indexed pool, address indexed factory);
```
Production hook. See #1125

pkg/pool-hooks/contracts/StableSurgeHook.sol
```diff
@@ -0,0 +1,244 @@
...
+    // Upgradeable contract in charge of setting permissions.
+    IAuthorizer internal _authorizer;
+
+    // Store the current threshold for each pool.
+    mapping(address pool => uint256 threshold) private _surgeThresholdPercentage;
+
+    constructor(
+        IVault vault,
+        address allowedFactory,
+        IAuthorizer newAuthorizer,
```
EndymionJkb, line 75: Why would we want to set an independent authorizer? Can just use the Vault one (see #1125). I suppose if you wanted to completely generalize it to use a different kind of governance - but the interface is pretty specific, and this is a Balancer hook.

pkg/pool-hooks/contracts/StableSurgeHook.sol
```diff
@@ -0,0 +1,244 @@
...
+    /**
+     * @notice A new authorizer is set by `setAuthorizer`.
+     * @param newAuthorizer The address of the new authorizer
+     */
+    event AuthorizerChanged(IAuthorizer indexed newAuthorizer);
+
+    /// @notice The threshold must be a valid percentage value.
+    error InvalidSurgeThresholdPercentage();
+
+    uint256 public constant MAX_SURGE_FEE_PERCENTAGE = 95e16; // 95%
```
EndymionJkb, line 58: They wanted to have this configurable (see #1125).

EndymionJkb (commented):
Most of the comments here I implemented in the factory version (#1125).

pkg/pool-hooks/test/foundry/StableSurgeHookTest.sol
elshan-eth: I think the only correct option is to call `pool.onSwap` inside the hook. This will increase gas usage, but maybe not significantly. In other words, I would rely here on whether the current rough calculations break the logic of the Surgery Hook or not.

pkg/pool-hooks/test/foundry/StableSurgeHookUnit.t.sol
```diff
@@ -0,0 +1,352 @@
...
+        }
+
+        amountGivenScaled18 = bound(amountGivenScaled18, 1, balances[tokenOut]);
+
+        return (length, tokenIn, tokenOut, amountGivenScaled18, balances);
+    }
+
+    function _buildSwapParams(
+        uint256 tokenIn,
+        uint256 tokenOut,
```
jubeira, line 320:
```suggestion
        uint256 indexIn,
        uint256 indexOut,
```

pkg/pool-hooks/test/foundry/StableSurgeHookUnit.t.sol
```diff
@@ -0,0 +1,352 @@
...
+            STATIC_FEE_PERCENTAGE
+        );
+        _registerPool();
+        assertEq(surgeFeePercentage, STATIC_FEE_PERCENTAGE, "Surge fee percentage should be staticFeePercentage");
+    }
+
+    function _boundValues(
+        uint256 length,
+        uint256 tokenIn,
+        uint256 tokenOut,
```
jubeira, line 297:
```suggestion
        uint256 indexIn,
        uint256 indexOut,
```

pkg/pool-hooks/test/foundry/StableSurgeMedianMath.t.sol
```diff
@@ -0,0 +1,96 @@
...
+contract StableSurgeMedianMathTest is BaseVaultTest {
+    using Arrays for uint256[];
+    using InputHelpers for uint256[];
+
+    uint256 constant MIN_TOKENS = 2;
+    uint256 constant MAX_TOKENS = 8;
+
+    StableSurgeMedianMathMock stableSurgeMedianMathMock = new StableSurgeMedianMathMock();
+
+    function testAbsSub_Fuzz(uint256 a, uint256 b) public view {
```
jubeira, line 19:
```suggestion
    function testAbsSub__Fuzz(uint256 a, uint256 b) public view {
```

pkg/pool-hooks/test/foundry/StableSurgeMedianMath.t.sol
```diff
@@ -0,0 +1,96 @@
...
+        assertEq(
+            stableSurgeMedianMathMock.absSub(1, MAX_UINT256),
+            MAX_UINT256 - 1,
+            "abs(1 - MAX_UINT256) != MAX_UINT256 - 1"
+        );
+        assertEq(stableSurgeMedianMathMock.absSub(MAX_UINT256, 0), MAX_UINT256, "abs(MAX_UINT256 - 0) != MAX_UINT256");
+        assertEq(stableSurgeMedianMathMock.absSub(0, MAX_UINT256), MAX_UINT256, "abs(0 - MAX_UINT256) != MAX_UINT256");
+    }
+
+    function testFindMedian_Fuzz(uint256 length, uint256[8] memory rawBalances) public view {
```
jubeira, line 52:
```suggestion
    function testFindMedian__Fuzz(uint256 length, uint256[8] memory rawBalances) public view {
```

pkg/pool-hooks/test/foundry/StableSurgeMedianMath.t.sol
```diff
@@ -0,0 +1,96 @@
...
+            expectedMedian = (sortedBalances[mid - 1] + sortedBalances[mid]) / 2;
+        } else {
+            expectedMedian = sortedBalances[mid];
+        }
+
+        uint256 median = stableSurgeMedianMathMock.findMedian(balances);
+        assertEq(median, expectedMedian, "Median is not correct");
+    }
+
+    function testCalculateImbalance_Fuzz(uint256 length, uint256[8] memory rawBalances) public view {
```
jubeira, line 74:
```suggestion
    function testCalculateImbalance__Fuzz(uint256 length, uint256[8] memory rawBalances) public view {
```

pkg/pool-hooks/test/foundry/StableSurgeMedianMath.t.sol
```diff
@@ -0,0 +1,96 @@
...
+        uint256 totalBalance = 0;
+        uint256 totalDiffs = 0;
+
+        for (uint256 i = 0; i < balances.length; i++) {
+            totalBalance += balances[i];
+
+            totalDiffs += stableSurgeMedianMathMock.absSub(balances[i], median);
+        }
+
+        uint256 expectedImbalance = (totalDiffs * 1e18) / totalBalance;
```
jubeira, line 91: alright, although I'd use FP math here too

pkg/solidity-utils/contracts/helpers/InputHelpers.sol
```diff
@@ -103,4 +103,19 @@ library InputHelpers {
             previous = current;
         }
     }
+
+    /// @dev Ensure an array of amounts is sorted. As above, does not validate length or uniqueness.
+    function ensureSortedAmounts(uint256[] memory amounts) internal pure {
+        uint256 previous = amounts[0];
+
+        for (uint256 i = 1; i < amounts.length; ++i) {
```
jubeira, line 111: kinda obvious, but maybe assert length > 2?

pkg/solidity-utils/contracts/openzeppelin/Arrays.sol
```diff
@@ -0,0 +1,237 @@
+// SPDX-License-Identifier: MIT
+// OpenZeppelin Contracts (last updated v5.1.0) (utils/Arrays.sol)
+// This file was procedurally generated from scripts/generate/templates/Arrays.js.
+
+// NOTE: This file copied only the sort and necessary helper functions from Arrays.sol (OpenZeppelin Contracts v5.1.0)
```
jubeira, line 5: Are we copying it because it wasn't released in the version we're using? If so I'd just consider upgrading and deleting this file.

pkg/solidity-utils/contracts/openzeppelin/Math.sol
jubeira, line 1: Is this necessary?

jubeira (commented):
Second pass done, mostly minor changes.

I see we're not calling `onSwap` anymore as pointed out [here](https://github.com/balancer/balancer-v3-monorepo/pull/1101#discussion_r1856637301). I need to think about it a bit further. The problem is that if you're farther away from the balanced point, the balance here is linear but in practice the output will be far from 1:1.

I agree that calling `onSwap` is overkill, but I wonder if there's a better approximation.

pkg/pool-hooks/test/foundry/StableSurgeHookTest.sol
jubeira: I get the feeling that we'd need to analyze this a bit further.

For a 2-pool token, the median is the average.
If you consider that the change is going to be 1:1, the median doesn't change after the operation. E.g. you start with 100, 200 tokens, add X. Original average is 150, new average is `[(100 + x) + (200 - x)] / 2 = 150`.

If you apply the trading curve that shouldn't be the case in the extremes, where you have to add a lot of tokens A to get just a few tokens B.

Then, that median is used to compute the imbalance.
I'm not saying that the current approach is wrong, but sounds like the effect will be quite different from the previous approach.

pkg/pool-hooks/contracts/StableSurgeHook.sol
```diff
@@ -0,0 +1,302 @@
...
+        address swapFeeManager = _vault.getPoolRoleAccounts(pool).swapFeeManager;
+
+        if (swapFeeManager == address(0)) {
+            if (_canPerform(getActionId(msg.sig), msg.sender, pool) == false) {
+                revert SenderNotAllowed();
+            }
+        } else if (swapFeeManager != msg.sender) {
+            revert SenderNotAllowed();
+        }
+    }
```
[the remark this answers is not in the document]
EndymionJkb: It depends how specific this is to StableSurge. We don't *have* to tie hook parameters to a particular pool role account. This is a Singleton hook with pool-specific parameters, but this pattern might not be common. Some hooks might be deployed per pool; others might have global parameters that aren't pool-dependent (should probably be the owner). 

It makes sense in this case to let the same account that can set swap fees for the corresponding pool call `setMaxSurgeFeePercentage` and `setSurgeThresholdPercentage`... except the governance part here looks like it wouldn't really work anyway. We don't have "where" on the basic authorizer, and anyway the functions are being called on the hook contract, not the pool.

I guess (on the TimelockAuthorizer anyway) we could create an actionId and check for a function that doesn't actually exist, but it seems hacky. We could use the actionId for `setStaticSwapFeePercentage`, but can't actually target the pool in the Basic Authorizer.

Maybe for this version of the hook/factory, simplify by using the swapFeeManager only?

pkg/pool-hooks/test/foundry/StableSurgeHookTest.sol
elshan-eth: @EndymionJkb did the calculation in excel and we came to the conclusion that median works better (if we talk about the case of more than 2 tokens)

pkg/pool-hooks/contracts/StableSurgeHook.sol
```diff
@@ -0,0 +1,338 @@
...
+                    newBalances[i] += amountCalculatedScaled18;
+                }
+            } else {
+                if (i == params.indexIn) {
+                    newBalances[i] += amountCalculatedScaled18;
+                } else if (i == params.indexOut) {
+                    newBalances[i] -= params.amountGivenScaled18;
+                }
+            }
+        }
```
jubeira, line 284: I think we don't need the loop.
Also I think the signs for exact in are upside down. We always need to add for index in, and always need to subtract for index out; it's a matter of which one we use in each case.

```suggestion
        if (params.kind == SwapKind.EXACT_IN) {
            newBalances[params.indexIn] += params.amountGivenScaled18;
            newBalances[params.indexOut] -= amountCalculatedScaled18;
        } else {
            newBalances[params.indexIn] += amountCalculatedScaled18;
            newBalances[params.indexOut] -= params.amountGivenScaled18;
        }
```

pkg/pool-hooks/contracts/StableSurgeHook.sol
```diff
@@ -0,0 +1,338 @@
...
+        } else {
+            amountCalculatedScaled18 = StableMath.computeInGivenExactOut(
+                currentAmp,
+                params.balancesScaled18,
+                params.indexIn,
+                params.indexOut,
+                params.amountGivenScaled18,
+                invariant
+            );
+        }
```
jubeira, line 265: I'd just call `onSwap`, removing `onlyVault` tbh. At the end of the day it's a `view` function in this case; I don't see how calling it can hurt anyone. If it wasn't `view` it'd be a different story. I don't think you can force any sort of read only reentrancy in this case. The code here should do exactly the same as all the inputs to `onSwap` should remain the same (amp, balances, index in / out); if you think I'm mistaken here please correct me :)

That won't work with the current deployed factory but I guess that's ok. I think I prefer that as opposed to copying the code here.

jubeira (commented):
Thanks for iterating on this; super useful to keep adding feedback. Just a few more comments for now.

The circular dependency between the fee percentage and the result of the swap function is tricky. I also didn't remember about `onlyVault`, but I think we can get rid of it in this case and that would be the best choice.

pkg/solidity-utils/contracts/helpers/InputHelpers.sol
elshan-eth: We use it only for tests

elshan-eth: As ensureSortedTokens

elshan-eth: I added return if length < 2

pkg/pool-hooks/contracts/StableSurgeHook.sol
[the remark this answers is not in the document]
elshan-eth: I think we should leave it as is for now. I also don't think it should be taken to the library because we use it only here.

pkg/pool-stable/contracts/StableSurgePoolFactory.sol
```diff
@@ -0,0 +1,117 @@
...
+import { BasePoolFactory } from "@balancer-labs/v3-pool-utils/contracts/BasePoolFactory.sol";
+import { StableMath } from "@balancer-labs/v3-solidity-utils/contracts/math/StableMath.sol";
+import { Version } from "@balancer-labs/v3-solidity-utils/contracts/helpers/Version.sol";
+
+import { StablePool } from "./StablePool.sol";
+
+/// @notice Stable Pool factory that deploys a standard StablePool with a StableSurgeHook.
+contract StableSurgePoolFactory is IPoolVersion, BasePoolFactory, Version {
+    uint256 private immutable _defaultSurgeThresholdPercentage;
+    uint256 private immutable _defaultMaxSurgeFeePercentage;
```
joaobrunoah, line 24: Looks like these are not used.

jubeira (approved):
LGTM, great job! I think this is good enough to move forward.

PR #1105, opened 2024-11-13
Add Buffer Balances to Wrap/Unwrap and Buffer liquidity events

### LINKED ISSUE #117: Address V2 Pain Points from Data Team
https://www.notion.so/V2-Pain-Points-a66a412d7a4644b8854467c45aabef37?pvs=4

joaobrunoah:
# Description

The Data team needs buffer balances in these events so they can properly measure the buffer interactions.

## Type of change

- [ ] Bug fix 
- [ ] New feature 
- [ ] Breaking change 
- [ ] Dependency changes
- [x] Code refactor / cleanup
- [ ] Other

## Checklist:

## Issue Resolution

Closes #117 
This is "addresses pain points from the SDK team." Is this the last such thing?


pkg/vault/contracts/Vault.sol
```diff
@@ -1171,25 +1182,21 @@ contract Vault is IVaultMain, VaultCommon, Proxy {
...
                 // We have verified above that this is safe to do unchecked.
-                newDerivedBalance = bufferBalances.getBalanceDerived() - amountOutWrapped;
+                bufferWrappedBalance = bufferBalances.getBalanceDerived() - amountOutWrapped;
             }
-
-            bufferBalances = PackedTokenBalance.toPackedBalance(
-                bufferBalances.getBalanceRaw() + amountInUnderlying,
-                newDerivedBalance
-            );
+            bufferBalances = PackedTokenBalance.toPackedBalance(bufferUnderlyingBalance, bufferWrappedBalance);
```
jubeira, line 1199: I mean this looks good in general, but the stakes are too high at this point. 

I don't feel comfortable modifying all of these lines just to modify a few fields in an event.

How about just returning the packed balances and unpacking outside? Maybe not the most efficient, but you won't need any extra (warm) storage reads and the diff will be even smaller.

jubeira (commented):
Thanks for pushing this @joaobrunoah !

The changes look good, but I'm a bit more reluctant than usual with things like these. Let me think a bit more about it.

pkg/vault/contracts/Vault.sol
jubeira: The function now returns 4 arguments, there are multiple if / else levels, early returns, etc. If we stare too much at the function we might break it.

joaobrunoah:
Superseded by #1107

PR #1107, opened 2024-11-13
Buffer balances on events

jubeira:
# Description

Based on #1105 from @joaobrunoah, but cutting a corner to simplify the code / save gas.
Emit the balances as packed `bytes32` instead of individually. We always read the balances inside `_wrap` and `_unwrap`, and the bytes32 field is always up to date at the end of the function because we actively use it inside every if / else combination. Therefore the changes are minimized, together with the gas as well.

@mendesfabio do you think this is acceptable? I _really_ want to minimize vault changes at this point. You can reconstruct the buffer balances by splitting the result into two `uint128` words. If we go this way we'll probably want to emit `bytes32` for add and remove liquidity for consistency (although in those cases we can afford to emit the values separately).

TODO:
- [ ] fix tests
- [x] adjust add / remove events

## Type of change

- [ ] Bug fix 
- [x] New feature 
- [ ] Breaking change 
- [ ] Dependency changes
- [ ] Code refactor / cleanup
- [ ] Other

## Checklist:

## Issue Resolution

N/A


mendesfabio:
Looks acceptable to me. Vini is verifying we can decode it on Dune but it should be possible

viniabussafi:
works on Dune as well, so all clear on our side

joaobrunoah (approved):
LGTM! :D

elshan-eth (approved):

PR #1109, opened 2024-11-13
Try external calls in recovery mode

jubeira:
# Description

`emitTransfer` and `emitApproval` are called inside recovery mode exit.
A malicious pool can revert during those calls when recovery mode is enabled to deny exits.

This should never be the case in practice; good behaved pools should just use `BalancerPoolToken` and simply emit the event, and bad pools should never attract liquidity anyways. In this regard, this change is completely redundant and unnecessary.

But if by any chance a pool dev gets very creative and implements something else that can fail and goes unnoticed, it's better to be safe than sorry.

## Type of change

- [x] Bug fix 
- [ ] New feature 
- [ ] Breaking change 
- [ ] Dependency changes
- [ ] Code refactor / cleanup
- [ ] Other

## Checklist:

## Issue Resolution

N/A


EndymionJkb (approved):
Verified that the events are reverting, and it still works. Could actually make a custom pool that does this, but that would be overkill.

joaobrunoah (approved):
LGTM! By the way, I also confirmed that the tests fail in main.

jubeira:
> Verified that the events are reverting, and it still works. Could actually make a custom pool that does this, but that would be overkill.

Using the mock pool would affect gas tests, so you'd need another one. That's why I opted for mock calls.

PR #1117, opened 2024-11-16
Check array lengths in permitBatchAndCall

EndymionJkb:
# Description

Unlikely to occur in practice, but the function is missing a length check. Unfortunately, this adds to the bytecode (though it was already over). Since this is really a pre-condition, I tried require, but that increased the bytecode even more.

If we decide not to do this, we could add a comment to IRouterCommon instead of changing the code:
@dev This function assumes that the `permitBatch` and `permitSignatures` arrays are parallel; this is not verified, and must be guaranteed externally.

## Type of change

- [X] Bug fix 
- [ ] New feature 
- [ ] Breaking change 
- [ ] Dependency changes
- [ ] Code refactor / cleanup
- [ ] Other

## Checklist:


jubeira (approved):
LGTM; this is fine. We'll have some bytes to spare when we move the buffer functionality into a separate router (#1108)

elshan-eth (approved):
LGTM but I didn't quite get one thing. Is there some vulnerability? If it is not there, then we can do without this check.

jubeira:
No, but still a good sanity check I guess.

PR #1126, opened 2024-11-20
Re-enable queries

### LINKED ISSUE #96: Re-enable queries
The mechanism to disable the queries is currently a one-way trigger; they cannot be re-enabled.

While this is not a fatal issue because queries can still be implemented elsewhere, it'd be good to be able to re-enable them (probably behind a timelock).

Ref: https://github.com/balancer/balancer-v3-monorepo/pull/82/files#r1299498871.

jubeira:
# Description

We've had quite some back and forth regarding re-enabling queries.
The rationale behind a permanent disable is that if there's a problem, there's likely no going back, and re-enabling them puts the protocol at risk. The argument for re-enabling is that we might need to rush to disable them during an emergency, but a false positive would be inconvenient.

Perhaps the solution is to have 2 buttons: a red button, and a big red button? The solution is dead simple and doesn't rely on timelocks to work. Just always use the simple one during an emergency, and the permanent one after confirmation.

Will add tests if this makes sense.

## Type of change

- [ ] Bug fix 
- [x] New feature 
- [ ] Breaking change 
- [ ] Dependency changes
- [ ] Code refactor / cleanup
- [ ] Other

## Checklist:

## Issue Resolution

Closes #96.


danielmkm (approved):
Looks good

elshan-eth (approved):
Cool!

PR #1128, opened 2024-11-21
Add stable pool amp aux events.

johngrantuk:
# Description

In an event based state system (e.g. as used by Paraswap) pool state, etc is tracked using events. It's very convenient to be able to listen for all relevant events from the Vault (which is already a known contract), rather than have to track/add pools. 

This PR makes the stable pool emit amp related events via the Vault using `emitAuxiliaryEvent`. I kept the original events emitted from the pool itself as they could be useful and the extra gas cost shouldn't be an issue in the infrequently called functions.

## Type of change

- [ ] Bug fix 
- [X] New feature 
- [ ] Breaking change 
- [ ] Dependency changes
- [ ] Code refactor / cleanup
- [ ] Other

## Checklist:


jubeira (approved):
LGTM!

elshan-eth (approved):
LGTM! But some tests were broken

PR #1127, opened 2024-11-21
add pools array to factory

elshan-eth:
# Description

The SDK team asked to be able to get a list of pool addresses from factories. The `getPools(uint256 start, uint256 count)` function allows you to get a list with pagination. Pagination is needed to avoid going over the gas limit when receiving data. The returned list will never exceed the length of the array, so if we want to get the entire array list, we can pass count as MAX_UINT_256.

## Type of change

- [ ] Bug fix 
- [ ] New feature 
- [ ] Breaking change 
- [ ] Dependency changes
- [ ] Code refactor / cleanup
- [ ] Optimization: [ ] gas / [ ] bytecode
- [ ] Documentation or wording changes
- [ ] Other

## Checklist:

- [ ] The diff is legible and has no extraneous changes
- [ ] Complex code has been commented, including external interfaces
- [ ] Tests have 100% code coverage
- [ ] The base branch is either `main`, or there's a description of how to merge

## Issue Resolution

Closes #1112

PR #1133, opened 2024-11-25
Generalization of pool count

EndymionJkb:
# Description

Built on #1127 (already merged), this uses more v2-like conventions (e.g., in the LM system), especially including a count, as otherwise off-chain processes would need to rely on the subgraph for this. It is also written to be tolerant, so that you could just give it 10,000 if there are 50 pools, and it would return 50, but it seems logical to also include an on-chain count.

We could go full v2 by having a `getPoolAtIndex`, but it might be overkill. It really depends on how we expect this to be used. Added some tests, and use a custom error (as we do elsewhere, including in this contract).

## Type of change

- [ ] Bug fix 
- [ ] New feature 
- [ ] Breaking change 
- [ ] Dependency changes
- [X] Code refactor / cleanup
- [ ] Other

## Checklist:


jubeira (approved):
LGTM!

elshan-eth (approved):
LGTM!

PR #1142, opened 2024-11-28
Fix Action Ids of pool functions

joaobrunoah:
# Description

Create3 uses a proxy contract to deploy the pools, so the pool receives the address of the proxy instead of the factory when it's created, which causes issues to generate action ids. To fix that, this PR replaces Create3 by Create2 when creating a new pool.

## Type of change

- [x] Bug fix 
- [ ] New feature 
- [ ] Breaking change 
- [ ] Dependency changes
- [ ] Code refactor / cleanup
- [ ] Other

## Checklist:


EndymionJkb:
There are comments referencing create3 in the pool factories that we'd need to update (including Gyro pools). _create3 is still used for the Vault, right? So we still need the library.

pkg/pool-stable/test/foundry/RoundingDirectionEdgeCases.t.sol
```diff
@@ -79,6 +83,18 @@ contract RoundingDirectionStablePoolEdgeCasesTest is BasePoolTest {
                 )
             )
         );
+
+        // poolArgs is used to check pool deployment address with create2.
+        poolArgs = abi.encode(
```
jubeira, line 88: Hm, this looks like an implicit return argument to me.

If it doesn't cause too much trouble I would consider returning this explicitly and storing it in the base pool test. Otherwise we're kinda breaking the encapsulation of the tests. It's not terrible at this point, but I don't think it's ideal.

pkg/pool-utils/contracts/BasePoolFactory.sol
```diff
@@ -79,8 +78,26 @@ abstract contract BasePoolFactory is IBasePoolFactory, SingletonAuthentication,
     }
 
     /// @inheritdoc IBasePoolFactory
-    function getDeploymentAddress(bytes32 salt) public view returns (address) {
-        return CREATE3.getDeployed(_computeFinalSalt(salt));
+    function getDeploymentAddress(
```
jubeira, line 81: I'd use OZ here: https://github.com/OpenZeppelin/openzeppelin-contracts/blob/master/contracts/utils/Create2.sol

pkg/pool-utils/contracts/BasePoolFactory.sol
```diff
@@ -117,7 +134,20 @@ abstract contract BasePoolFactory is IBasePoolFactory, SingletonAuthentication,
     }
 
     function _create(bytes memory constructorArgs, bytes32 salt) internal returns (address pool) {
-        pool = CREATE3.deploy(_computeFinalSalt(salt), abi.encodePacked(_creationCode, constructorArgs), 0);
+        bytes memory creationCode = abi.encodePacked(_creationCode, constructorArgs);
+        bytes32 finalSalt = _computeFinalSalt(salt);
+        assembly {
+            pool := create2(0, add(creationCode, 32), mload(creationCode), finalSalt)
```
jubeira, line 140: Same here: https://github.com/OpenZeppelin/openzeppelin-contracts/blob/master/contracts/utils/Create2.sol. This low level bit is abstracted by `deploy`.

I understand you still need to pack the constructor args with the library.

pkg/pool-weighted/test/foundry/WeightedPool8020Factory.t.sol
```diff
@@ -92,7 +106,7 @@ contract WeightedPool8020FactoryTest is WeightedPoolContractsDeployer, VaultCont
         _createPool(tokenA, tokenB);
 
         // Should not be able to deploy identical pool
-        vm.expectRevert("DEPLOYMENT_FAILED");
+        vm.expectRevert();
```
jubeira, line 109: Using OZ will give you a better reason if I'm not mistaken.

Let's try to avoid generic expect reverts in the tests.

jubeira (commented):
Looks good, thanks @joaobrunoah ! Just a few comments; if it's too invasive to return `poolArgs` in `createPool` and breaks too many tests let me know.

Can we also add a test that creates a few pools and checks that the action ID for a given action is the same? That's the original probblem we were trying to solve.

pkg/vault/package.json
```diff
@@ -26,7 +26,7 @@
     "prettier": "npx prettier --write --plugin=prettier-plugin-solidity 'contracts/**/*.sol' 'test/**/*.sol'",
     "test": "yarn test:hardhat && yarn test:forge",
     "test:hardhat": "hardhat test",
-    "test:forge": "yarn build && ./forge-test.sh",
+    "test:forge": "yarn build && REUSING_HARDHAT_ARTIFACTS=true ./forge-test.sh",
```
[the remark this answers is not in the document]
joaobrunoah: Just because it makes the vault's `test:forge` similar to other packages `test:forge`, it doesn't change anything, but it's also easier to delete the argument if we want to run the tests with forge.

pkg/pool-stable/test/foundry/RoundingDirectionEdgeCases.t.sol
joaobrunoah: Done!

pkg/pool-utils/contracts/BasePoolFactory.sol
joaobrunoah: Done!

pkg/pool-utils/contracts/BasePoolFactory.sol
joaobrunoah: Done!

EndymionJkb (commented):
Added a follow-on with some refactors (easier than describing them)

elshan-eth (approved):
I agree with the previous comments, but overall LGTM!

pkg/vault/test/foundry/ActionIds.t.sol
```diff
@@ -0,0 +1,68 @@
...
+                roleAccounts,
+                1e12, // Set min swap fee
+                address(0),
+                false, // Do not enable donations
+                false, // Do not disable unbalanced add/remove liquidity
+                bytes32("salt2")
+            )
+        );
+
+        bytes4 selector = bytes4(keccak256(bytes("transfer(address,uint256)")));
```
jubeira, line 60: You can use an actual selector from stable pool (which is actually the original motivation for this change).
```suggestion
        bytes4 selector = StablePool.startAmplificationParameterUpdate.selector;
```

jubeira (approved):
LGTM, awesome job! Just one minor comment for simplicity in the test; other than that looks great! Thanks for pushing this and addressing the comments.

PR #1143, opened 2024-11-29
Add indexing to Auxiliary events

EndymionJkb:
# Description

Add indexing to VaultAuxiliary events. And change the key to bytes32 so that it can also be indexed.

I recall some discussion about this (not sure of the context). I don't think we care too much about the efficiency of this event (expected to be used by aggregators), but it doesn't seem to hurt anything.

If we want to keep the string key for some reason, it would be possible to retain that and only index the pool.

## Type of change

- [ ] Bug fix 
- [ ] New feature 
- [ ] Breaking change 
- [ ] Dependency changes
- [X] Code refactor / cleanup
- [ ] Other

## Checklist:


joaobrunoah (approved):
LGTM!

jubeira (commented):
@johngrantuk here's an alternative version of the auxiliary event you've proposed.

I think it makes sense; 32 bytes should be OK for the event key, and having the arguments indexed might be useful. WDYT? Deferring final approval to you.

EndymionJkb:
This was motivated by an audit finding.

johngrantuk:
This was the original feedback from Daniel:
dont think we care about indexing, and these will be primarily admin related functions, so not gas sensitive
the idea is the pool will also emit a more readable version of the event, this event is for ease of integration for aggregators

I also agreed with that but don't see a big problem if we really want to change. Main issue I see is aggregators are currently working on deploy11 and this would be a change they need to handle on launch.

johngrantuk (approved):

PR #1151, opened 2024-12-01
Roundtrip fee per session

jubeira:
# Description

If one `addLiquidity` operation hits a given pool before a `removeLiquidityProportional`, roundtrip fees will be charged.
There's a small chance this may happen in the same transaction, while being two separate legitimate user actions. In practice, slippage should make the remove operation revert, which is annoying.

Strictly speaking, the roundtrip fee is designed to defend the system against add / remove in the same transient context, not the whole transaction. The initial problem was that since the flag works per pool, you'd need to register the flags in a place where they can be iterated to be able to delete them when the transient context finishes.

This PR adjust the roundtrip fee to have ideal behavior, so that it's only applied within the same transient context.
If two bundled operations add and remove liquidity from the same pool in the same tx it won't be charged as long as the vault is locked before removing.

~~Draft for now to illustrate the concept; no tests. This is still an edge case and it might not be worth the extra complexity.~~

## Type of change

- [ ] Bug fix 
- [x] New feature 
- [ ] Breaking change 
- [ ] Dependency changes
- [ ] Code refactor / cleanup
- [ ] Other

## Checklist:

## Issue Resolution

See https://cantina.xyz/code/d9a77a0d-c21e-4f49-8889-fbfae22419a7/findings/4.


elshan-eth:
This pr is ready for review

pkg/vault/test/foundry/VaultLiquidity.t.sol
```diff
@@ -394,17 +395,83 @@ contract VaultLiquidityTest is BaseVaultTest {
...
+        (amountsIn, , , amountsOut) = router.manualAddAndRemoveLiquidity(
+            RouterMock.ManualAddRemoveLiquidityParams({
+                pool: pool,
+                sender: alice,
+                maxAmountsIn: amountsIn,
+                minBptAmountOut: defaultAmount
+            })
+        );
 
         // The whole test runs in the same transaction, so transient storage is set.
```
EndymionJkb, line 414:
```suggestion
        // The whole test runs in the same transaction, so transient storage is set for sessionId 0.
```

pkg/vault/test/foundry/VaultLiquidity.t.sol
```diff
@@ -394,17 +395,83 @@ contract VaultLiquidityTest is BaseVaultTest {
...
-        assertTrue(vault.getAddLiquidityCalledFlag(pool), "addLiquidityCalled flag not set");
+        assertTrue(vault.manualGetAddLiquidityCalledFlagBySession(pool, 0), "addLiquidityCalled flag not set");
```
EndymionJkb, line 415:
```suggestion
        assertTrue(vault.manualGetAddLiquidityCalledFlagBySession(pool, 0), "addLiquidityCalled flag not set");
        // But will not be set for the current session (1).
        assertEq(vault.manualGetCurrentUnlockSessionId(), 1, "Wrong sessionId");
        assertFalse(vault.getAddLiquidityCalledFlag(pool), "addLiquidityCalled flag still set");
```
Test that the main flag gets reset after the `unlock` returns.

EndymionJkb: Add this:

IVaultMainMock.sol:
```
    function manualGetCurrentUnlockSessionId() external view returns (uint256);
```
VaultMock.sol:
```
    function manualGetCurrentUnlockSessionId() public view returns (uint256) {
        return _sessionIdSlot().tload();
    }
```

EndymionJkb (commented):
Have some suggestions for expanding/documenting a bit more.

jubeira (commented):
Nice tests! Only one comment (I'll push a small fix)

pkg/vault/test/foundry/RecoveryMode.t.sol
```diff
@@ -77,7 +75,7 @@ contract RecoveryModeTest is BaseVaultTest {
         // Put pool in recovery mode.
         vault.manualEnableRecoveryMode(pool);
         // Check that we are paying round-trip fees
-        assertTrue(vault.getAddLiquidityCalledFlag(pool), "Transient AddLiquidity flag set");
+        assertTrue(vault.manualGetAddLiquidityCalledFlagBySession(pool, 0), "Transient AddLiquidity flag set");
```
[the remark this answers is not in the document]
EndymionJkb: See https://github.com/balancer/balancer-v3-monorepo/blob/session-roundtrip-fee/pkg/vault/test/foundry/VaultLiquidity.t.sol#L414-L418 for an example of checking both (added a function in the mock for getting the current id to support it)

pkg/vault/contracts/Vault.sol
```diff
@@ -109,6 +109,22 @@ contract Vault is IVaultMain, VaultCommon, Proxy {
             }
 
             _isUnlocked().tstore(false);
+
+            // If a user adds liquidity to a pool, then does a proportional withdrawal from that pool during the same
+            // interaction, the system charges a "round-trip" fee on the withdrawal. (This is not a valid use case, and
+            // might be part of an attack.)
```
joaobrunoah, line 115:
```suggestion
            // If a user adds liquidity to a pool, then does a proportional withdrawal from that pool during the same
            // interaction, the system charges a "round-trip" fee on the withdrawal. This fee makes it harder for an
            // user to add liquidity to a pool using a virtually infinite flash loan, swapping in the same pool in a way 
            // that benefits him and removes liquidity in the same transaction, which is not a valid use case.
```

Just complementing on why it's not a valid use-case.

joaobrunoah (approved):
LGTM!

elshan-eth (approved):
Approve for merging

PR #1154, opened 2024-12-02
Vault factory upgrades

jubeira:
# Description

Adjust `VaultFactory` for ease of use:
- replace authorizer with only owner.
- add getters for deployed contract addresses.
- allow to call `create` twice without any effects (useful to verify contracts in deployments repo).
- make `create` nonReentrant (why not?)

## Type of change

- [ ] Bug fix 
- [x] New feature 
- [ ] Breaking change 
- [ ] Dependency changes
- [ ] Code refactor / cleanup
- [ ] Other

## Checklist:

## Issue Resolution

N/A


pkg/vault/contracts/VaultFactory.sol
```diff
@@ -73,9 +80,13 @@ contract VaultFactory is Authentication {
         bytes32 salt,
         address targetAddress,
         bytes calldata vaultCreationCode,
-        bytes calldata vaultAdminCreationCode,
-        bytes calldata vaultExtensionCreationCode
-    ) external authenticate {
+        bytes calldata vaultExtensionCreationCode,
+        bytes calldata vaultAdminCreationCode
+    ) external onlyOwner nonReentrant {
+        if (deploymentComplete) {
```
elshan-eth, line 86: It seems that this is not necessary because on re-entry it will crash on create2 and nonReentrant the same

pkg/vault/contracts/VaultFactory.sol
```diff
@@ -2,28 +2,37 @@
...
-import { IVault } from "@balancer-labs/v3-interfaces/contracts/vault/IVault.sol";
-import { Create2 } from "@openzeppelin/contracts/utils/Create2.sol";
 
 import { ProtocolFeeController } from "./ProtocolFeeController.sol";
 import { VaultExtension } from "./VaultExtension.sol";
 import { VaultAdmin } from "./VaultAdmin.sol";
 
 /// @notice One-off factory to deploy the Vault at a specific address.
-contract VaultFactory is Authentication {
+contract VaultFactory is ReentrancyGuardTransient, Ownable2Step {
```
elshan-eth, line 21: What are the advantages as opposed to Authentication?

elshan-eth (approved):
LGTM! Only few comments

pkg/vault/contracts/VaultFactory.sol
EndymionJkb: It does seem like create2 would revert the second time, because we're hard-coding the salt to 0. So if it deployed once, then you tried it again with the same salt, it would fail. We'd need a counter for the salt or something to allow it to run multiple times with create2.

pkg/vault/contracts/VaultFactory.sol
```diff
@@ -89,9 +100,9 @@ contract VaultFactory is Authentication {
...
         }
 
-        ProtocolFeeController feeController = new ProtocolFeeController(IVault(vaultAddress));
+        protocolFeeController = new ProtocolFeeController(IVault(vaultAddress));
 
-        VaultAdmin vaultAdmin = VaultAdmin(
+        vaultAdmin = VaultAdmin(
             payable(
                 Create2.deploy(
                     0,
```
EndymionJkb, line 108:
```suggestion
                    0, // ETH value in
                    bytes32(0x00), // salt
```
Could document what these parameters mean

pkg/vault/contracts/VaultFactory.sol
EndymionJkb: Ownable2Step makes sense here, in case the account gets compromised. We'd need to use this again if we ever had to redeploy the Vault.

EndymionJkb (approved):
Also have some questions about the deployment complete - and some OCD notes - but otherwise looks simpler and sufficient.

pkg/vault/test/foundry/VaultFactory.t.sol
```diff
@@ -4,6 +4,8 @@ pragma solidity ^0.8.24;
 
 import "forge-std/Test.sol";
 
+import { Ownable } from "@openzeppelin/contracts/access/Ownable.sol";
+
 import { IAuthentication } from "@balancer-labs/v3-interfaces/contracts/solidity-utils/helpers/IAuthentication.sol";
```
gosuto-inzasheru, line 9: unnecessary import

pkg/vault/contracts/VaultFactory.sol
gosuto-inzasheru: it will only revert on create2 (`CreateCollision`) if passed the same salt.

my understanding is that this boolean thus prevents the factory from deploying more than once

pkg/vault/contracts/VaultFactory.sol
jubeira: it can't be redeployed with the flag anyways. We'd need to redeploy the factory in that case.
We can change the flag to a mapping (salt => boolean) if that was the case. But I think the current approach is fine.

> What are the advantages as opposed to Authentication?

Authentication needs governance to grant the permission to deploy, which makes it operationally more cumbersome.

The deployment depends on the deployer account anyways, which is the one that generates the right address for the vault factory, and in turn generates the right vault address given the correct salt. So now we're saying that besides deploying the factory, the deployer also needs to call `create`.

pkg/vault/test/foundry/VaultFactory.t.sol
jubeira: you need this for the test with the error selector

pkg/vault/contracts/VaultFactory.sol
```diff
@@ -89,13 +112,14 @@ contract VaultFactory is Authentication {
...
+        ProtocolFeeController protocolFeeController_ = new ProtocolFeeController(IVault(vaultAddress));
+        protocolFeeController[vaultAddress] = protocolFeeController_;
 
-        VaultAdmin vaultAdmin = VaultAdmin(
+        VaultAdmin vaultAdmin_ = VaultAdmin(
             payable(
                 Create2.deploy(
-                    0,
-                    bytes32(0x00),
+                    0, // ETH value.
```
EndymionJkb, line 121:
```suggestion
                    0, // ETH value
```

EndymionJkb: nit: inlines don't have periods

pkg/vault/contracts/VaultFactory.sol
```diff
@@ -2,28 +2,38 @@
...
-contract VaultFactory is Authentication {
+contract VaultFactory is ReentrancyGuardTransient, Ownable2Step {
     bytes32 public immutable vaultCreationCodeHash;
     bytes32 public immutable vaultAdminCreationCodeHash;
     bytes32 public immutable vaultExtensionCreationCodeHash;
 
+    mapping(address vaultAddress => ProtocolFeeController) public protocolFeeController;
+    mapping(address vaultAddress => VaultExtension) public vaultExtension;
+    mapping(address vaultAddress => VaultAdmin) public vaultAdmin;
+    mapping(address vaultAddress => bool deployed) public isDeployed;
```
EndymionJkb, line 30:
```suggestion
    mapping(address vaultAddress => ProtocolFeeController) public deployedProtocolFeeControllers;
    mapping(address vaultAddress => VaultExtension) public deployedVaultExtensions;
    mapping(address vaultAddress => VaultAdmin) public deployedVaultAdmins;
    mapping(address vaultAddress => bool deployed) public deployedVaults;
```
Consider making the names clearer. (And then you wouldn't have all the name clashes with `protocolFeeController_`, `vaultExtension_`, etc.)

pkg/vault/contracts/VaultFactory.sol
```diff
@@ -34,21 +44,28 @@ contract VaultFactory is Authentication {
...
-    /// @notice The bytecode for the given contract does not match the expected bytecode.
+    /**
+     * @notice The bytecode for the given contract does not match the expected bytecode.
+     * @param contractName The name of the mismatched contract
+     */
     error InvalidBytecode(string contractName);
 
+    /**
+     * @notice The Vault has already been deployed at this target address.
+     * @param vault Vault address under contention.
```
EndymionJkb, line 55:
```suggestion
     * @param vault Vault address already consumed by a previous deployment
```

pkg/vault/contracts/VaultFactory.sol
```diff
@@ -109,20 +133,22 @@ contract VaultFactory is Authentication {
...
+        vaultAdmin[vaultAddress] = vaultAdmin_;
 
-        VaultExtension vaultExtension = VaultExtension(
+        VaultExtension vaultExtension_ = VaultExtension(
             payable(
                 Create2.deploy(
-                    0,
-                    bytes32(uint256(0x01)),
-                    abi.encodePacked(vaultExtensionCreationCode, abi.encode(vaultAddress, vaultAdmin))
+                    0, // ETH value.
```
EndymionJkb, line 141:
```suggestion
                    0, // ETH value
```

EndymionJkb (approved):
Looks better with the passed in salt. Some minor suggestions (mostly renaming and format). The ultimate would be a fork test, just to make sure there are no (more) surprises; might be overkill?

jubeira:
> The ultimate would be a fork test, just to make sure there are no (more) surprises; might be overkill?

I'm doing that in the deployments repo

PR #1164, opened 2024-12-04
Add note about front-running pool registration and initialization

EndymionJkb:
# Description

Per ToB issue 4, couldn't find any existing explicit documentation of this.

## Type of change

- [ ] Bug fix 
- [ ] New feature 
- [ ] Breaking change 
- [ ] Dependency changes
- [ ] Code refactor / cleanup
- [ ] Other

## Checklist:


joaobrunoah (approved):
LGTM!

pkg/pool-utils/contracts/BasePoolFactory.sol
```diff
@@ -30,6 +30,16 @@ import { SingletonAuthentication } from "@balancer-labs/v3-vault/contracts/Singl
...
+ *
+ * Use of factories is also important for security. Calls to `registerPool` or `initialize` made directly on the Vault
+ * could potentially be frontrun. In the case of registration, a DoS attack could register a pool with malicious
+ * parameters, causing the legitimate registration transaction to fail. The standard Balancer factories avoid this by
+ * deploying and registering in a single `create` function.
+ *
+ * It would also be possible to frontrun `initialize` (e.g., with unbalanced liquidity), and cause the intended
+ * initialization to fail. Like registration, initialization only happens once. While the Balancer standard factories
+ * do not initialize on create, this is a factor to consider when launching new pools. To avoid any possibility of
+ * frontrunning, best practice would be to create (i.e., deploy and register) and initialize in the same transaction.
```
jubeira, line 42:
```suggestion
 * It would also be possible to frontrun `initialize` (e.g., with unbalanced liquidity), and cause the intended
 * initialization to fail. Like registration, initialization only happens once. While the Balancer standard factories
 * do not initialize on create (there's no side effect other than one `intialize` tx failing), this is a factor to consider when launching new pools.
 * To avoid any possibility of frontrunning, best practice would be to create (i.e., deploy and register) and initialize in the same transaction whenever necessary.
```

jubeira, line 42:
```suggestion
 * It would also be possible to frontrun `initialize` (e.g., with unbalanced liquidity), and cause the intended
 * initialization to fail. Like registration, initialization only happens once. While the Balancer standard factories
 * do not initialize on create (there's no side effect other than one `intialize` tx failing), this is a factor to consider when launching new pools.
 * To avoid any possibility of frontrunning, best practice would be to create (i.e., deploy and register) and initialize in the same transaction whenever necessary.
```

jubeira, line 42:
```suggestion
 * It would also be possible to frontrun `initialize` (e.g., with unbalanced liquidity), and cause the intended
 * initialization to fail. Like registration, initialization only happens once. While the Balancer standard factories
 * do not initialize on create (there's no side effect other than one `intialize` tx failing), this is a factor to consider when launching new pools.
 * To avoid any possibility of frontrunning, best practice would be to create (i.e., deploy and register) and initialize in the same transaction whenever necessary.
```

jubeira (approved):
LGTM! Small comment; frontrunning initialize is usually not a problem.

PR #1167, opened 2024-12-05
Medusa swap tests

elshan-eth:
# Description

Add medusa swap tests to check that the invariant never decreases.

## Type of change

- [ ] Bug fix 
- [ ] New feature 
- [ ] Breaking change 
- [ ] Dependency changes
- [ ] Code refactor / cleanup
- [ ] Other

## Checklist:


pkg/vault/test/foundry/fuzz/Swap.medusa.sol
```diff
@@ -0,0 +1,88 @@
...
+    uint256 internal constant MIN_SWAP_AMOUNT = 1e6;
+
+    int256 internal invariantDiff = 0;
+
+    constructor() BaseMedusaTest() {
+        vault.manuallySetSwapFee(address(pool), 0);
+    }
+
+    function optimize_currentInvariant() public view returns (int256) {
+        return -int256(invariantDiff);
```
joaobrunoah, line 28: You could also optimize the negative of the invariant, instead of the diff. Like:

`return -int256(currentInvariant);`

Where currentInvariant is the invariant of the pool in the end of the swap. This way, the system would search for operations that would reduce the size of the invariant, trying to put it closer to 0.

pkg/vault/test/foundry/fuzz/Swap.medusa.sol
```diff
@@ -0,0 +1,88 @@
...
+    }
+
+    function property_currentInvariant() public returns (bool) {
+        return invariantDiff >= 0;
```
joaobrunoah, line 32: This will only measure the invariantDiff in the last swap, so it won't direct the sequence of calls to a sequence that decreases the invariant of the pool. I'd recommend measuring the invariant of the pool in the constructor, and then measuring the last invariant of the pool here, checking if the final invariant is not smaller than the first one. Something like:

```
    uint256 currentInvariant = computeInvariant();
    return currentInvariant >= firstInvariant;
```

pkg/vault/test/foundry/fuzz/Swap.medusa.sol
```diff
@@ -0,0 +1,88 @@
...
+            exactAmountIn,
+            0,
+            MAX_UINT256,
+            false,
+            bytes("")
+        );
+
+        int256 invariantAfter = computeInvariant();
+
+        invariantDiff = invariantAfter - invariantBefore;
```
joaobrunoah, line 67: As I said, I'd store only the currentInvariant, and not the invariantDiff.

joaobrunoah (commented):
Nice! I just commented on the property and optimization, but overall looks good!

pkg/pool-stable/medusa.json
```diff
@@ -8,7 +8,7 @@
     "callSequenceLength": 10,
     "corpusDirectory": "medusa-corpus",
     "coverageEnabled": true,
-    "targetContracts": ["AddAndRemoveLiquidityStableMedusaTest"],
+    "targetContracts": ["AddAndRemoveLiquidityStableMedusaTest", "SwapMedusaTest"],
```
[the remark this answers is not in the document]
elshan-eth: Did you mean we need to have two `medusa.json` settings for each medusa test file?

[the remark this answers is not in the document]
joaobrunoah: No, not at all; the way it is, it's good.

pkg/vault/test/foundry/fuzz/Swap.medusa.sol
```diff
@@ -0,0 +1,88 @@
...
+        return -int256(computeInvariant());
+    }
+
+    function property_currentInvariant() public returns (bool) {
+        int256 currentInvariant = computeInvariant();
+        return currentInvariant >= prevInvariant;
+    }
+
+    function computeSwapExactIn(uint256 tokenIndexIn, uint256 tokenIndexOut, uint256 exactAmountIn) public {
+        prevInvariant = computeInvariant();
```
joaobrunoah, line 38: We don't need to update this one. Just use the first invariant from the test, which is calculated in line 24. Line 24 will be calculated for each of the sequence of calls, so it's ok to use that value, it does not need to be updated.

In the property function, we want to compare the current invariant with the first invariant, to make sure there's no sequence of swaps that can decrease the pool invariant.

pkg/vault/test/foundry/fuzz/Swap.medusa.sol
```diff
@@ -0,0 +1,88 @@
...
+    function optimize_currentInvariant() public view returns (int256) {
+        return -int256(computeInvariant());
+    }
+
+    function property_currentInvariant() public returns (bool) {
+        int256 currentInvariant = computeInvariant();
+        return currentInvariant >= prevInvariant;
+    }
+
+    function computeSwapExactIn(uint256 tokenIndexIn, uint256 tokenIndexOut, uint256 exactAmountIn) public {
```
joaobrunoah, line 37: Will we implement the exact out too? Maybe we could add them together in this file, exact in and exact out.

joaobrunoah (approved):
One small comment, but once that's fixed LGTM

jubeira (approved):
LGTM, I think this is a good starting point! Agree with @joaobrunoah about adding exact out if possible.

pkg/pool-weighted/test/foundry/fuzz/SwapWeighted.medusa.sol
```diff
@@ -0,0 +1,68 @@
...
+                false, // Do not disable unbalanced add/remove liquidity
+                // NOTE: sends a unique salt.
+                bytes32(poolCreationNonce++)
+            )
+        );
+
+        // Cannot set the pool creator directly on a standard Balancer weighted pool factory.
+        vault.manualSetPoolCreator(address(newPool), lp);
+
+        feeController.manualSetPoolCreator(address(newPool), lp);
```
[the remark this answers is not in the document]
EndymionJkb: If we're not checking the accuracy of pool creator fees, I'd say we don't need this. (There's a manual setter because the standard pool factories prohibit having a pool creator.)

EndymionJkb (approved):
Looks good - finally installed medusa and everything ran fine. Note that I fixed a README typo myself (easier than adding a comment about it).

PR #1182, opened 2024-12-16
Add createFactory function in BaseVaultTest

### LINKED ISSUE #1180: Factory is an implicit return value of `createPool` for base pool test
For base pool tests, we need to assign the factory inside `createPool`. This is an implicit return value, and it's not obvious for builders trying to reuse base pool test.

See https://github.com/balancer/balancer-v3-monorepo/blob/main/pkg/pool-weighted/test/foundry/WeightedPool.t.sol#L70 and https://github.com/balancer/balancer-v3-monorepo/blob/main/pkg/vault/test/foundry/utils/BasePoolTest.sol#L56.

Let's make it more explicit. I don't think we need to add the value to `createPool`, but maybe create another method `createFactory` that is specific to base pool tests would be better. Pool specializations will need to override it properly.

elshan-eth:
# Description

This PR adds a `createFactory` function in `BaseVaultTest` to standardize the deploying of factories in tests.

## Type of change

- [ ] Bug fix 
- [ ] New feature 
- [ ] Breaking change 
- [ ] Dependency changes
- [X] Code refactor / cleanup
- [ ] Other

## Checklist:

## Issue Resolution

Closes #1180

### DIFF
--- a/pkg/pool-stable/test/foundry/E2eBatchSwap.t.sol
+++ b/pkg/pool-stable/test/foundry/E2eBatchSwap.t.sol
@@ -57,6 +57,7 @@ contract E2eBatchSwapStableTest is E2eBatchSwapTest, StablePoolContractsDeployer
 
         PoolRoleAccounts memory roleAccounts;
 
+        bytes32 salt = keccak256(abi.encode(label));
         newPool = StablePoolFactory(poolFactory).create(
             name,
             symbol,
@@ -67,7 +68,7 @@ contract E2eBatchSwapStableTest is E2eBatchSwapTest, StablePoolContractsDeployer
             address(0),
             false, // Do not enable donations
             false, // Do not disable unbalanced add/remove liquidity
-            ZERO_BYTES32
+            salt
         );
         vm.label(address(newPool), label);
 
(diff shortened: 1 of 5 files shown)


pkg/pool-hooks/test/foundry/ExitFeeHookExample.t.sol
```diff
@@ -39,6 +42,8 @@ contract ExitFeeHookExampleTest is BaseVaultTest {
         super.setUp();
 
         (daiIdx, usdcIdx) = getSortedIndexes(address(dai), address(usdc));
+
+        poolFactoryMock = PoolFactoryMock(address(vault.getPoolFactoryMock()));
```
joaobrunoah, line 46: Since this test only uses the PoolFactoryMock, maybe it should use the base vault test factory?

joaobrunoah (approved):
Nice! LGTM!

jubeira (approved):
Makes sense; thanks for pushing this to the end! LGTM.

pkg/pool-hooks/test/foundry/ExitFeeHookExample.t.sol
elshan-eth: Good catch! Thanks!

elshan-eth: unfortunately we need this because this contract is used in other tests

PR #1186, opened 2024-12-18
Stable math medusa test

elshan-eth:
# Description

Added  a medusa test for stable mathto check that `computeInvariant(balances) `<= `computeInvariant(balances + deltas)
`

## Type of change

- [ ] Bug fix 
- [ ] New feature 
- [ ] Breaking change 
- [ ] Dependency changes
- [ ] Code refactor / cleanup
- [X] Other

## Checklist:

### DIFF
--- a/pkg/solidity-utils/test/foundry/fuzz/StableMathInvariant.medusa.sol
+++ b/pkg/solidity-utils/test/foundry/fuzz/StableMathInvariant.medusa.sol
@@ -12,6 +12,7 @@ import { InputHelpers } from "@balancer-labs/v3-solidity-utils/contracts/helpers
 
 import { BaseMedusaTest } from "@balancer-labs/v3-vault/test/foundry/utils/BaseMedusaTest.sol";
 import { Rounding } from "@balancer-labs/v3-interfaces/contracts/vault/VaultTypes.sol";
+import { ScalingHelpers } from "@balancer-labs/v3-solidity-utils/contracts/helpers/ScalingHelpers.sol";
 
 import { StableMathMock } from "../../../contracts/test/StableMathMock.sol";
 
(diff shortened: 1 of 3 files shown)


pkg/solidity-utils/test/foundry/fuzz/StableMathInvariant.medusa.sol
```diff
@@ -0,0 +1,85 @@
...
+        }
+
+        uint256[] memory newBalances = new uint256[](tokenCount);
+        ScalingHelpers.copyToArray(balances, newBalances);
+
+        for (uint256 i = 0; i < deltaCount; i++) {
+            uint256 tokenIndex = bound(indexes[i], 0, tokenCount - 1);
+            emit Debug("tokenIndex", tokenIndex);
+
+            uint256 delta = bound(deltas[i], 0, type(uint128).max - newBalances[tokenIndex]);
```
jubeira, line 68: Alright this sounds like a lot, but Medusa should choose small deltas given the optimization params right? Is there any way to confirm this is effectively the case?

/cc @joaobrunoah

pkg/solidity-utils/test/foundry/fuzz/StableMathInvariant.medusa.sol
```diff
@@ -0,0 +1,85 @@
...
+        return currentInvariant;
+    }
+
+    function optimize_InvariantWithDelta() public view returns (int256) {
+        return -invariantWithDelta;
+    }
+
+    function property_currentInvariant() public returns (bool) {
+        return invariantWithDelta >= currentInvariant;
+    }
```
jubeira, line 43: Let's add a few comments here. I understand this makes current invariant higher and new invariant lower.

I don't think we actually need to optimize the current invariant. We just want to start from some point and not make it decrease. Also if we always do new balances, isn't this just a foundry stateless fuzz test really?

jubeira (commented):
Thanks @elshan-eth !

The idea looks mostly good, although I think this is actually a stateless test with the way it's written given that the balances are just reset on every run.
Maybe I'd put the test as is in a regular forge test, and modify this one a bit to make it stateful (e.g. start with one set of balances, and modify them from run to run). We can also include amp updates as part of the stateful test I suppose.

Also looping @joaobrunoah in the conversation.

pkg/solidity-utils/test/foundry/fuzz/StableMathInvariant.medusa.sol
joaobrunoah: Yes, agreed! I believe the invariant test is partially implemented in the StableMath.t.sol (solidity-utils package). Maybe, if there's some case missing in there, it can be added.

But, checking the invariant is not enough, because remove liquidity lowers the invariant. We should check if there isn't a set of operations that combined can decrease the rate of the token. It means, a stateful test that stores the first rate of the pool in the constructor, then optimizes the negative of the current rate and checks the current rate at the end, to make sure the current rate is not lower than the first rate. AddAndRemoveLiquidityStable almost does it all, but does not have the swaps in there. I believe putting the Swaps test with the Add/Remove tests should be enough.

elshan-eth: I think the difference is that the fuzz test just uses random data, but the medusa test will try to increase `currentInvariant` and `invariantWithDelta` decrease. This is the case that we are trying to find, that there is some currentInvariant to which if we add delta, then the invariant will be less than with delta. Am I missing something?

PR #1191, opened 2024-12-19
Encapsulation of initialization variables and minor refactoring of BaseTest

elshan-eth:
# Description

A variable that we initialize before setUp, defaultAccountBalance has been encapsulated in a separate method to be able to get a readable error and prevent uncontrolled changes to this variable.

## Type of change

- [ ] Bug fix 
- [ ] New feature 
- [ ] Breaking change 
- [ ] Dependency changes
- [ ] Code refactor / cleanup
- [ ] Other

## Checklist:

### DIFF
--- a/pkg/solidity-utils/test/foundry/utils/BaseTest.sol
+++ b/pkg/solidity-utils/test/foundry/utils/BaseTest.sol
@@ -88,7 +88,7 @@ abstract contract BaseTest is Test, GasSnapshot {
     }
 
     function setDefaultAccountBalance(uint256 balance) internal {
-        if (_initialized) {
+        if (isInitialized()) {
             revert("Cannot change default account balance after initialization");
         }
         _defaultAccountBalance = balance;
@@ -98,6 +98,7 @@ abstract contract BaseTest is Test, GasSnapshot {
         return _defaultAccountBalance;
     }
 
+    // @dev Returns whether the test has been initialized.
     function isInitialized() internal view returns (bool) {
         return _initialized;
     }
@@ -163,8 +164,8 @@ abstract contract BaseTest is Test, GasSnapshot {
      * wrapped and underlying amounts. For certain tests, we may need to override these rates for simplicity.
      */
     function mockERC4626TokenRates() internal virtual {
-        waDAI.inflateUnderlyingOrWrapped(0, 6 * _defaultAccountBalance);
-        waUSDC.inflateUnderlyingOrWrapped(23 * _defaultAccountBalance, 0);
+        waDAI.inflateUnderlyingOrWrapped(0, 6 * defaultAccountBalance());
+        waUSDC.inflateUnderlyingOrWrapped(23 * defaultAccountBalance(), 0);
     }
 
     function getSortedIndexes(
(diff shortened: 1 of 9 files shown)


pkg/solidity-utils/test/foundry/utils/BaseTest.sol
```diff
@@ -65,16 +67,42 @@ abstract contract BaseTest is Test, GasSnapshot {
...
         if (block.chainid == 31337) {
             // Set block.timestamp to something better than 0
             vm.warp(START_TIMESTAMP);
         }
 
+        _initTokens();
+        _initAccounts();
+
+        // Must mock rates after giving wrapped tokens to users, but before creating pools and initializing buffers.
+        mockERC4626TokenRates();
```
joaobrunoah, line 87: Nice!

joaobrunoah (approved):
LGTM!

pkg/solidity-utils/test/foundry/utils/BaseTest.sol
```diff
@@ -65,16 +67,42 @@ abstract contract BaseTest is Test, GasSnapshot {
...
+    }
+
+    function setDefaultAccountBalance(uint256 balance) internal {
+        if (_initialized) {
```
EndymionJkb, line 91: `isInitialized()` is never called; could call it here, or we don't need it.

pkg/solidity-utils/test/foundry/utils/BaseTest.sol
```diff
@@ -65,16 +67,42 @@ abstract contract BaseTest is Test, GasSnapshot {
...
+            revert("Cannot change default account balance after initialization");
+        }
+        _defaultAccountBalance = balance;
+    }
+
+    function defaultAccountBalance() internal view returns (uint256) {
+        return _defaultAccountBalance;
```
EndymionJkb, line 98: `require` that it be initialized? Could even lazy initialize it, if there's a reasonable default. Like if it's not initialized, set it to a global default and then set _initialized. If a particular test wants to use a different specific value, it can just be sure to call `setDefaultAccountBalance` before anything that references it.

pkg/solidity-utils/test/foundry/utils/BaseTest.sol
```diff
@@ -128,14 +155,16 @@ abstract contract BaseTest is Test, GasSnapshot {
...
+
     /**
      * @notice Manipulate rates of ERC4626 tokens.
      * @dev It's important to not have a 1:1 rate when testing ERC4626 tokens, so we can differentiate between
      * wrapped and underlying amounts. For certain tests, we may need to override these rates for simplicity.
      */
     function mockERC4626TokenRates() internal virtual {
-        waDAI.inflateUnderlyingOrWrapped(0, 6 * defaultBalance);
-        waUSDC.inflateUnderlyingOrWrapped(23 * defaultBalance, 0);
+        waDAI.inflateUnderlyingOrWrapped(0, 6 * _defaultAccountBalance);
```
EndymionJkb, line 166: Consider using the function consistently, instead of the variable directly (as it's internal)?

pkg/solidity-utils/test/foundry/utils/BaseTest.sol
EndymionJkb: Or make a virtual `defaultAccountBalance()` that just returns the standard value, and if tests want something different, they can override it?

EndymionJkb (approved):
LGTM

PR #1193, opened 2024-12-19
Fix rounding of 2CLP pools

joaobrunoah:
# Description

Certora found some rounding issues related to some functions of 2-CLP pool:
1. Function `calculateQuadratic()`, from `Gyro2CLPMath.sol`, was not rounded properly.
2. When calculating virtual offsets, rounding was not applied.

## Type of change

- [x] Bug fix 
- [ ] New feature 
- [ ] Breaking change 
- [ ] Dependency changes
- [ ] Code refactor / cleanup
- [ ] Other

## Checklist:

### DIFF
--- a/pkg/pool-gyro/contracts/Gyro2CLPPool.sol
+++ b/pkg/pool-gyro/contracts/Gyro2CLPPool.sol
@@ -166,14 +166,7 @@ contract Gyro2CLPPool is IGyro2CLPPool, BalancerPoolToken {
 
         (uint256 sqrtAlpha, uint256 sqrtBeta) = _getSqrtAlphaAndBeta();
 
-        // If EXACT_IN, round invariant down to give less tokens to the user. If EXACT_OUT, round invariant up to
-        // charge more tokens from the user.
-        uint256 currentInvariant = Gyro2CLPMath.calculateInvariant(
-            balances,
-            sqrtAlpha,
-            sqrtBeta,
-            kind == SwapKind.EXACT_IN ? Rounding.ROUND_DOWN : Rounding.ROUND_UP
-        );
+        uint256 currentInvariant = Gyro2CLPMath.calculateInvariant(balances, sqrtAlpha, sqrtBeta, Rounding.ROUND_DOWN);
 
         // virtualBalanceIn is always rounded up, because:
         // * If swap is EXACT_IN: a bigger virtualBalanceIn leads to a lower amount out;
--- a/pkg/pool-gyro/contracts/Gyro2CLPPool.sol
+++ b/pkg/pool-gyro/contracts/Gyro2CLPPool.sol
@@ -167,7 +167,7 @@ contract Gyro2CLPPool is IGyro2CLPPool, BalancerPoolToken {
 
         uint256 currentInvariant = Gyro2CLPMath.calculateInvariant(balances, sqrtAlpha, sqrtBeta, rounding);
 
-        uint256[2] memory virtualOffsets = _calculateVirtualOffsets(currentInvariant, sqrtAlpha, sqrtBeta);
+        uint256[2] memory virtualOffsets = _calculateVirtualOffsets(currentInvariant, sqrtAlpha, sqrtBeta, rounding);
 
         virtualBalanceIn = tokenInIsToken0 ? virtualOffsets[0] : virtualOffsets[1];
         virtualBalanceOut = tokenInIsToken0 ? virtualOffsets[1] : virtualOffsets[0];
@@ -177,10 +177,11 @@ contract Gyro2CLPPool is IGyro2CLPPool, BalancerPoolToken {
     function _calculateVirtualOffsets(
         uint256 invariant,
         uint256 sqrtAlpha,
-        uint256 sqrtBeta
+        uint256 sqrtBeta,
+        Rounding rounding
     ) internal view virtual returns (uint256[2] memory virtualOffsets) {
-        virtualOffsets[0] = Gyro2CLPMath.calculateVirtualParameter0(invariant, sqrtBeta);
-        virtualOffsets[1] = Gyro2CLPMath.calculateVirtualParameter1(invariant, sqrtAlpha);
+        virtualOffsets[0] = Gyro2CLPMath.calculateVirtualParameter0(invariant, sqrtBeta, rounding);
+        virtualOffsets[1] = Gyro2CLPMath.calculateVirtualParameter1(invariant, sqrtAlpha, rounding);
         return virtualOffsets;
     }
 
--- a/pkg/pool-gyro/contracts/Gyro2CLPPool.sol
+++ b/pkg/pool-gyro/contracts/Gyro2CLPPool.sol
@@ -116,7 +116,7 @@ contract Gyro2CLPPool is IGyro2CLPPool, BalancerPoolToken {
             balanceTokenInScaled18,
             balanceTokenOutScaled18,
             tokenInIsToken0,
-            request.kind == SwapKind.EXACT_IN ? Rounding.ROUND_DOWN : Rounding.ROUND_UP
+            request.kind
         );
 
         if (request.kind == SwapKind.EXACT_IN) {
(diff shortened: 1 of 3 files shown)


pkg/pool-gyro/contracts/Gyro2CLPPool.sol
```diff
@@ -90,16 +90,24 @@ contract Gyro2CLPPool is IGyro2CLPPool, BalancerPoolToken {
...
 
         if (tokenInIndex == 0) {
             // if newBalance = newX
+
+            // L / sqrt(beta), rounded down to minimize newBalance.
+            uint256 a = invariant.divDown(sqrtBeta);
+            // L * sqrt(alpha), rounded up to minimize newBalance (b is in the denominator).
+            uint256 b = invariant.mulUp(sqrtAlpha);
+
             newBalance = squareNewInv.divUpRaw(b + balancesLiveScaled18[1]) - a;
```
jubeira, line 102: `a` rounded down is fine, but `b` should be rounded down too. We always want to round `newBalance` up, and b is n the denominator so we want a smaller `b`.

pkg/pool-gyro/contracts/Gyro2CLPPool.sol
```diff
@@ -90,16 +90,24 @@ contract Gyro2CLPPool is IGyro2CLPPool, BalancerPoolToken {
...
         } else {
             // if newBalance = newY
+
+            // L / sqrt(beta), rounded up to minimize newBalance (a is in the denominator).
+            uint256 a = invariant.divUp(sqrtBeta);
+            // L * sqrt(alpha), rounded down to minimize newBalance.
+            uint256 b = invariant.mulDown(sqrtAlpha);
```
jubeira, line 109: Same as above; shouldn't `a` be rounded down? I think I agree more with the original code before these changes. Am I missing something?

pkg/pool-gyro/contracts/lib/Gyro2CLPMath.sol
```diff
@@ -116,17 +116,26 @@ library Gyro2CLPMath {
...
-        uint256 denominator = a.mulUp(2 * FixedPoint.ONE);
+        function(uint256, uint256) pure returns (uint256) _mulUpOrDown = rounding == Rounding.ROUND_DOWN
+            ? FixedPoint.mulDown
+            : FixedPoint.mulUp;
+
+        function(uint256, uint256) pure returns (uint256) _divUpOrDown = rounding == Rounding.ROUND_DOWN
+            ? FixedPoint.divDown
+            : FixedPoint.divUp;
+
+        uint256 denominator = 2 * a;
```
jubeira, line 130: ah nice. This result is exact so it doesn't need rounding right?

pkg/pool-gyro/contracts/lib/Gyro2CLPMath.sol
```diff
@@ -116,17 +116,26 @@ library Gyro2CLPMath {
...
         // Order multiplications for fixed point precision.
-        uint256 addTerm = (mc.mulDown(4 * FixedPoint.ONE)).mulDown(a);
+        uint256 addTerm = _mulUpOrDown(4 * mc, a);
```
jubeira, line 132: :+1:

pkg/pool-gyro/contracts/Gyro2CLPPool.sol
jubeira: note that we want `newBalance` to be as high as possible (see comments in `IBasePool`)

jubeira (commented):
Nice, thanks @joaobrunoah !

I don't agree with the changes for `computeBalance`, but otherwise makes sense!

pkg/pool-gyro/contracts/Gyro2CLPPool.sol
joaobrunoah: Ooooff, I inverted the logic. Sure, we should always round newBalance up.

pkg/pool-gyro/contracts/lib/Gyro2CLPMath.sol
joaobrunoah: Exactly

joaobrunoah: Besides, solc 0.8.24 handles overflows and underflows. I believe that was the reason to use mulDown for exact operations in the old library.

pkg/pool-gyro/contracts/Gyro2CLPPool.sol
```diff
@@ -167,7 +168,7 @@ contract Gyro2CLPPool is IGyro2CLPPool, BalancerPoolToken {
 
         uint256 currentInvariant = Gyro2CLPMath.calculateInvariant(balances, sqrtAlpha, sqrtBeta, rounding);
 
-        uint256[2] memory virtualOffsets = _calculateVirtualOffsets(currentInvariant, sqrtAlpha, sqrtBeta);
+        uint256[2] memory virtualOffsets = _calculateVirtualOffsets(currentInvariant, sqrtAlpha, sqrtBeta, rounding);
```
jubeira, line 171: I have some doubts about these.

We're rounding both in the same direction, but then those are used inside swap for `calcInGivenOut` and `calcOutGivenIn`, which need to round the result differently.

joaobrunoah: Fixed!

pkg/pool-gyro/contracts/Gyro2CLPPool.sol
```diff
@@ -158,32 +158,44 @@ contract Gyro2CLPPool is IGyro2CLPPool, BalancerPoolToken {
...
-        virtualBalanceOut = tokenInIsToken0 ? virtualOffsets[1] : virtualOffsets[0];
-    }
+        // If EXACT_IN, round invariant down to give less tokens to the user. If EXACT_OUT, round invariant up to
+        // charge more tokens from the user.
+        uint256 currentInvariant = Gyro2CLPMath.calculateInvariant(
+            balances,
+            sqrtAlpha,
+            sqrtBeta,
+            kind == SwapKind.EXACT_IN ? Rounding.ROUND_DOWN : Rounding.ROUND_UP
+        );
```
jubeira, line 176: The virtual parameters are always proportional to the invariant, and we're rounding one of them up and the other one down/ 

Technically we should send an invariant rounded down to the virtual parameter that we round down, and an invariant rounded up to the parameter that we're rounding up.

I'm not sure it's really worth it though, as you'd need to compute two different invariants. This is only used for swaps, so rounding is not as impactful as with joins / exits.

pkg/pool-gyro/contracts/Gyro2CLPPool.sol
```diff
@@ -158,32 +158,44 @@ contract Gyro2CLPPool is IGyro2CLPPool, BalancerPoolToken {
...
-    ) internal view virtual returns (uint256[2] memory virtualOffsets) {
-        virtualOffsets[0] = Gyro2CLPMath.calculateVirtualParameter0(invariant, sqrtBeta, rounding);
-        virtualOffsets[1] = Gyro2CLPMath.calculateVirtualParameter1(invariant, sqrtAlpha, rounding);
-        return virtualOffsets;
+        // virtualBalanceIn is always rounded up, because:
+        // * If swap is EXACT_IN: a bigger virtualBalanceIn leads to a lower amount out;
+        // * If swap is EXACT_OUT: a bigger virtualBalanceIn leads to a bigger amount in;
+        // virtualBalanceOut is always rounded down, because:
+        // * If swap is EXACT_IN: a lower virtualBalanceOut leads to a lower amount out;
+        // * If swap is EXACT_OUT: a lower virtualBalanceOut leads to a bigger amount in;
```
jubeira, line 183: This looks correct now.

jubeira (commented):
Looks better now! I have a doubt about the invariant rounding direction. Not 100% sure it actually matters in the context of a swap though. Let's try to find out before merging.

pkg/pool-gyro/contracts/Gyro2CLPPool.sol
joaobrunoah: Done, always rounding down now. There's an [issue](https://github.com/balancer/balancer-v3-monorepo/issues/1205) opened so the Gyro team can check this.

jubeira (approved):
LGTM! Let's just not forget to check #1205 before deploying.

elshan-eth (approved):
Nice!

Scutua (approved):

PR #1197, opened 2024-12-20
One stable math fuzz test 

elshan-eth:
# Description

Added a fuzz test for stable math to check that `computeInvariant(balances) <= computeInvariant(balances + deltas)`

## Type of change

- [ ] Bug fix 
- [ ] New feature 
- [ ] Breaking change 
- [ ] Dependency changes
- [ ] Code refactor / cleanup
- [X] Other

## Checklist:

### DIFF
--- a/pkg/solidity-utils/test/foundry/StableMath.t.sol
+++ b/pkg/solidity-utils/test/foundry/StableMath.t.sol
@@ -399,7 +399,7 @@ contract StableMathTest is Test {
         uint256[8] memory deltas,
         uint256[8] memory balancesRaw
     ) public view {
-        _testComputeInvariantLessThenInvariantWithDelta(
+        _testComputeInvariantLessThanInvariantWithDelta(
             amp,
             tokenCount,
             deltaCount,
@@ -410,7 +410,7 @@ contract StableMathTest is Test {
         );
     }
 
-    function _testComputeInvariantLessThenInvariantWithDelta(
+    function _testComputeInvariantLessThanInvariantWithDelta(
         uint256 amp,
         uint256 tokenCount,
         uint256 deltaCount,
@@ -447,8 +447,8 @@ contract StableMathTest is Test {
                 uint256 invariantWithDelta
             ) {
                 assertLe(
-                    currentInvariant - 5,
-                    invariantWithDelta,
+                    currentInvariant,
+                    invariantWithDelta + 5,
                     "Current invariant should be less than or equal to invariant with delta (5 wei tolerance)"
                 );
             } catch {}
--- a/pkg/solidity-utils/test/foundry/StableMath.t.sol
+++ b/pkg/solidity-utils/test/foundry/StableMath.t.sol
@@ -450,8 +450,8 @@ contract StableMathTest is Test {
                     assertApproxEqAbs(
                         currentInvariant,
                         invariantWithDelta,
-                        1,
-                        "Current invariant should be approximately equal to invariant with delta (within 1 wei)"
+                        5,
+                        "Current invariant should be approximately equal to invariant with delta (within 5 wei)"
                     );
                 } else {
                     assertLe(
--- a/pkg/solidity-utils/test/foundry/StableMath.t.sol
+++ b/pkg/solidity-utils/test/foundry/StableMath.t.sol
@@ -372,7 +372,7 @@ contract StableMathTest is Test {
         assertLe(invariantRatioDown, invariantRatioRegular, "Invariant ratio should have gone down");
     }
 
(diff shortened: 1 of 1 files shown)


jubeira (approved):
LGTM, good finding! This makes sense.

We can now try a stateful test that checks the rate in a separate PR.

pkg/solidity-utils/test/foundry/StableMath.t.sol
```diff
@@ -368,4 +371,96 @@ contract StableMathTest is Test {
...
+                        5,
+                        "Current invariant should be approximately equal to invariant with delta (within 5 wei)"
+                    );
+                } else {
+                    assertLe(
+                        currentInvariant,
+                        invariantWithDelta,
+                        "Current invariant should be less than or equal to invariant with delta"
+                    );
+                }
```
joaobrunoah, line 462:
```suggestion
                assertLe(
                    currentInvariant - 5,
                    invariantWithDelta,
                    "Current invariant should be less than or equal to invariant with delta (5 wei tolerance)"
                );
```

This also does the trick, it's shorter, but maybe the current version is clearer.

joaobrunoah (approved):
LGTM!

pkg/solidity-utils/test/foundry/StableMath.t.sol
```diff
@@ -368,4 +371,96 @@ contract StableMathTest is Test {
...
+
+    function testComputeInvariantLessThenInvariantWithLargeDelta__Fuzz(
+        uint256 amp,
+        uint256 tokenCount,
+        uint256 deltaCount,
+        uint256[8] memory indexes,
+        uint256[8] memory deltas,
+        uint256[8] memory balancesRaw
+    ) public view {
+        _testComputeInvariantLessThenInvariantWithDelta(
```
EndymionJkb, line 383:
```suggestion
        _testComputeInvariantLessThanInvariantWithDelta(
```
Here and elsewhere

pkg/solidity-utils/test/foundry/StableMath.t.sol
```diff
@@ -368,4 +371,96 @@ contract StableMathTest is Test {
...
+            balances[i] = bound(balancesRaw[i], 1, type(uint128).max);
+        }
+
+        uint256[] memory newBalances = new uint256[](tokenCount);
+        ScalingHelpers.copyToArray(balances, newBalances);
+
+        for (uint256 i = 0; i < deltaCount; i++) {
+            uint256 tokenIndex = bound(indexes[i], 0, tokenCount - 1);
+            uint256 delta;
+            if (maxDelta > type(uint128).max - newBalances[tokenIndex]) {
```
EndymionJkb, line 437: Consider locally defining `uint256 constant MAX_UINT128 = type(uint128).max` for readability/consistency. Probably not worth deriving from BaseTest vs. Test to inherit it, which would require overriding setup, etc.

pkg/solidity-utils/test/foundry/StableMath.t.sol
```diff
@@ -368,4 +371,96 @@ contract StableMathTest is Test {
...
+        uint256 amp,
+        uint256 tokenCount,
+        uint256 deltaCount,
+        uint256[8] memory indexes,
+        uint256[8] memory deltas,
+        uint256[8] memory balancesRaw,
+        uint256 maxDelta
+    ) internal view {
+        amp = boundAmp(amp);
+        tokenCount = bound(tokenCount, MIN_TOKENS, MAX_TOKENS);
```
EndymionJkb, line 423:
```suggestion
        tokenCount = bound(tokenCount, MIN_TOKENS, StableMath.MAX_STABLE_TOKENS);
```
Could just use this, and not need to define `MAX_TOKENS`

pkg/solidity-utils/test/foundry/StableMath.t.sol
EndymionJkb: I'd say the shorter version is clear enough.

Also, do we need 5 wei? 1 wei works for me. Do lower values fail on CI? (Or it is because we're using 8 tokens, and StableMath is only verified up to 5? Though it seems to work with 1 even with 8 tokens for me.)

EndymionJkb (approved):
I like testing with different token counts too; we don't do that enough. Some comments on naming, variables, and ranges.

pkg/solidity-utils/test/foundry/StableMath.t.sol
jubeira: Try fuzzing 1M values or more; that's where the failure appears with 1 wei.

PR #1201, opened 2024-12-24
Using wrappedToken in CompositeLiquidityRouter in ERC4626Pool operations

### LINKED ISSUE #1192: Fix automatic wraps / unwraps in CLR
The CLR was made considering that pools that have buffers would always wrap / unwrap, and that doesn't seem to be the case in practice. 

E.g. a pool with sDAI might have a buffer for it, but users will still want to operate with sDAI directly. So we might need to specify when to wrap / unwrap (as with the batch router).

elshan-eth:
# Description

The current CLR implementation does not support adding mixed assets. For example, if my pool consists of two wrapped tokens, and I already have one of them and want to use it, this is not possible with the current CLR. The reason is that it treats all input tokens as underlying assets. This PR resolves the issue by introducing a new flag in the interface, allowing us to specify which type of token to use.

## Type of change

- [ ] Bug fix 
- [X] New feature 
- [ ] Breaking change 
- [ ] Dependency changes
- [ ] Code refactor / cleanup
- [ ] Other

## Checklist:

## Issue Resolution

Closes #1192


pkg/vault/contracts/CompositeLiquidityRouter.sol
```diff
@@ -291,6 +320,14 @@ contract CompositeLiquidityRouter is ICompositeLiquidityRouter, BatchRouterCommo
             IERC4626 wrappedToken = IERC4626(address(erc4626PoolTokens[i]));
             IERC20 underlyingToken = IERC20(_vault.getBufferAsset(wrappedToken));
 
+            if (useWrappedTokens[i] == true) {
+                amountsOut[i] = wrappedAmountsOut[i];
+                if (isStaticCall == false) {
+                    _sendTokenOut(params.sender, erc4626PoolTokens[i], amountsOut[i], params.wethIsEth);
+                }
+                continue;
```
jubeira, line 328: I understand we need to apply limits before we continue (like in [here](https://github.com/balancer/balancer-v3-monorepo/pull/1201/files#diff-06fda6cf7b962215183d91acc7ad1adf10fc3260b79cd46e9a0c70a78526c88fL298))

pkg/vault/contracts/CompositeLiquidityRouter.sol
```diff
@@ -347,15 +384,20 @@ contract CompositeLiquidityRouter is ICompositeLiquidityRouter, BatchRouterCommo
             IERC4626 wrappedToken = IERC4626(address(erc4626PoolTokens[i]));
             IERC20 underlyingToken = IERC20(_vault.getBufferAsset(wrappedToken));
 
+            if (useWrappedTokens[i] == true) {
+                finalAmountsIn[i] = amountsIn[i];
+                _takeTokenIn(params.sender, wrappedToken, finalAmountsIn[i], params.wethIsEth);
+                continue;
```
jubeira, line 390: Same as above; we need to apply limits here as in [here](https://github.com/balancer/balancer-v3-monorepo/pull/1201/files#diff-06fda6cf7b962215183d91acc7ad1adf10fc3260b79cd46e9a0c70a78526c88fR397)

pkg/vault/contracts/CompositeLiquidityRouter.sol
```diff
@@ -347,15 +384,20 @@ contract CompositeLiquidityRouter is ICompositeLiquidityRouter, BatchRouterCommo
             IERC4626 wrappedToken = IERC4626(address(erc4626PoolTokens[i]));
             IERC20 underlyingToken = IERC20(_vault.getBufferAsset(wrappedToken));
 
+            if (useWrappedTokens[i] == true) {
+                finalAmountsIn[i] = amountsIn[i];
+                _takeTokenIn(params.sender, wrappedToken, finalAmountsIn[i], params.wethIsEth);
```
jubeira, line 389: Only if not static call?

pkg/vault/contracts/CompositeLiquidityRouter.sol
```diff
@@ -347,15 +384,20 @@ contract CompositeLiquidityRouter is ICompositeLiquidityRouter, BatchRouterCommo
...
+                continue;
+            }
+
             // If the Vault returns address 0 as underlying, it means that the ERC4626 token buffer was not
             // initialized. Thus, the Router treats it as a non-ERC4626 token.
             if (address(underlyingToken) == address(0)) {
```
jubeira, line 395: This shouldn't be accepted anymore.

If the user doesn't specify `useWrappedTokens` it means that they want to wrap. If there's no registered asset, the operation is invalid, so just revert.

In other words: in the `main` branch we're guessing whether to use wrapped or underlying. That should no longer be the case; the user input shall be the truth from now on. If the operation can't be fulfilled as specified, revert instead of guessing.

pkg/vault/contracts/CompositeLiquidityRouter.sol
```diff
@@ -387,16 +431,16 @@ contract CompositeLiquidityRouter is ICompositeLiquidityRouter, BatchRouterCommo
...
 
             if (isStaticCall == false && kind == SwapKind.EXACT_OUT) {
+                bool wethIsEth = params.wethIsEth;
                 // If the SwapKind is EXACT_OUT, the limit of underlying tokens was taken from the user, so the
                 // difference between limit and exact underlying amount needs to be returned to the sender.
-                _sendTokenOut(params.sender, underlyingToken, limits[i] - underlyingAmounts[i], params.wethIsEth);
+                _sendTokenOut(params.sender, underlyingToken, limits[i] - underlyingAmount, wethIsEth);
             }
+
+            finalAmountsIn[i] = kind == SwapKind.EXACT_IN ? wrappedAmount : underlyingAmount;
```
jubeira, line 443: I don't understand why this is like this.

This function always wraps, so isn't `tokenIn` always underlying if we get to this point? Am I missing something?

pkg/vault/contracts/CompositeLiquidityRouter.sol
```diff
@@ -291,6 +320,14 @@ contract CompositeLiquidityRouter is ICompositeLiquidityRouter, BatchRouterCommo
...
+                amountsOut[i] = wrappedAmountsOut[i];
+                if (isStaticCall == false) {
+                    _sendTokenOut(params.sender, erc4626PoolTokens[i], amountsOut[i], params.wethIsEth);
+                }
+                continue;
+            }
+
             // If the Vault returns address 0 as underlying, it means that the ERC4626 token buffer was not
             // initialized. Thus, the Router treats it as a non-ERC4626 token.
             if (address(underlyingToken) == address(0)) {
```
jubeira, line 333: Same; we don't want to guess anymore in this case.

If the user wants underlying and the buffer is not initialized, revert.

jubeira (commented):
First pass done. I didn't get to the tests yet, but I think there are a few things that need to be adjusted before proceeding.

Overall the idea is correct; great job. It's just that some of the edge cases should behave differently now that we're not automatically wrapping / unwrapping.

pkg/vault/contracts/CompositeLiquidityRouter.sol
```diff
@@ -347,55 +375,57 @@ contract CompositeLiquidityRouter is ICompositeLiquidityRouter, BatchRouterCommo
             IERC4626 wrappedToken = IERC4626(address(erc4626PoolTokens[i]));
             IERC20 underlyingToken = IERC20(_vault.getBufferAsset(wrappedToken));
 
-            // If the Vault returns address 0 as underlying, it means that the ERC4626 token buffer was not
-            // initialized. Thus, the Router treats it as a non-ERC4626 token.
-            if (address(underlyingToken) == address(0)) {
+            if (useWrappedTokens[i] == true) {
+                amountsIn[i] = amountsInRaw[i];
```
jubeira, line 379: So, in the current `main` we had an array for underlying amounts in and another one for wrapped amounts in, and we used one _or_ the other.

Now we're merging it into a single one, because we only need one which sounds fine. For exact in, we need wrapped amounts in, and for exact out we need underlying amounts in.

However, if we set these flags to true that doesn't hold anymore. In other words, `amountsIn[i]`:
- Always represents a wrapped amount if `useWrappedTokens[i]` is `true`
- Represents a wrapped amount if `useWrappedTokens[i]` is `false` and the operation is exact in
- Represents an underlying amount if `useWrappedTokens[i]` is `false` and the operation is exact out

This function is only used internally, but we should very clearly document this.
I'd also document this in the interface for add / remove proportional; the meaning of the returned array changes with this PR, as it may contain a mixed bag of underlying / wrapped amounts now.

pkg/vault/contracts/CompositeLiquidityRouter.sol
```diff
@@ -347,55 +375,57 @@ contract CompositeLiquidityRouter is ICompositeLiquidityRouter, BatchRouterCommo
...
-
                 if (isStaticCall == false) {
-                    _takeTokenIn(params.sender, erc4626PoolTokens[i], amountsIn[i], params.wethIsEth);
+                    _takeTokenIn(params.sender, wrappedToken, amountsIn[i], params.wethIsEth);
                 }
+            } else {
+                require(
+                    address(underlyingToken) != address(0),
+                    "CompositeLiquidityRouter: ERC4626 buffer not initialized"
+                );
```
jubeira, line 392: Let's use the `if / revert` pattern as elsewhere.

pkg/vault/contracts/CompositeLiquidityRouter.sol
```diff
@@ -291,52 +320,51 @@ contract CompositeLiquidityRouter is ICompositeLiquidityRouter, BatchRouterCommo
...
+                if (isStaticCall == false) {
+                    _sendTokenOut(params.sender, underlyingToken, amountsOut[i], params.wethIsEth);
+                }
             }
         }
     }
 
     /// @dev Assumes array lengths have been checked externally.
-    function _wrapTokens(
+    function _prepareTokens(
```
jubeira, line 358: I'd just be gigaexplicit with this function name. How about `_wrapTokensIfRequired`?

pkg/vault/contracts/CompositeLiquidityRouter.sol
```diff
@@ -347,55 +375,57 @@ contract CompositeLiquidityRouter is ICompositeLiquidityRouter, BatchRouterCommo
             IERC4626 wrappedToken = IERC4626(address(erc4626PoolTokens[i]));
             IERC20 underlyingToken = IERC20(_vault.getBufferAsset(wrappedToken));
 
-            // If the Vault returns address 0 as underlying, it means that the ERC4626 token buffer was not
-            // initialized. Thus, the Router treats it as a non-ERC4626 token.
-            if (address(underlyingToken) == address(0)) {
+            if (useWrappedTokens[i] == true) {
+                amountsIn[i] = amountsInRaw[i];
+
                 if (amountsIn[i] > params.maxAmountsIn[i]) {
```
jubeira, line 381: One thing that I'm realizing is that we have `maxAmountsIn` and also `limits` here. This is because in add unbalanced we don't really need to apply limits.

But truth is that it doesn't really matter. For add unbalanced we fix the inputs to max, so this check won't fail. In other words, we can get rid of the limits argument. I'll do a draft; that's out of scope of this PR.

jubeira (commented):
Another pass done.

Changes are looking good for the most part. Just a few minor comments around names / docs / style; otherwise seems to be correct. I'll finish covering the tests in one more pass. Great job!

pkg/vault/contracts/CompositeLiquidityRouter.sol
jubeira: hmm that's not the case actually. I'll double check.

jubeira: ok so my confusion here is that limits is actually `minAmountsOut` whenever we need to wrap exact in, and matches `maxAmountsIn` when it's exact out and we need to wrap.

I'm not liking the multiple meanings of things. It's just too tricky to follow. Let's think it through a bit further (this would be out of scope of this PR anyways).

jubeira: Perhaps applying the limit unconditionally at the end is a good start.
But we still need to handle the limits in the wrap operation, which has different meanings when given in / given out.

pkg/vault/contracts/CompositeLiquidityRouter.sol
elshan-eth: If this is relevant with the new implementation, let me know. It seems the code is cleaner and more understandable now.

pkg/interfaces/contracts/vault/ICompositeLiquidityRouter.sol
```diff
@@ -109,6 +117,7 @@ interface ICompositeLiquidityRouter {
      * @notice Queries an `addLiquidityProportionalToERC4626Pool` operation without actually executing it.
      * @dev An "ERC4626 pool" contains IERC4626 yield-bearing tokens (e.g., waDAI).
      * @param pool Address of the liquidity pool
+     * @param useWrappedTokens An array indicating whether the input token is a wrapper or underlying token
```
EndymionJkb, line 120:
```suggestion
     * @param useWrappedTokens An array indicating whether the input token is a wrapped or underlying token
```

pkg/interfaces/contracts/vault/ICompositeLiquidityRouter.sol
```diff
@@ -39,7 +40,8 @@ interface ICompositeLiquidityRouter {
      */
     function addLiquidityUnbalancedToERC4626Pool(
         address pool,
-        uint256[] memory exactUnderlyingAmountsIn,
+        bool[] memory useWrappedTokens,
```
EndymionJkb, line 43: I think `useWrappedTokens` could be a good name. It could be `isWrappedToken` (in the style of `wethIsEth`), but the point of this is to allow having wrapped tokens that are treated as standard tokens, so the flag does indicate whether the token should be "used" as a wrapped token.

However, it seems like the sense of the boolean is reversed from what I would expect. To me "useWrapped" means "use the token as an ERC4626"; i.e., wrap/unwrap. But from the code, it looks like it means the opposite: if the flag is set, consider it a regular token.

Maybe `wrapUnwrapTokens` would be clearer? If true, wrap/unwrap; otherwise, consider it a standard token.

pkg/interfaces/contracts/vault/ICompositeLiquidityRouter.sol
```diff
@@ -126,6 +136,7 @@ interface ICompositeLiquidityRouter {
      * @notice Queries a `removeLiquidityProportionalFromERC4626Pool` operation without actually executing it.
      * @dev An "ERC4626 pool" contains IERC4626 yield-bearing tokens (e.g., waDAI).
      * @param pool Address of the liquidity pool
+     * @param useWrappedTokens An array indicating whether the output token is a wrapper or underlying token
```
EndymionJkb, line 139:
```suggestion
     * @param useWrappedTokens An array indicating whether the output token is a wrapped or underlying token
```

pkg/vault/contracts/CompositeLiquidityRouter.sol
```diff
@@ -2,8 +2,8 @@
 
 pragma solidity ^0.8.24;
 
-import { IERC20 } from "@openzeppelin/contracts/token/ERC20/IERC20.sol";
 import { IPermit2 } from "permit2/src/interfaces/IPermit2.sol";
+import { IERC20 } from "@openzeppelin/contracts/token/ERC20/IERC20.sol";
```
EndymionJkb, line 6: Suggest restoring this. Minor style point, but for main (i.e., not test) contracts, I prefer organizing them in descending length order for readability.

pkg/vault/contracts/CompositeLiquidityRouter.sol
```diff
@@ -291,116 +319,174 @@ contract CompositeLiquidityRouter is ICompositeLiquidityRouter, BatchRouterCommo
...
-            // If the Vault returns address 0 as underlying, it means that the ERC4626 token buffer was not
-            // initialized. Thus, the Router treats it as a non-ERC4626 token.
-            if (address(underlyingToken) == address(0)) {
-                if (wrappedAmountsOut[i] < params.minAmountsOut[i]) {
-                    revert IVaultErrors.AmountOutBelowMin(
-                        erc4626PoolTokens[i],
-                        wrappedAmountsOut[i],
-                        params.minAmountsOut[i]
-                    );
+            if (useWrappedTokens[i] == true) {
```
EndymionJkb, line 322:
```suggestion
            if (useWrappedTokens[i]) {
```
We use `== false` because it's easier to read than `!variable`, but don't need `== true`

pkg/vault/contracts/CompositeLiquidityRouter.sol
```diff
@@ -291,116 +319,174 @@ contract CompositeLiquidityRouter is ICompositeLiquidityRouter, BatchRouterCommo
...
-        underlyingAmounts = new uint256[](poolTokensLength);
-        wrappedAmounts = new uint256[](poolTokensLength);
+        wrappedAmountsIn = new uint256[](poolTokensLength);
 
         bool isStaticCall = EVMCallModeHelpers.isStaticCall();
 
-        // Wrap given underlying tokens for wrapped tokens.
         for (uint256 i = 0; i < poolTokensLength; ++i) {
             // Treat all ERC4626 pool tokens as wrapped. The next step will verify if we can use the wrappedToken as
             // a valid ERC4626.
```
EndymionJkb, line 374:
```suggestion
            // Treat all ERC4626 pool tokens as wrapped. The next step will verify if we can use the wrappedToken as
            // a valid ERC4626. Note that if `useWrappedTokens[i]` is false, we will treat it as a standard token.
```

pkg/vault/contracts/CompositeLiquidityRouter.sol
```diff
@@ -291,116 +319,174 @@ contract CompositeLiquidityRouter is ICompositeLiquidityRouter, BatchRouterCommo
...
             IERC4626 wrappedToken = IERC4626(address(erc4626PoolTokens[i]));
             IERC20 underlyingToken = IERC20(_vault.getBufferAsset(wrappedToken));
 
-            // If the Vault returns address 0 as underlying, it means that the ERC4626 token buffer was not
-            // initialized. Thus, the Router treats it as a non-ERC4626 token.
-            if (address(underlyingToken) == address(0)) {
-                if (amountsIn[i] > params.maxAmountsIn[i]) {
-                    revert IVaultErrors.AmountInAboveMax(erc4626PoolTokens[i], amountsIn[i], params.maxAmountsIn[i]);
+            // if caller wants to use wrapped token, we will use it directly
+            if (useWrappedTokens[i] == true) {
```
EndymionJkb, line 379:
```suggestion
            // Check whether the caller wants to use the token as an ERC4626 (i.e., wrap/unwrap it), or just use it as
            // a standard token.
            if (useWrappedTokens[i]) {
```

pkg/vault/contracts/CompositeLiquidityRouter.sol
```diff
@@ -291,116 +319,174 @@ contract CompositeLiquidityRouter is ICompositeLiquidityRouter, BatchRouterCommo
...
+        bool isStaticCall = EVMCallModeHelpers.isStaticCall();
+
+        for (uint256 i = 0; i < poolTokensLength; ++i) {
+            // Treat all ERC4626 pool tokens as wrapped. The next step will verify if we can use the wrappedToken as
+            // a valid ERC4626.
+            IERC4626 wrappedToken = IERC4626(address(erc4626PoolTokens[i]));
+            IERC20 underlyingToken = IERC20(_vault.getBufferAsset(wrappedToken));
+
+            // if caller wants to use wrapped token, we will use it directly
+            if (useWrappedTokens[i] == true) {
```
EndymionJkb, line 438:
```suggestion
            // Check whether the caller wants to use the token as an ERC4626 (i.e., wrap/unwrap it), or just use it as
            // a standard token.
            if (useWrappedTokens[i]) {
```

pkg/vault/contracts/CompositeLiquidityRouter.sol
```diff
@@ -291,116 +319,174 @@ contract CompositeLiquidityRouter is ICompositeLiquidityRouter, BatchRouterCommo
...
+        bool wethIsEth
+    ) private returns (uint256[] memory amountsIn) {
+        uint256 poolTokensLength = erc4626PoolTokens.length;
+        amountsIn = new uint256[](poolTokensLength);
+
+        bool isStaticCall = EVMCallModeHelpers.isStaticCall();
+
+        for (uint256 i = 0; i < poolTokensLength; ++i) {
+            // Treat all ERC4626 pool tokens as wrapped. The next step will verify if we can use the wrappedToken as
+            // a valid ERC4626.
```
EndymionJkb, line 433:
```suggestion
            // Treat all ERC4626 pool tokens as wrapped. The next step will verify if we can use the wrappedToken as
            // a valid ERC4626. Note that if `useWrappedTokens[i]` is false, we will treat it as a standard token.
```

pkg/vault/contracts/CompositeLiquidityRouter.sol
```diff
@@ -190,44 +211,48 @@ contract CompositeLiquidityRouter is ICompositeLiquidityRouter, BatchRouterCommo
...
     function addLiquidityERC4626PoolUnbalancedHook(
-        AddLiquidityHookParams calldata params
+        AddLiquidityHookParams calldata params,
+        bool[] calldata useWrappedTokens
     ) external nonReentrant onlyVault returns (uint256 bptAmountOut) {
         IERC20[] memory erc4626PoolTokens = _vault.getPoolTokens(params.pool);
         uint256 poolTokensLength = erc4626PoolTokens.length;
 
         // Revert if tokensIn length does not match with maxAmountsIn length.
         InputHelpers.ensureInputLengthMatch(poolTokensLength, params.maxAmountsIn.length);
```
EndymionJkb, line 240:
```suggestion
        // Revert if `tokensIn` length does not match `maxAmountsIn` and `useWrappedTokens`.
        InputHelpers.ensureInputLengthMatch(poolTokensLength, params.maxAmountsIn.length, useWrappedTokens.length);
```

pkg/vault/contracts/CompositeLiquidityRouter.sol
```diff
@@ -236,8 +261,9 @@ contract CompositeLiquidityRouter is ICompositeLiquidityRouter, BatchRouterCommo
     }
 
     function addLiquidityERC4626PoolProportionalHook(
-        AddLiquidityHookParams calldata params
-    ) external nonReentrant onlyVault returns (uint256[] memory underlyingAmountsIn) {
+        AddLiquidityHookParams calldata params,
+        bool[] calldata useWrappedTokens
+    ) external nonReentrant onlyVault returns (uint256[] memory amountsIn) {
         IERC20[] memory erc4626PoolTokens = _vault.getPoolTokens(params.pool);
         uint256 poolTokensLength = erc4626PoolTokens.length;
```
EndymionJkb, line 269: // Revert if `tokensIn` length does not match `maxAmountsIn` and `useWrappedTokens`.
        InputHelpers.ensureInputLengthMatch(poolTokensLength, params.maxAmountsIn.length, useWrappedTokens.length);

pkg/vault/contracts/CompositeLiquidityRouter.sol
```diff
@@ -258,21 +284,23 @@ contract CompositeLiquidityRouter is ICompositeLiquidityRouter, BatchRouterCommo
...
     function removeLiquidityERC4626PoolProportionalHook(
-        RemoveLiquidityHookParams calldata params
-    ) external nonReentrant onlyVault returns (uint256[] memory underlyingAmountsOut) {
+        RemoveLiquidityHookParams calldata params,
+        bool[] calldata useWrappedTokens
+    ) external nonReentrant onlyVault returns (uint256[] memory amountsOut) {
         IERC20[] memory erc4626PoolTokens = _vault.getPoolTokens(params.pool);
         uint256 poolTokensLength = erc4626PoolTokens.length;
-        underlyingAmountsOut = new uint256[](poolTokensLength);
+        amountsOut = new uint256[](poolTokensLength);
```
EndymionJkb, line 304: // Revert if `tokensOut` length does not match `minAmountsOut` and `useWrappedTokens`.
        InputHelpers.ensureInputLengthMatch(poolTokensLength, params.minAmountsOut.length, useWrappedTokens.length);

pkg/vault/contracts/CompositeLiquidityRouter.sol
```diff
@@ -291,116 +319,174 @@ contract CompositeLiquidityRouter is ICompositeLiquidityRouter, BatchRouterCommo
...
-            if (isStaticCall == false) {
-                _sendTokenOut(params.sender, underlyingToken, underlyingAmountsOut[i], params.wethIsEth);
+                if (isStaticCall == false) {
+                    _sendTokenOut(params.sender, underlyingToken, amountsOut[i], params.wethIsEth);
+                }
             }
         }
     }
 
     /// @dev Assumes array lengths have been checked externally.
```
EndymionJkb, line 359:
```suggestion
    /// @dev Assumes array lengths have been checked externally.
```
Make sure we do this (see other comments where this is added).

EndymionJkb (commented):
Looks like a pretty elegant solution (especially to the limits problem). I have a conceptual issue with the `useWrappedTokens` name, and lots of little details.

I also think these PRs need a *lot* more in the description. Why do we need it? What problem does it address? What options did we consider and reject (e.g., in this case, the limits hack, replaced by splitting the wrap/unwrap functions)? 

Otherwise we're not going to remember all this context when we go back and look at this in 6 months or a year.

pkg/interfaces/contracts/vault/ICompositeLiquidityRouter.sol
```diff
@@ -52,7 +54,8 @@ interface ICompositeLiquidityRouter {
      * the "parent" pool, and also make sure limits are set properly.
      *
      * @param pool Address of the liquidity pool
-     * @param maxUnderlyingAmountsIn Maximum amounts of underlying tokens in, sorted in token registration order of
+     * @param isWrappedToken An array indicating whether the input token is a wrappe or underlying token
```
joaobrunoah, line 57:
```suggestion
     * @param isWrappedToken An array indicating whether the input token is a wrapped or underlying token of
```

pkg/interfaces/contracts/vault/ICompositeLiquidityRouter.sol
```diff
@@ -30,7 +30,8 @@ interface ICompositeLiquidityRouter {
      * the "parent" pool, and also make sure limits are set properly.
      *
      * @param pool Address of the liquidity pool
-     * @param exactUnderlyingAmountsIn Exact amounts of underlying tokens in, sorted in token registration order of
+     * @param isWrappedToken An array indicating whether the input token is a wrapped or underlying token
+     * @param exactAmountsIn Exact amounts of underlying/wrapped tokens in, sorted in token registration order
      * wrapped tokens in the pool
```
joaobrunoah, line 35:
```suggestion
     * @param isWrappedToken An array indicating whether the input token is a wrapped or underlying token, 
     * sorted in token registration order of wrapped tokens in the pool
     * @param exactAmountsIn Exact amounts of underlying/wrapped tokens in, sorted in token registration order 
     * of wrapped tokens in the pool
```

Here and elsewhere. No matter if we're using the underlying or wrapped token, we're still receiving values in the order of the wrapped token registration order. And I'd suggest adding an order comment to the isWrappedToken array (maybe the same as exactAmountsIn "sorted in token registration...", or "same order as exactAmountsIn", or something).

pkg/interfaces/contracts/vault/ICompositeLiquidityRouter.sol
```diff
@@ -72,35 +76,39 @@ interface ICompositeLiquidityRouter {
      * @notice Remove proportional amounts of underlying from an ERC4626 pool, burning an exact pool token amount.
      * @dev An "ERC4626 pool" contains IERC4626 yield-bearing tokens (e.g., waDAI).
      * @param pool Address of the liquidity pool
+     * @param isWrappedToken An array indicating whether the output token is a wrapper or underlying token
      * @param exactBptAmountIn Exact amount of pool tokens provided
-     * @param minUnderlyingAmountsOut Minimum amounts of underlying tokens out, sorted in token registration order of
+     * @param minAmountsOut Minimum amounts of underlying tokens out, sorted in token registration order
```
joaobrunoah, line 81:
```suggestion
     * @param minAmountsOut Minimum amounts of underlying tokens out, sorted in token registration order of
```

pkg/interfaces/contracts/vault/ICompositeLiquidityRouter.sol
```diff
@@ -72,35 +76,39 @@ interface ICompositeLiquidityRouter {
      * @notice Remove proportional amounts of underlying from an ERC4626 pool, burning an exact pool token amount.
      * @dev An "ERC4626 pool" contains IERC4626 yield-bearing tokens (e.g., waDAI).
      * @param pool Address of the liquidity pool
+     * @param isWrappedToken An array indicating whether the output token is a wrapper or underlying token
```
joaobrunoah, line 79: Same as before, some comment about the order here would be good

pkg/interfaces/contracts/vault/ICompositeLiquidityRouter.sol
```diff
@@ -72,35 +76,39 @@ interface ICompositeLiquidityRouter {
...
      * @param exactBptAmountIn Exact amount of pool tokens provided
-     * @param minUnderlyingAmountsOut Minimum amounts of underlying tokens out, sorted in token registration order of
+     * @param minAmountsOut Minimum amounts of underlying tokens out, sorted in token registration order
      * wrapped tokens in the pool
      * @param wethIsEth If true, incoming ETH will be wrapped to WETH and outgoing WETH will be unwrapped to ETH
      * @param userData Additional (optional) data required for removing liquidity
-     * @return underlyingAmountsOut Actual amounts of tokens received, sorted in token registration order of wrapped
+     * @return amountsOut Actual amounts of tokens received, sorted in token registration order
```
joaobrunoah, line 85:
```suggestion
     * @return amountsOut Actual amounts of tokens received, sorted in token registration order of wrapped
```

The order is still related to the order of wrapped tokens in the pool, no matter if you're returning underlying or wrapped tokens

pkg/interfaces/contracts/vault/ICompositeLiquidityRouter.sol
```diff
@@ -72,35 +76,39 @@ interface ICompositeLiquidityRouter {
...
     function removeLiquidityProportionalFromERC4626Pool(
         address pool,
+        bool[] memory isWrappedToken,
         uint256 exactBptAmountIn,
-        uint256[] memory minUnderlyingAmountsOut,
+        uint256[] memory minAmountsOut,
         bool wethIsEth,
         bytes memory userData
-    ) external payable returns (uint256[] memory underlyingAmountsOut);
+    ) external payable returns (uint256[] memory amountsOut);
```
joaobrunoah, line 95: Since we're not returning only underlying tokens, I believe it's confusing to don't return an array with the token addresses, so the user could validate if the router returned underlying or wrapped tokens, and which underlying it has returned.

pkg/interfaces/contracts/vault/ICompositeLiquidityRouter.sol
```diff
@@ -72,35 +76,39 @@ interface ICompositeLiquidityRouter {
...
 
     /**
      * @notice Queries an `addLiquidityUnbalancedToERC4626Pool` operation without actually executing it.
      * @dev An "ERC4626 pool" contains IERC4626 yield-bearing tokens (e.g., waDAI).
      * @param pool Address of the liquidity pool
-     * @param exactUnderlyingAmountsIn Exact amounts of underlying tokens in, sorted in token registration order of
+     * @param isWrappedToken An array indicating whether the input token is a wrapped or underlying token
+     * @param exactAmountsIn Exact amounts of underlying/wrapped tokens in, sorted in token registration order
      * wrapped tokens in the pool
```
joaobrunoah, line 103: Same as other comments, clarify about the order of both arrays.

joaobrunoah (commented):
I'd recommend changing `isWrappedToken` to `useWrappedToken`, it's probably clearer. Also, since the amounts out are mixing wrapped and underlying, I'd return a token array, for UX purposes. Not sure if gas is a constraint in this case.

joaobrunoah:
> I'd recommend changing `isWrappedToken` to `useWrappedToken`, it's probably clearer. Also, since the amounts out are mixing wrapped and underlying, I'd return a token array, for UX purposes. Not sure if gas is a constraint in this case.

Ohh, I just noticed it was `useWrappedToken` before. In my opinion, `useWrappedToken` is more intuitive than `isWrappedToken`. Unless you use something like `returnInWrappedTokens` (remove liquidity) and `chargeInWrappedTokens` (add liquidity)

pkg/interfaces/contracts/vault/ICompositeLiquidityRouter.sol
EndymionJkb: Adding "of" at the end doesn't make sense; did you mean to suggest that?

EndymionJkb:
> > I'd recommend changing `isWrappedToken` to `useWrappedToken`, it's probably clearer. Also, since the amounts out are mixing wrapped and underlying, I'd return a token array, for UX purposes. Not sure if gas is a constraint in this case.
> 
> Ohh, I just noticed it was `useWrappedToken` before. In my opinion, `useWrappedToken` is more intuitive than `isWrappedToken`. Unless you use something like `returnInWrappedTokens` (remove liquidity) and `chargeInWrappedTokens` (add liquidity)

I didn't have an issue with the name (also prefer `useWrappedToken`) - it just seemed like the logic was backwards to me. If the flag is true, it's treated like a standard token, whereas to me, "using" a wrapped token means wrapping/unwrapping with it. Maybe reverse it, like `isStandardToken`. If true, it ignores that it's a wrapped token. Also, that way it would default to false, which should be the usual case. Most of these will have wrapped tokens that we want the Vault to wrap/unwrap; the special case we're handling is when we don't want to do that.

EndymionJkb (commented):
I think we need to address the boolean array name (see my suggestion to use `isStandardToken` or `useAsStandardToken`), and returning the token addresses, which would resolve all the "sorted in ... order" comment issues.

elshan-eth:
My eyes are already lost in the description, I double checked several times, but if you can double check, I would be very grateful

pkg/interfaces/contracts/vault/ICompositeLiquidityRouter.sol
```diff
@@ -109,35 +121,41 @@ interface ICompositeLiquidityRouter {
...
      * @param pool Address of the liquidity pool
+     * @param useAsStandardToken An array indicating whether to use the token as standard or wrap/unwrap it,
+     * sorted in token registration order of wrapped tokens in the pool
      * @param exactBptAmountOut Exact amount of pool tokens to be received
      * @param sender The sender passed to the operation. It can influence results (e.g., with user-dependent hooks)
      * @param userData Additional (optional) data required for the query
-     * @return underlyingAmountsIn Expected amounts of tokens to add, sorted in token registration order of wrapped
-     * tokens in the pool
+     * @return tokensIn Expected tokens added in the pool
+     * @return amountsIn Expected amounts of tokens added in the pool
```
EndymionJkb, line 130:
```suggestion
     * @return tokensIn Expected tokens added to the pool
     * @return amountsIn Expected amounts of tokens added to the pool
```

pkg/interfaces/contracts/vault/ICompositeLiquidityRouter.sol
```diff
@@ -109,35 +121,41 @@ interface ICompositeLiquidityRouter {
...
         bytes memory userData
-    ) external returns (uint256[] memory underlyingAmountsIn);
+    ) external returns (address[] memory tokensIn, uint256[] memory amountsIn);
 
     /**
      * @notice Queries a `removeLiquidityProportionalFromERC4626Pool` operation without actually executing it.
      * @dev An "ERC4626 pool" contains IERC4626 yield-bearing tokens (e.g., waDAI).
      * @param pool Address of the liquidity pool
+     * @param useAsStandardToken An array indicating whether to use the token as standard or unwrap it,
+     * sorted in token registration order of wrapped tokens in the pool
```
EndymionJkb, line 145:
```suggestion
     * @param useAsStandardToken Flags indicating whether the corresponding token should be treated as an ERC4626
     * (unwrap and transfer the underlying asset) or as a standard ERC20 (transfer the wrapped token directly)
```

pkg/interfaces/contracts/vault/ICompositeLiquidityRouter.sol
```diff
@@ -196,7 +214,7 @@ interface ICompositeLiquidityRouter {
      * @param tokensOut Output token addresses, sorted by user preference. `tokensOut` array must have all tokens from
      * child pools and all tokens that are not BPTs from the nested pool (parent pool). If not all tokens are informed,
      * balances are not settled and the operation reverts. Tokens that repeat must be informed only once.
-     * @param minAmountsOut Minimum amounts of each outgoing underlying token, sorted according to tokensIn array
+     * @param minAmountsOut Minimum amounts of each token, sorted according to tokensIn array
```
EndymionJkb, line 217:
```suggestion
     * @param minAmountsOut Minimum amounts of each outgoing token, corresponding to `tokensOut`
```

pkg/interfaces/contracts/vault/ICompositeLiquidityRouter.sol
```diff
@@ -196,7 +214,7 @@ interface ICompositeLiquidityRouter {
...
      * @param wethIsEth If true, incoming ETH will be wrapped to WETH and outgoing WETH will be unwrapped to ETH
      * @param userData Additional (optional) data required for the operation
      * @return amountsOut Actual amounts of tokens received, parallel to `tokensOut`
```
EndymionJkb, line 220:
```suggestion
     * @return amountsOut Actual amounts of tokens received, corresponding to `tokensOut`
```
Parallel would work too, but trying to be consistent.

pkg/interfaces/contracts/vault/ICompositeLiquidityRouter.sol
```diff
@@ -152,7 +170,7 @@ interface ICompositeLiquidityRouter {
      * @param parentPool Address of the highest level pool (which contains BPTs of other pools)
      * @param tokensIn Input token addresses, sorted by user preference. `tokensIn` array must have all tokens from
      * child pools and all tokens that are not BPTs from the nested pool (parent pool).
-     * @param exactAmountsIn Amount of each underlying token in, sorted according to tokensIn array
+     * @param exactAmountsIn Amount of each token in, sorted according to tokensIn array
```
EndymionJkb, line 173:
```suggestion
     * @param exactAmountsIn Amount of each token in, corresponding to `tokensIn`
```
This seems more generic, as we might be interpreting tokens differently, such that they're not necessarily sorted

pkg/interfaces/contracts/vault/ICompositeLiquidityRouter.sol
```diff
@@ -172,7 +190,7 @@ interface ICompositeLiquidityRouter {
...
```
EndymionJkb, line 193:
```suggestion
     * @param exactAmountsIn Amount of each token in, corresponding to `tokensIn`
```

pkg/vault/contracts/CompositeLiquidityRouter.sol
```diff
@@ -287,120 +322,187 @@ contract CompositeLiquidityRouter is ICompositeLiquidityRouter, BatchRouterCommo
...
-        wrappedAmounts = new uint256[](poolTokensLength);
+        wrappedAmountsIn = new uint256[](poolTokensLength);
 
         bool isStaticCall = EVMCallModeHelpers.isStaticCall();
 
-        // Wrap given underlying tokens for wrapped tokens.
         for (uint256 i = 0; i < poolTokensLength; ++i) {
             // Treat all ERC4626 pool tokens as wrapped. The next step will verify if we can use the wrappedToken as
-            // a valid ERC4626.
+            // a valid ERC4626. Note that if `useWrappedTokens[i]` is false, we will treat it as a standard token.
```
EndymionJkb, line 384:
```suggestion
            // a valid ERC4626.
```
It's self-documenting now; don't even need a comment.

pkg/vault/contracts/CompositeLiquidityRouter.sol
```diff
@@ -287,120 +322,187 @@ contract CompositeLiquidityRouter is ICompositeLiquidityRouter, BatchRouterCommo
...
+        uint256 poolTokensLength = erc4626PoolTokens.length;
+        amountsIn = new uint256[](poolTokensLength);
+
+        bool isStaticCall = EVMCallModeHelpers.isStaticCall();
+
+        tokensIn = new address[](poolTokensLength);
+
+        for (uint256 i = 0; i < poolTokensLength; ++i) {
+            // Treat all ERC4626 pool tokens as wrapped. The next step will verify if we can use the wrappedToken as
+            // a valid ERC4626. Note that if `useAsStandardToken[i]` is false, we will treat it as a standard token.
```
EndymionJkb, line 446:
```suggestion
            // a valid ERC4626.
```
Don't need the comment (and backwards as written anyway)

EndymionJkb (commented):
Looking much better! Note the suggested language for NatSpec, which I believe is much clearer.

jubeira:
Sorry for the late comment, but I feel we're making up new terms here :sweat_smile: 

We're already dealing with wrapped and underlying, and that should be enough. I can't know which of them is `standard` in that context without reading the docs, which is not ideal.

The argument is consumed by a function called `_wrapTokensExactInIfRequired` or `_wrapTokensExactOutIfRequired`. Then, a more appropriate name for the array of flags would be `wrapUnderlying`. It's pretty clear that if an element is true, you wrap the given underlying amount; if not, you don't.

EndymionJkb:
> The argument is consumed by a function called _wrapTokensExactInIfRequired or _wrapTokensExactOutIfRequired. Then, a > more appropriate name for the array of flags would be wrapUnderlying. It's pretty clear that if an element is true, you wrap > the given underlying amount; if not, you don't.

Makes sense - especially because there isn't an unwrap, so there's only one name for the set of flags. The only drawback is the "default" value would be to set everything to true (kind of annoying), but I agree that reusing standard terms is more important.

jubeira (commented):
This is looking great.

Just a few more comments; otherwise LGTM. I like the new structure much better; it reuses less code, but it's more clear and goes straight to the point.

pkg/vault/contracts/CompositeLiquidityRouter.sol
```diff
@@ -258,21 +287,27 @@ contract CompositeLiquidityRouter is ICompositeLiquidityRouter, BatchRouterCommo
...
+            params.maxAmountsIn,
+            params.wethIsEth
         );
     }
 
     function removeLiquidityERC4626PoolProportionalHook(
-        RemoveLiquidityHookParams calldata params
-    ) external nonReentrant onlyVault returns (uint256[] memory underlyingAmountsOut) {
+        RemoveLiquidityHookParams calldata params,
+        bool[] calldata useAsStandardToken
```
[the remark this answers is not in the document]
EndymionJkb: The opposite of `wrapUnderlying` would be `unwrapWrapped`, so we do need it in both directions, and the default values would be true.

Unless `useAsStandardToken` -> `doNotWrapOrUnwrap`  or `doNotUseAsWrappedToken` or `doNotUseAsERC4626` (defaults to false, single name), or `doNotWrap`, `doNotUnwrap` (defaults to false, two names), without making up new/potentially confusing terms.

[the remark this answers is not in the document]
jubeira: I don't think we have 'default values' anymore. The default in the current version is always wrap / unwrap, but that doesn't mean it has to be that way for this version. Sending `false` arrays is easier than `true` in solidity, but for the SDK I don't think it's a big deal either way.

I don't have a strong opinion about the tone, but I do prefer positive flags (i.e. true == do something as opposed to true == don't do something) from a readability / comprehension perspective.

pkg/vault/test/foundry/CompositeLiquidityRouterERC4626Pool.t.sol
jubeira, line 1: One more thing: we should add a test that hits the limit in the new condition where we don't wrap / unwrap automatically.

pkg/interfaces/contracts/vault/ICompositeLiquidityRouter.sol
```diff
@@ -24,73 +24,71 @@ interface ICompositeLiquidityRouter {
...
+     * @notice Add arbitrary amounts of tokens to an ERC4626 pool through the buffer.
      * @dev An "ERC4626 pool" contains IERC4626 yield-bearing tokens (e.g., waDAI). Ensure that any buffers associated
      * with the wrapped tokens in the ERC4626 pool have been initialized before initializing or adding liquidity to
      * the "parent" pool, and also make sure limits are set properly.
      *
      * @param pool Address of the liquidity pool
-     * @param useAsStandardToken An array indicating whether to use the token as standard or wrap it,
-     * sorted in token registration order of wrapped tokens in the pool
+     * @param wrapUnderlying Flags indicating whether the corresponding token should be wrapped or
+     * use as a standard ERC20
```
EndymionJkb, line 34:
```suggestion
     * used as a standard ERC20
```

pkg/interfaces/contracts/vault/ICompositeLiquidityRouter.sol
```diff
@@ -24,73 +24,71 @@ interface ICompositeLiquidityRouter {
...
+     * @notice Add proportional amounts of tokens to an ERC4626 pool through the buffer.
      * @dev An "ERC4626 pool" contains IERC4626 yield-bearing tokens (e.g., waDAI). Ensure that any buffers associated
      * with the wrapped tokens in the ERC4626 pool have been initialized before initializing or adding liquidity to
      * the "parent" pool, and also make sure limits are set properly.
      *
      * @param pool Address of the liquidity pool
-     * @param useAsStandardToken An array indicating whether to use the token as standard or wrap it,
-     * sorted in token registration order of wrapped tokens in the pool
+     * @param wrapUnderlying Flags indicating whether the corresponding token should be wrapped or
+     * use as a standard ERC20
```
EndymionJkb, line 58:
```suggestion
     * used as a standard ERC20
```

pkg/interfaces/contracts/vault/ICompositeLiquidityRouter.sol
```diff
@@ -24,73 +24,71 @@ interface ICompositeLiquidityRouter {
...
 
     /**
-     * @notice Remove proportional amounts of underlying from an ERC4626 pool, burning an exact pool token amount.
+     * @notice Remove proportional amounts of tokens from an ERC4626 pool, burning an exact pool token amount.
      * @dev An "ERC4626 pool" contains IERC4626 yield-bearing tokens (e.g., waDAI).
      * @param pool Address of the liquidity pool
-     * @param useAsStandardToken An array indicating whether to use the token as standard or unwrap it,
-     * sorted in token registration order of wrapped tokens in the pool
+     * @param unwrapWrapper Flags indicating whether the corresponding token should be unwrapped or
+     * use as a standard ERC20
```
EndymionJkb, line 81:
```suggestion
     * used as a standard ERC20
```

pkg/interfaces/contracts/vault/ICompositeLiquidityRouter.sol
```diff
@@ -24,73 +24,71 @@ interface ICompositeLiquidityRouter {
...
+     * @param minAmountsOut Minimum amounts of each token, corresponding to `tokensOut`
      * @param wethIsEth If true, incoming ETH will be wrapped to WETH and outgoing WETH will be unwrapped to ETH
      * @param userData Additional (optional) data required for removing liquidity
      * @return tokensOut Actual tokens received
      * @return amountsOut Actual amounts of tokens received
      */
     function removeLiquidityProportionalFromERC4626Pool(
         address pool,
-        bool[] memory useAsStandardToken,
+        bool[] memory unwrapWrapper,
```
EndymionJkb, line 91:
```suggestion
        bool[] memory unwrapWrapped,
```
If we're saying "wrapped" and "underlying," this should be `unwrapWrapped`

pkg/interfaces/contracts/vault/ICompositeLiquidityRouter.sol
```diff
@@ -101,17 +99,16 @@ interface ICompositeLiquidityRouter {
      * @notice Queries an `addLiquidityUnbalancedToERC4626Pool` operation without actually executing it.
      * @dev An "ERC4626 pool" contains IERC4626 yield-bearing tokens (e.g., waDAI).
      * @param pool Address of the liquidity pool
-     * @param useAsStandardToken An array indicating whether to use the token as standard or wrap it,
-     * sorted in token registration order of wrapped tokens in the pool
+     * @param wrapUnderlying Flags indicating whether the corresponding token should be wrapped or
+     * use as a standard ERC20
```
EndymionJkb, line 103:
```suggestion
     * used as a standard ERC20
```

pkg/interfaces/contracts/vault/ICompositeLiquidityRouter.sol
```diff
@@ -121,17 +118,17 @@ interface ICompositeLiquidityRouter {
      * @notice Queries an `addLiquidityProportionalToERC4626Pool` operation without actually executing it.
      * @dev An "ERC4626 pool" contains IERC4626 yield-bearing tokens (e.g., waDAI).
      * @param pool Address of the liquidity pool
-     * @param useAsStandardToken An array indicating whether to use the token as standard or wrap/unwrap it,
-     * sorted in token registration order of wrapped tokens in the pool
+     * @param wrapUnderlying Flags indicating whether the corresponding token should be wrapped or
+     * use as a standard ERC20
```
EndymionJkb, line 122:
```suggestion
     * used as a standard ERC20
```

pkg/interfaces/contracts/vault/ICompositeLiquidityRouter.sol
```diff
@@ -141,8 +138,8 @@ interface ICompositeLiquidityRouter {
      * @notice Queries a `removeLiquidityProportionalFromERC4626Pool` operation without actually executing it.
      * @dev An "ERC4626 pool" contains IERC4626 yield-bearing tokens (e.g., waDAI).
      * @param pool Address of the liquidity pool
-     * @param useAsStandardToken An array indicating whether to use the token as standard or unwrap it,
-     * sorted in token registration order of wrapped tokens in the pool
+     * @param unwrapWrapper Flags indicating whether the corresponding token should be unwrapped or
+     * use as a standard ERC20
```
EndymionJkb, line 142:
```suggestion
     * used as a standard ERC20
```

pkg/interfaces/contracts/vault/ICompositeLiquidityRouter.sol
```diff
@@ -151,7 +148,7 @@ interface ICompositeLiquidityRouter {
      */
     function queryRemoveLiquidityProportionalFromERC4626Pool(
         address pool,
-        bool[] memory useAsStandardToken,
+        bool[] memory unwrapWrapper,
```
EndymionJkb, line 151:
```suggestion
        bool[] memory unwrapWrapped,
```

pkg/vault/contracts/CompositeLiquidityRouter.sol
```diff
@@ -114,7 +114,7 @@ contract CompositeLiquidityRouter is ICompositeLiquidityRouter, BatchRouterCommo
     /// @inheritdoc ICompositeLiquidityRouter
     function removeLiquidityProportionalFromERC4626Pool(
         address pool,
-        bool[] memory useAsStandardToken,
+        bool[] memory unwrapWrapper,
```
EndymionJkb, line 117:
```suggestion
        bool[] memory unwrapWrapped,
```

pkg/vault/contracts/CompositeLiquidityRouter.sol
```diff
@@ -134,7 +134,7 @@ contract CompositeLiquidityRouter is ICompositeLiquidityRouter, BatchRouterCommo
                             wethIsEth: wethIsEth,
                             userData: userData
                         }),
-                        useAsStandardToken
+                        unwrapWrapper
```
EndymionJkb, line 137:
```suggestion
                        unwrapWrapped
```

pkg/vault/contracts/CompositeLiquidityRouter.sol
```diff
@@ -303,13 +303,13 @@ contract CompositeLiquidityRouter is ICompositeLiquidityRouter, BatchRouterCommo
 
     function removeLiquidityERC4626PoolProportionalHook(
         RemoveLiquidityHookParams calldata params,
-        bool[] calldata useAsStandardToken
+        bool[] calldata unwrapWrapper
```
EndymionJkb, line 306:
```suggestion
        bool[] calldata unwrapWrapped
```

pkg/vault/contracts/CompositeLiquidityRouter.sol
```diff
@@ -303,13 +303,13 @@ contract CompositeLiquidityRouter is ICompositeLiquidityRouter, BatchRouterCommo
...
     ) external nonReentrant onlyVault returns (address[] memory tokensOut, uint256[] memory amountsOut) {
         IERC20[] memory erc4626PoolTokens = _vault.getPoolTokens(params.pool);
         uint256 poolTokensLength = erc4626PoolTokens.length;
 
-        // Revert if `poolTokens` length does not match `minAmountsOut` and `useAsStandardToken`.
-        InputHelpers.ensureInputLengthMatch(poolTokensLength, params.minAmountsOut.length, useAsStandardToken.length);
+        // Revert if `poolTokens` length does not match `minAmountsOut` and `unwrapWrapper`.
+        InputHelpers.ensureInputLengthMatch(poolTokensLength, params.minAmountsOut.length, unwrapWrapper.length);
```
EndymionJkb, line 312:
```suggestion
        // Revert if `poolTokens` length does not match `minAmountsOut` and `unwrapWrapped`.
        InputHelpers.ensureInputLengthMatch(poolTokensLength, params.minAmountsOut.length, unwrapWrapped.length);
```

pkg/vault/contracts/CompositeLiquidityRouter.sol
```diff
@@ -331,18 +331,7 @@ contract CompositeLiquidityRouter is ICompositeLiquidityRouter, BatchRouterCommo
...
-
-                if (amountsOut[i] < params.minAmountsOut[i]) {
-                    revert IVaultErrors.AmountOutBelowMin(erc4626PoolTokens[i], amountsOut[i], params.minAmountsOut[i]);
-                }
-
-                if (isStaticCall == false) {
-                    _sendTokenOut(params.sender, erc4626PoolTokens[i], amountsOut[i], params.wethIsEth);
-                }
-            } else {
+            if (unwrapWrapper[i]) {
```
EndymionJkb, line 334:
```suggestion
            if (unwrapWrapped[i]) {
```

pkg/vault/test/foundry/CompositeLiquidityRouterERC4626Pool.t.sol
```diff
@@ -205,10 +210,15 @@ contract CompositeLiquidityRouterERC4626PoolTest is BaseERC4626BufferTest {
 
         TestBalances memory balancesBefore = _getTestBalances(alice);
 
+        bool[] memory wrapUnderlying = new bool[](exactUnderlyingAmountsIn.length);
+        for (uint256 i = 0; i < wrapUnderlying.length; i++) {
+            wrapUnderlying[i] = true;
+        }
```
EndymionJkb, line 216: Might be able to refactor this section of code out

EndymionJkb (approved):
Looks good now; have one last (very minor) rename suggestion, and some comments.

joaobrunoah (approved):
LGTM!

jubeira (commented):
A bit late, but for the record the latest changes are looking good. Great job @elshan-eth !

PR #1203, opened 2024-12-27
Refactoring addLiquidityUnbalancedNestedPool in CLR

elshan-eth:
# Description

This refactoring fixes implicit recursion by turning it into explicit recursion. 

## Type of change

- [ ] Bug fix 
- [ ] New feature 
- [ ] Breaking change 
- [ ] Dependency changes
- [X] Code refactor / cleanup
- [ ] Other

## Checklist:


jubeira (commented):
First pass done. Just a few comments, otherwise makes sense.

Other than code specific comments, just a few generic comments:
- I don't think we want to support nesting beyond 1 level. That makes me feel a bit reluctant to use recursion. The implementation here is rather simple, but we've had instances in the past where being too flexible ended up being counterproductive. I'd give it a second thought, maybe just using a function that goes down only one level instead of full blown recursion.
- I get that the case described in the PR title is being covered. What I'd like to see is an explicit test where a pool has mixed tokens (e.g. child pool has sDAI and waUSDT) and the user is able to add liquidity using sDAI (wrapped) and USDT (underlying), which is the case we don't support today (in the current main we must add dai and usdt).

pkg/vault/contracts/CompositeLiquidityRouter.sol
```diff
@@ -530,78 +558,14 @@ contract CompositeLiquidityRouter is ICompositeLiquidityRouter, BatchRouterCommo
...
-            uint256 amountIn = params.maxAmountsIn[i] - _settledTokenAmounts().tGet(tokenIn);
-            // Reset _settledTokensAmount, in case the router is called again in the same transaction.
-            _settledTokenAmounts().tSet(tokenIn, 0);
-            if (amountIn > 0) {
-                _currentSwapTokensIn().add(tokensIn[i]);
-                _currentSwapTokenInAmounts().tSet(tokenIn, amountIn);
+                amountsIn[i] = _wrapAndUpdateTokenInAmounts(IERC4626(childToken), params.sender, params.wethIsEth);
+            } else if (_settledTokenAmounts().tGet(childToken) == 0) {
+                amountsIn[i] = _currentSwapTokenInAmounts().tGet(childToken);
+                _settledTokenAmounts().tSet(childToken, amountsIn[i]);
```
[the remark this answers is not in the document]
joaobrunoah: There are three cases being covered:
* tokens that are pools (we keep adding recursively)
* wrapper case: user wants to add using underlying (buffer is initialized, there is no wrapped amount specified)
* Normal ERC20 tokens

If the buffer is not initialized, that token may or may not be ERC4626, we treat it as a normal ERC20 token.

The `if` is in there to make sure the token amount was not spent in a previous child pool. In the previous solution I modified the token in amounts directly, so the if was not needed, but then I needed to restore the token in amounts array at the end. This solution is better.

pkg/vault/contracts/CompositeLiquidityRouter.sol
```diff
@@ -494,11 +526,9 @@ contract CompositeLiquidityRouter is ICompositeLiquidityRouter, BatchRouterCommo
             if (_vault.isPoolRegistered(childToken)) {
                 // Token is a BPT, so add liquidity to the child pool.
 
-                IERC20[] memory childPoolTokens = _vault.getPoolTokens(childToken);
-                (uint256[] memory childPoolAmountsIn, bool childPoolAmountsEmpty) = _getPoolAmountsIn(
-                    childPoolTokens,
-                    params.sender,
-                    params.wethIsEth
+                (uint256[] memory childPoolAmountsIn, bool childPoolAmountsEmpty) = _addLiquidityRecursive(
```
joaobrunoah, line 529: I'm wondering if there's a drawback in this recursion. It's possible to refer to the same pool twice cyclicly, but that would work, probably (maybe a scenario that should be tested if we proceed with this).

pkg/vault/contracts/CompositeLiquidityRouter.sol
```diff
@@ -642,9 +604,8 @@ contract CompositeLiquidityRouter is ICompositeLiquidityRouter, BatchRouterCommo
             })
         );
 
-        // Remove the underlying amount from `_currentSwapTokenInAmounts` and add the wrapped amount.
+        _currentSwapTokensIn().remove(underlyingToken);
         _currentSwapTokenInAmounts().tSet(underlyingToken, 0);
-        _currentSwapTokenInAmounts().tSet(address(wrappedToken), wrappedAmountOut);
```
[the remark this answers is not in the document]
joaobrunoah: The recursive approach doesn't require to know the amount of the wrapped token, because the amountsIn array is used directly to add liquidity to the parent pool.

I believe we still need a comment to say why we are removing the underlying token here (we paid the wrap in advance)

pkg/vault/test/foundry/CompositeLiquidityRouterNestedPools.t.sol
```diff
@@ -210,21 +210,20 @@ contract CompositeLiquidityRouterNestedPoolsTest is BaseERC4626BufferTest {
         );
 
         // Check ChildPoolB balances.
-        assertApproxEqAbs(
-            vars.childPoolBAfter.dai,
-            vars.childPoolBBefore.dai + amountsIn[vars.daiIdx],
-            MAX_ROUND_ERROR,
-            "ChildPoolB Dai Balance is wrong"
-        );
+        assertEq(vars.childPoolBAfter.dai, vars.childPoolBBefore.dai, "ChildPoolB Dai Balance is wrong");
```
joaobrunoah, line 213: Although this test adds DAI to the top-most pool, that's not always true. DAI address comes before the ChildPoolB BPT address. So, the logic is still the same as before: adds the token to the first pool that needs it that appears in the recursion. The difference is that the parent tokens are checked first.

joaobrunoah (approved):
It mostly LGTM, but I'm worried too with the recursion. I mean, the Vault is safe, but the router can behave unpredictably if there's a cyclic path where the same pool appears multiple times. Maybe it reverts, maybe it adds zero tokens, but what happens with the transient accounting? I believe we'd need to check that case if we want to proceed with the unlimited recursion.

elshan-eth:
@jubeira @joaobrunoah
 
We use almost the same code inside these two functions (`_getPoolAmountsIn` and `addLiquidityUnbalancedNestedPoolHook`), so I think the recursion here will look more obvious. I will add recursion constraints to avoid such flexibility. What do you think about it?

pkg/vault/contracts/CompositeLiquidityRouter.sol
```diff
@@ -482,10 +484,50 @@ contract CompositeLiquidityRouter is ICompositeLiquidityRouter, BatchRouterCommo
...
+            if (params.maxAmountsIn[i] == 0) {
+                continue;
+            }
+
             _currentSwapTokenInAmounts().tSet(tokensIn[i], params.maxAmountsIn[i]);
+            _currentSwapTokensIn().add(tokensIn[i]);
         }
 
-        IERC20[] memory parentPoolTokens = _vault.getPoolTokens(params.pool);
+        (uint256[] memory amountsIn, ) = _addLiquidityRecursive(params.pool, params);
```
EndymionJkb, line 495: At first glance it looks like you could just have `_addLiquidityRecursive` take a single parameter, since the pool is already in the params - but you can't, because sometimes it's the child pool, different from the one in params. Would be good to document that somewhere (maybe where the function's defined).

pkg/vault/contracts/CompositeLiquidityRouter.sol
```diff
@@ -482,10 +484,50 @@ contract CompositeLiquidityRouter is ICompositeLiquidityRouter, BatchRouterCommo
...
+            })
+        );
+
+        // Settle the amounts in.
+        if (isStaticCall == false) {
+            _settlePaths(params.sender, params.wethIsEth);
+        }
+    }
+
+    function _addLiquidityRecursive(
```
EndymionJkb, line 515:
```suggestion
    function _addLiquidity(
```
Rather than overloading _addLiquidityRecursive, consider naming the "outer" one _addLiquidity, leaving the "inner" one (that is actually recursive, with the level argument) as _addLiquidityRecursive.

pkg/vault/contracts/CompositeLiquidityRouter.sol
```diff
@@ -32,6 +32,8 @@ contract CompositeLiquidityRouter is ICompositeLiquidityRouter, BatchRouterCommo
     using TransientEnumerableSet for TransientEnumerableSet.AddressSet;
     using TransientStorageHelpers for *;
 
+    uint256 constant _MAX_LEVEL_IN_NESTED_OPERATIONS = 2;
```
EndymionJkb, line 35:
```suggestion
    // The level is 1-based, so the outermost pool (level 1) can contain a nested BPT, but any pool tokens in the child
    // pool (level 2) will not be expanded further, but simply treated as standard tokens.
    uint256 constant _MAX_LEVEL_IN_NESTED_OPERATIONS = 2;
```
Clarify what 2 levels means (there was some confusion off-line).

pkg/vault/contracts/CompositeLiquidityRouter.sol
```diff
@@ -494,11 +536,19 @@ contract CompositeLiquidityRouter is ICompositeLiquidityRouter, BatchRouterCommo
             if (_vault.isPoolRegistered(childToken)) {
                 // Token is a BPT, so add liquidity to the child pool.
 
-                IERC20[] memory childPoolTokens = _vault.getPoolTokens(childToken);
-                (uint256[] memory childPoolAmountsIn, bool childPoolAmountsEmpty) = _getPoolAmountsIn(
-                    childPoolTokens,
-                    params.sender,
-                    params.wethIsEth
+                if (level > _MAX_LEVEL_IN_NESTED_OPERATIONS) {
+                    // if we have reached the maximum level of nested operations, the token will be calculated as ERC20
```
EndymionJkb, line 540:
```suggestion
                    // If we have exceeded the maximum level of nested operations, the token will be considered a standard ERC20.
```

pkg/vault/contracts/CompositeLiquidityRouter.sol
EndymionJkb: We might also note that the Vault explicitly prohibits "Linear pool-style" recursion, where a pool contains its own BPT, so that case is impossible and we don't need to handle it.

pkg/vault/contracts/CompositeLiquidityRouter.sol
```diff
@@ -530,78 +578,15 @@ contract CompositeLiquidityRouter is ICompositeLiquidityRouter, BatchRouterCommo
                 // The ERC4626 token has a buffer initialized within the Vault. Additionally, since the sender did not
                 // specify an input amount for the wrapped token, the function will wrap the underlying asset and use
                 // the resulting wrapped tokens to add liquidity to the pool.
-                _wrapAndUpdateTokenInAmounts(IERC4626(childToken), params.sender, params.wethIsEth);
+                amountsIn[i] = _wrapAndUpdateTokenInAmounts(IERC4626(childToken), params.sender, params.wethIsEth);
+            } else if (_settledTokenAmounts().tGet(childToken) == 0) {
+                // if this token is ERC20 and the amount was not settled in a previous operation, it should be added
```
EndymionJkb, line 583:
```suggestion
                // Set this token's amountIn if it's a standard token that was not previously settled.
```

pkg/vault/contracts/CompositeLiquidityRouter.sol
```diff
@@ -642,9 +625,10 @@ contract CompositeLiquidityRouter is ICompositeLiquidityRouter, BatchRouterCommo
             })
         );
 
-        // Remove the underlying amount from `_currentSwapTokenInAmounts` and add the wrapped amount.
+        // Remove the underlying token from the tokensIn and set the amount as zero because we took it here.
+        // We will take other tokens at the end of the calculation.
```
EndymionJkb, line 629:
```suggestion
        // Remove the underlying token from `_currentSwapTokensIn` and zero out the amount, as these tokens were paid
        // in advance and wrapped. Remaining tokens will be transferred in at the end of the calculation.
```
Is this correct? I wasn't sure what "taking the token" meant in the original.

EndymionJkb (commented):
I'm not sure I completely follow all this, but tried to clarify some things in the comments.

pkg/vault/contracts/CompositeLiquidityRouter.sol
EndymionJkb: Was my suggestion inaccurate? I still think "took" is ambiguous.

EndymionJkb (approved):
Just one question about a previous comment suggestion.

pkg/vault/contracts/CompositeLiquidityRouter.sol
elshan-eth: Sorry, I misunderstood at first

pkg/vault/contracts/CompositeLiquidityRouter.sol
```diff
@@ -530,78 +585,15 @@ contract CompositeLiquidityRouter is ICompositeLiquidityRouter, BatchRouterCommo
                 // The ERC4626 token has a buffer initialized within the Vault. Additionally, since the sender did not
                 // specify an input amount for the wrapped token, the function will wrap the underlying asset and use
                 // the resulting wrapped tokens to add liquidity to the pool.
-                _wrapAndUpdateTokenInAmounts(IERC4626(childToken), params.sender, params.wethIsEth);
+                amountsIn[i] = _wrapAndUpdateTokenInAmounts(IERC4626(childToken), params.sender, params.wethIsEth);
+            } else if (_settledTokenAmounts().tGet(childToken) == 0) {
+                // Set this token's amountIn if it's a standard token that was not previously settled.
+                amountsIn[i] = _currentSwapTokenInAmounts().tGet(childToken);
+                _settledTokenAmounts().tSet(childToken, amountsIn[i]);
```
jubeira, line 592: We're only using `_settledTokenAmounts` to check whether a given token appears twice in the tree; is that right?
Shouldn't we revert if that happens? We're not adding amounts anywhere.

pkg/vault/contracts/CompositeLiquidityRouter.sol
```diff
@@ -482,23 +486,76 @@ contract CompositeLiquidityRouter is ICompositeLiquidityRouter, BatchRouterCommo
...
-                    params.sender,
-                    params.wethIsEth
+                if (level > _MAX_LEVEL_IN_NESTED_OPERATIONS) {
+                    // If we have exceeded the maximum level of nested operations,
+                    // the token will be considered a standard ERC20.
+                    if (_settledTokenAmounts().tGet(childToken) == 0) {
+                        amountsIn[i] = _currentSwapTokenInAmounts().tGet(childToken);
+                        _settledTokenAmounts().tSet(childToken, amountsIn[i]);
+                    }
+                    continue;
```
jubeira, line 551: Namely if settled token amounts is not 0 here it means that the token already appeared somewhere else. Shouldn't we revert there?

Also I wonder if there's a better way to do that, as settled token amounts means something different in the batch router.

jubeira (commented):
Great job @elshan-eth.

The logic looks correct, although I do have some second thoughts about the long if / else chain inside `_addLiquidityRecursive`. Feels like we're doing way too much in there.

Besides that, just one comment about settled amounts.

PR #1210, opened 2025-01-06
Alternative LBP initialization

jubeira:
# Description

Simplify LBP initialization by using the trusted router.
Note: This will require a new router deployment; the current one does not call `saveSender` on initialize. Also adds missing getters for the router.

Hint for the reviewers: hide whitespace.

## Type of change

- [ ] Bug fix 
- [ ] New feature 
- [x] Breaking change 
- [ ] Dependency changes
- [ ] Code refactor / cleanup
- [ ] Other

## Checklist:

## Issue Resolution

N/A


pkg/pool-weighted/contracts/lbp/LBPool.sol
```diff
@@ -215,9 +215,23 @@ contract LBPool is WeightedPool, Ownable2Step, BaseHooks {
...
         hookFlags.shouldCallBeforeAddLiquidity = true;
     }
 
+    function onBeforeInitialize(uint256[] memory, bytes memory) public view override onlyVault returns (bool success) {
+        // We don't have the router argument here, but with only one trusted router this should be enough considering
+        // `initialize` can only happen once.
+        // If the sender is correct in the trusted router, either everything is fine, or the owner is doing something
+        // else with the trusted router while at the same time giving away the execution to a frontrunner, which
+        // is highly unlikely.
+        // In any case, this is just an extra guardrail to start the pool with the right proportion, and for that
```
EndymionJkb, line 228:
```suggestion
        // In any case, this is just an extra guardrail to start the pool with the correct proportions, and for that
```

pkg/pool-weighted/contracts/lbp/LBPoolFactory.sol
```diff
@@ -189,7 +102,7 @@ contract LBPoolFactory is IPoolVersion, ReentrancyGuardTransient, BasePoolFactor
             pool,
             tokenConfig,
             swapFeePercentage,
-            false, // protocol fee exempt
+            false, // (not) protocol fee exempt
```
EndymionJkb, line 105:
```suggestion
            false, // not exempt from protocol fees
```
To me this reads more naturally

pkg/pool-weighted/contracts/lbp/LBPoolFactory.sol
```diff
@@ -27,30 +22,24 @@ import { LBPool } from "./LBPool.sol";
...
         uint32 pauseWindowDuration,
         string memory factoryVersion,
         string memory poolVersion,
-        address trustedRouter,
-        IPermit2 permit2
+        address trustedRouter
     ) BasePoolFactory(vault, pauseWindowDuration, type(LBPool).creationCode) Version(factoryVersion) {
         _poolVersion = poolVersion;
 
         // LBPools are deployed with a router known to reliably report the originating address on operations.
```
EndymionJkb, line 41: Consider elaborating a bit: e.g., 
// This is used to ensure that only the owner can add liquidity to an LBP (including on initialization).

EndymionJkb (approved):
LGTM! Cleaner than adding a new factory create + initialize.

elshan-eth (approved):
LGTM!

pkg/vault/contracts/test/RouterCommonMock.sol
```diff
@@ -41,14 +41,6 @@ contract RouterCommonMock is RouterCommon {
...
     }
 
-    function getWeth() external view returns (IWETH) {
-        return _weth;
-    }
-
-    function getPermit2() external view returns (IPermit2 permit2) {
-        return _permit2;
-    }
-
```
joaobrunoah, line 51: These are used only in tests, and removing them from here increases gas cost a bit and increases all router sizes. Does it need to be in router common?

joaobrunoah (approved):
I have only one comment, but LGTM! Much simpler than before.

pkg/vault/contracts/test/RouterCommonMock.sol
jubeira: It's a bit annoying, but I realized that initializing a pool in the same tx it is created is cumbersome without this if you want to use the standard router. The problem is that you need to deal with token transfers, and the factory needs to approve `permit2` besides the router. Then, you need to have `permit2` address available in the pool factory.

It's not ideal, but it might become useful at some point in the future. Initializing in the same tx as creation is not the most common use case, but if by any reason the creator doesn't want the possibility of initialization frontrun, it should be doable.

Scutua (approved):

PR #1218, opened 2025-01-09
Limits workaround (not good).

jubeira:
Illustrate difficulty handling limits.


jubeira:
As discussed off-thread, the approach will be splitting `_wrapTokens` into 2 functions.

PR #1219, opened 2025-01-09
Registry of Balancer standard contracts (v2)

EndymionJkb:
# Description

Based on #1206, this is designed to be a more straight-forward implementation that still covers all the use cases, but is less error-prone and without troublesome edge cases.

It maintains name => address, and address => info (i.e., state) mappings, but through guaranteeing uniqueness conceptually acts like a single name => address => info data structure, only split to enable querying by either name or address.

These entries are fixed once created (except that `isActive` can be set to false through deprecation), but it is possible to deregister (delete) a contract, which is the simplest mechanism to fix errors, without complex "replacement" logic.

The "alias" use case is enabled with another mapping, and a single function to add or update an alias. This separates concerns, and prevents possible corruption of the core index (e.g., setting an alias that is the same name as a real contract, which could break things if allowed on the main registry).

## Type of change

- [ ] Bug fix 
- [x] New feature 
- [ ] Breaking change 
- [ ] Dependency changes
- [ ] Code refactor / cleanup
- [ ] Other

## Checklist:


pkg/vault/contracts/BalancerContractRegistry.sol
```diff
@@ -0,0 +1,278 @@
...
+
+        if (existingContract == address(0)) {
+            existingContract = _contractAliases[contractId];
+        }
+
+        if (existingContract != address(0)) {
+            info = _contractInfo[existingContract];
+
+            revert ContractAlreadyRegistered(info.contractType, contractName);
+        }
```
jubeira, line 141: I find this bit weird to understand.

Feels like `registerBalancerContract` is mixing concerns. The source of truth should be the registry and the info; aren't aliases orthogonal to the source of truth? Why are we checking the aliases here?

I would just revert if there's something in the registry for this ID, and if the info says it's registered, and leave the aliases alone.

pkg/vault/contracts/BalancerContractRegistry.sol
```diff
@@ -0,0 +1,278 @@
...
+            contractType: contractType,
+            isRegistered: true,
+            isActive: true
+        });
+
+        emit BalancerContractRegistered(contractType, contractName, contractAddress);
+    }
+
+    /// @inheritdoc IBalancerContractRegistry
+    function deregisterBalancerContract(string memory contractName) external authenticate {
```
jubeira, line 158: `deregister` doesn't care about the aliases, which is much more intuitive.

pkg/vault/contracts/BalancerContractRegistry.sol
```diff
@@ -0,0 +1,278 @@
...
+        bytes32 contractId = _getContractId(contractName);
+        address contractAddress = _contractRegistry[contractId];
+
+        if (contractAddress == address(0)) {
+            revert ContractNotRegistered();
+        }
+
+        ContractInfo memory info = _contractInfo[contractAddress];
+        // This should be impossible: the registry and info mappings must be in sync.
+        require(info.isRegistered, "Registry entry with no contract info");
```
jubeira, line 169: i'd stick to `if` / `revert`, unless there's consensus to change the compiler version for everything and start using `require` with custom errors from now on, but that's a larger change.

I think we'd have used `require` if we had it from the start, but at the point the compiler started supporting it it was already too late to update the vault.

pkg/vault/contracts/BalancerContractRegistry.sol
jubeira: or in any case, the problem would be an existing alias for the same contract ID pointing to another address right?

In other words

```
        // Ensure name isn't already in use (including as an alias).
        bytes32 contractId = _getContractId(contractName);
        address registryContractAddress = _contractRegistry[contractId];

        if (registryContractAddress != address(0)) {
            revert AddressAlreadyInRegistry();
        }

        address aliasContractAddress = _contractAliases[contractId];

        if (aliasContractAddress != contractAddress) {
            revert AliasMismatch(name, aliasContractAddress, contractAddress);
        }
```
(also note that I'm using 2 different variables to cache the address from the registry and from the aliases for more clarity).

jubeira (commented):
Thanks @EndymionJkb, looks good! I have a few comments / questions, but otherwise LGTM.

Bottom line, I'm trying to simplify the management of the mappings as much as possible (3 feels already too much to me, but I understand it's required to use it in the way it's intended to). That being said, the more we can split the source of truth from the aliases, the better.

pkg/vault/contracts/BalancerContractRegistry.sol
EndymionJkb: Do we have a distinction between errors that a user can cause (which would revert with custom errors), and errors that are "invariant violations" that, if triggered, mean something unforeseen has happened, making the state internally inconsistent?

If not, could just remove this. If we still want a custom error here, it would be something like "InconsistentState"

pkg/vault/contracts/BalancerContractRegistry.sol
EndymionJkb: I'm not thrilled about this either, but it seems we have to check this both ways. If you have "Contract A" registered to 0x1 and "Contract B" registered to 0x2, then tried to add an alias "Contract B" pointing to 0x1, it's would be internally inconsistent. That's why adding an alias checks that the name isn't already used.

Similarly, you could do this in the opposite order. If you registered "Contract A" to 0x1, then added an alias called "Contract B" to 0x1, then tried to register "Contract B" to 0x2. If that worked, you would again have a registered name and alias pointing to different things, depending on the order of operations.

I suppose we could not worry about this and just say "registered names are checked first" - so in the first case above, "Contract B" would return 0x2 (ignoring the alias pointing to 0x1), and in the second case, "Contract B" would also point to 0x2, ignoring the alias pointing to 0x1.

I'm ok with that if you think just documenting that behavioral quirk is enough (it shouldn't happen in practice anyway). If we should enforce internal consistency, I can try to factor out the check so that it looks better.

jubeira: > If you have "Contract A" registered to 0x1 and "Contract B" registered to 0x2, then tried to add an alias "Contract B" pointing to 0x1, it would be internally inconsistent. That's why adding an alias checks that the name isn't already used.

This is fine, but it's part of the alias setting / updating mechanism.

> Similarly, you could do this in the opposite order. If you registered "Contract A" to 0x1, then added an alias called "Contract B" to 0x1, then tried to register "Contract B" to 0x2. If that worked, you would again have an identical registered name and alias pointing to different things, depending on the order of operations.

Correct, but that's a problem because the existing alias points to a different address than the one you're registering. Doesn't [this option](https://github.com/balancer/balancer-v3-monorepo/pull/1219/files#r1909343709) address the issue?

pkg/vault/contracts/BalancerContractRegistry.sol
jubeira: Not really.
We've used `require` inside tests because it's convenient to differentiate from assertions, and also because we don't really care about error codes inside tests. But elsewhere we've always used `if / revert` (except in gyro pools, which are a port, and using `if / revert` was an abomination, so we upgraded the contract version).

I'd leave the check, but use `if / revert` instead to continue with the usual pattern.

pkg/vault/contracts/BalancerContractRegistry.sol
```diff
@@ -272,6 +259,25 @@ contract BalancerContractRegistry is IBalancerContractRegistry, SingletonAuthent
...
+        if (includeAliases && existingContract == address(0)) {
+            existingContract = _contractAliases[contractId];
+        }
+
+        if (existingContract != address(0)) {
+            ContractInfo memory info = _contractInfo[existingContract];
+
+            revert ContractAlreadyRegistered(info.contractType, contractName);
+        }
+    }
```
jubeira, line 279: Let's write this inline; I don't think it's worth reusing this piece of code if you need a flag to tweak the behavior inside the function. That just makes the logic harder to follow for a reader that hasn't been following up the changes since the beginning.

Also I'd suggest to just revert right away when you find the first incorrect condition as opposed to continuing.

Just to be more specific:
```
        if (includeAliases && existingContract == address(0)) {
            existingContract = _contractAliases[contractId];
        }

        if (existingContract != address(0)) {
            ContractInfo memory info = _contractInfo[existingContract];

            revert ContractAlreadyRegistered(info.contractType, contractName);
        }
```

Here we're reusing `existingContract` to cache the result from the registry and the aliases, and the `if` condition in one case checks `== address(0)` and in the other one `!= address(0)`. Maybe it's just me, but that makes me feel dumb. See suggestions below.

pkg/vault/contracts/BalancerContractRegistry.sol
```diff
@@ -127,18 +130,7 @@ contract BalancerContractRegistry is IBalancerContractRegistry, SingletonAuthent
...
-        if (existingContract == address(0)) {
-            existingContract = _contractAliases[contractId];
-        }
-
-        if (existingContract != address(0)) {
-            info = _contractInfo[existingContract];
-
-            revert ContractAlreadyRegistered(info.contractType, contractName);
-        }
+        bytes32 contractId = _ensureUniqueName(contractName, true);
```
jubeira, line 134:
```suggestion
        bytes32 contractId = _getContractId(contractName);
        address existingRegistryAddress = _contractRegistry[contractId];

        if (existingRegistryAddress != address(0)) {
            revert ContractAlredyRegistered(contractName, existingRegistryAddress); // Optionally use contractId instead of name
        }
        
        address existingAliasAddress = _contractAliases[contractId];

        if (existingAliasAddress != address(0)) {
            revert ContractAliasAlreadyRegistered(contractName, existingAliasAddress); // Optionally use contract ID
        }
```

So just a small clarification here: your code is correct, but I find the logic harder to follow. I think this is because we're reusing variables in order to get the new contract info once, but I don't think that's very relevant. You can have 2 different error codes that are easy to identify instead, which point to the collision (`existing*Address`) and its identifier (name or ID, as you prefer). And if you write your code like this it's crystal clear what's going on: you find something wrong, you revert instead of "continuing if you don't find something wrong" as in the first `if`:

```
        if (existingContract == address(0)) {
            existingContract = _contractAliases[contractId];
        }
```

jubeira: And I'd apply that same pattern to the aliases, inline.

jubeira (commented):
Sorry for being so annoying :innocent: but I think there's some room to simplify the code.

pkg/vault/contracts/BalancerContractRegistry.sol
jubeira: Late comment: let's always avoid `if`s with more than one condition unless it's very necessary. We don't care about gas or bytecode that much here, so simplicity >> everything in this case.

elshan-eth (approved):
Cool!

jubeira (approved):
LGTM, great job! :rocket:

pkg/interfaces/contracts/vault/IBalancerContractRegistry.sol
```diff
@@ -0,0 +1,224 @@
...
+     * @param isActive If there is an entry, this flag indicates whether it is active or deprecated
+     */
+    struct ContractInfo {
+        ContractType contractType;
+        bool isRegistered;
+        bool isActive;
+    }
+
+    /**
+     * @notice Emitted wen a new contract is registered.
```
joaobrunoah, line 34:
```suggestion
     * @notice Emitted when a new contract is registered.
```

Here and elsewhere

joaobrunoah (approved):
Super cool, LGTM!

PR #1220, opened 2025-01-09
Aggregators router 

elshan-eth:
# Description

Router for performing exchange without built-in payment. Aggregator must pay in advance.

The alternative option is #1221, which I like better.

## Type of change

- [ ] Bug fix 
- [x] New feature 
- [ ] Breaking change 
- [ ] Dependency changes
- [ ] Code refactor / cleanup
- [ ] Other

## Checklist:


elshan-eth:
@jubeira I think that you can make a payment hook to collect payment for exactIn and axactOut. Then the aggregator will not think about when to pay, it will just have to implement the hook

jubeira (commented):
Thanks @elshan-eth, good job!
Just a few minor comments; otherwise LGTM.

I'll think a bit more about `wethIsEth`. I think recycling the existing interface is a good call, even if we don't use native eth wrapping.

We can include that feature in the future, but the problem is that in that case you'll need to forward eth to the router and not the vault. Perhaps that's fine and it's still useful. Will double check.

pkg/vault/test/foundry/AggregatorsRouter.t.sol
```diff
@@ -0,0 +1,108 @@
...
+        aggregatorsRouter.querySwapSingleTokenExactIn(pool, usdc, dai, 1e18, address(this), bytes(""));
+    }
+
+    function testSwapExactInWithoutPayment() public {
+        vm.prank(alice);
+        vm.expectRevert(IVaultErrors.BalanceNotSettled.selector);
+        aggregatorsRouter.swapSingleTokenExactIn(address(pool), usdc, dai, 1e18, 0, MAX_UINT256, false, bytes(""));
+    }
+
+    function testSwapExactIn_Fuzz(uint256 swapAmount) public {
```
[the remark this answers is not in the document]
elshan-eth: I forget about it every time :)

pkg/vault/contracts/AggregatorsRouter.sol
```diff
@@ -0,0 +1,170 @@
...
+    ) external nonReentrant onlyVault returns (uint256) {
+        (uint256 amountCalculated, , ) = _swapHook(params);
+
+        return amountCalculated;
+    }
+
+    /***************************************************************************
+                                       Overrides
+    ***************************************************************************/
+    function permitBatchAndCall(
```
EndymionJkb, line 161:
```suggestion
    ***************************************************************************/

    function permitBatchAndCall(
```
Are we not supporting because it's payable? What about `multicall`? Should we also override and reject that? Or only reject them if msg.value > 0?

If we are reverting, I think `OperationNotSupported` can just be local to the AggregatorRouter (not in the interface), as it's not an error a proper user should get. And I don't think it needs an argument.

EndymionJkb (commented):
I'm not totally sure of the goal here. Is this supposed to be a replacement for the regular Router (in which case it should support the full interface), or is it a special one just used for single-hop swaps?

elshan-eth:
> I'm not totally sure of the goal here. Is this supposed to be a replacement for the regular Router (in which case it should support the full interface), or is it a special one just used for single-hop swaps?

@EndymionJkb This is a specific router designed for aggregators like 1inch and others. They want to use payments via direct transfer instead of approve/permit2.

joaobrunoah (approved):
LGTM!

pkg/vault/contracts/AggregatorRouter.sol
```diff
@@ -0,0 +1,159 @@
...
+import { IAggregatorRouter } from "@balancer-labs/v3-interfaces/contracts/vault/IAggregatorRouter.sol";
+import { IVault } from "@balancer-labs/v3-interfaces/contracts/vault/IVault.sol";
+import "@balancer-labs/v3-interfaces/contracts/vault/VaultTypes.sol";
+
+import { RouterCommon } from "./RouterCommon.sol";
+
+/**
+ * @notice Entrypoint for aggregators who want make swaps without built-in payment logic.
+ * @dev The external API functions unlock the Vault, which calls back into the corresponding hook functions.
+ * These interact with the Vault and settle accounting.
```
EndymionJkb, line 20:
```suggestion
 * These interact with the Vault and settle accounting. This is not a full-featured Router; it only implements
 * `swapSingleTokenExactIn` and the associated query.
```

pkg/vault/contracts/AggregatorRouter.sol
```diff
@@ -0,0 +1,159 @@
...
+import { IWETH } from "@balancer-labs/v3-interfaces/contracts/solidity-utils/misc/IWETH.sol";
+import { IRouter } from "@balancer-labs/v3-interfaces/contracts/vault/IRouter.sol";
+import { IAggregatorRouter } from "@balancer-labs/v3-interfaces/contracts/vault/IAggregatorRouter.sol";
+import { IVault } from "@balancer-labs/v3-interfaces/contracts/vault/IVault.sol";
+import "@balancer-labs/v3-interfaces/contracts/vault/VaultTypes.sol";
+
+import { RouterCommon } from "./RouterCommon.sol";
+
+/**
+ * @notice Entrypoint for aggregators who want make swaps without built-in payment logic.
```
EndymionJkb, line 18:
```suggestion
 * @notice Entrypoint for aggregators who want to swap without the standard permit2 payment logic.
```

pkg/vault/contracts/AggregatorRouter.sol
```diff
@@ -0,0 +1,159 @@
...
+                ),
+                (uint256)
+            );
+    }
+
+    /**
+     * @notice Hook for swaps.
+     * @dev Can only be called by the Vault.
+     * This router expects the caller to pay upfront by sending tokens to the vault directly,
+     * so this call only accounts for the amount that has already been paid, skipping transfers of any kind.
```
EndymionJkb, line 73:
```suggestion
     * @dev Can only be called by the Vault. This router expects the caller to pay upfront by sending tokens to the
     * Vault directly, so this call only accounts for the amount that has already been paid, skipping transfers of
     * any kind (specifically, the permit2 transfers triggered by the standard Router).
     *
```

EndymionJkb (approved):
Looks good; just added some clarifying comments.

PR #1234, opened 2025-01-17
Basic Protocol Fee Sweeper

### LINKED ISSUE #1215: Fee sweeper: basics
Create a contract that collects and withdraws fees from the vault.

This should be enough to start with. The method shall be permissioned; the idea is to do this only when a certain threshold is reached, which requires offchain computation.

See [here](https://github.com/balancer/v3_fee_sweeper) for reference

EndymionJkb:
# Description

This is a helper contract used in conjunction with "Burners" (e.g., CoWBurner) to, after collection from the Vault, withdraw protocol fees from the ProtocolFeeController, convert them to a preferred token (e.g., USDC), and forward them to a recipient (e.g., a multi-sig tied to the treasury).

This is a "basic" version, without a real burner, that simply forwards all withdrawn fee tokens to the designated recipient (and emitting an event for the direct transfer).

I included some infrastructure and a mock burner as a basis for further iterations. If you set a target token and use the mock burner, it will simulate a burn by consuming the fee token and minting the target token to the recipient, emitting an event like the real burner would. Since the burners operate asynchronously, these events might come out in any order, and the sweeper contract could have temporary token balances while the burners are operating.

One thing we might want to add in future iterations is validation of the burners. Governance has to approve sweepers (by granting permission to withdraw protocol fees), and must also grant permission to set the burner contract address. If we wanted to decrease the trust level here, we could use the `BalancerContractRegistry` (perhaps with a new type) to register valid burners, and only allow adding vetted ones.

There is also a failsafe withdrawal in case tokens get "stuck" in the sweeper due to burner issues. (They could also get stuck in the burner contracts, which have a similar mechanism.) Since this is the basic version, it simply withdraws the full balance of the given set of tokens, which assumes the sweeper is no longer operating.

For instance, revoking a sweeper's permission to call `withdrawProtocolFees` would prevent it from accumulating any further tokens, and assuming there are no ongoing burner transactions, it would be safe to withdraw the full balances. If we need to support recovery *during* operations, withdrawing the full balances would not be correct, as some of the balance might reflect ongoing burns that would succeed, in addition to failed ones.

One strategy to deal with that would be to emit an event when the sweeper initiates a burn: e.g., BurnOrderEntered(burnNonce++, feeToken, feeTokenAmount), while also passing the nonce to the burner, which could include it in its "order completed" event. That way, an off-chain process could match up the nonces, identify any failed burns, and calculate exactly how many and which tokens were stuck in the sweeper. We could then pass in an array of amounts along with the array of tokens, and withdraw only the stuck ones without removing any tokens needed for ongoing burns. (Maybe 0 could mean "withdraw the whole balance," so `recoverProtocolFees(tokens, new uint256[](tokens.length))` would behave the same as the current version with no value inputs.

Whether this is needed or makes sense depends on requirements and how the burners actually work. Just something to keep in mind for future versions.

## Type of change

- [ ] Bug fix 
- [X] New feature 
- [ ] Breaking change 
- [ ] Dependency changes
- [ ] Code refactor / cleanup
- [ ] Other

## Checklist:

- [X] The diff is legible and has no extraneous changes (there's one extraneous change: adding ZERO_ADDRESS to BaseTest)

## Issue Resolution

Closes #1215

### DIFF
--- a/pkg/interfaces/contracts/standalone-utils/IProtocolFeeBurner.sol
+++ b/pkg/interfaces/contracts/standalone-utils/IProtocolFeeBurner.sol
@@ -6,9 +6,9 @@ import { IERC20 } from "@openzeppelin/contracts/token/ERC20/IERC20.sol";
 
 interface IProtocolFeeBurner {
     function burn(
-        IERC20 tokenIn,
-        uint256 tokenInAmount,
-        IERC20 tokenOut,
+        IERC20 feeToken,
+        uint256 feeTokenAmount,
+        IERC20 targetToken,
         address recipient
-    ) external returns (uint256 tokenOutAmount);
+    ) external returns (uint256 targetTokenAmount);
 }
--- a/pkg/interfaces/contracts/standalone-utils/IProtocolFeeBurner.sol
+++ b/pkg/interfaces/contracts/standalone-utils/IProtocolFeeBurner.sol
@@ -5,7 +5,25 @@ pragma solidity ^0.8.24;
 import { IERC20 } from "@openzeppelin/contracts/token/ERC20/IERC20.sol";
 
 interface IProtocolFeeBurner {
+    event ProtocolFeeBurned(
+        address indexed pool,
+        IERC20 indexed feeToken,
+        uint256 feeTokenAmount,
+        IERC20 indexed targetToken,
+        uint256 targetTokenAmount,
+        address recipient
+    );
+
+    /**
+     * @notice Swap an exact amount of `feeToken` for the `targetToken`, and send proceeds to the `recipient`.
+     * @param pool The pool the fees came from (only used for documentation in the event)
+     * @param feeToken The feeToken collected from the pool
+     * @param feeTokenAmount The number of fee tokens collected
+     * @param targetToken The desired target token (token out of the swap)
+     * @param recipient The recipient of the swap proceeds
+     */
     function burn(
+        address pool,
         IERC20 feeToken,
         uint256 feeTokenAmount,
         IERC20 targetToken,
--- a/pkg/interfaces/contracts/standalone-utils/IProtocolFeeBurner.sol
+++ b/pkg/interfaces/contracts/standalone-utils/IProtocolFeeBurner.sol
@@ -16,6 +16,7 @@ interface IProtocolFeeBurner {
 
     /**
      * @notice Swap an exact amount of `feeToken` for the `targetToken`, and send proceeds to the `recipient`.
+     * @dev Assumes the sweeper has transferred the tokens to the burner prior to the call.
      * @param pool The pool the fees came from (only used for documentation in the event)
      * @param feeToken The feeToken collected from the pool
      * @param feeTokenAmount The number of fee tokens collected
--- a/pkg/interfaces/contracts/standalone-utils/IProtocolFeeBurner.sol
+++ b/pkg/interfaces/contracts/standalone-utils/IProtocolFeeBurner.sol
@@ -5,5 +5,10 @@ pragma solidity ^0.8.24;
 import { IERC20 } from "@openzeppelin/contracts/token/ERC20/IERC20.sol";
 
 interface IProtocolFeeBurner {
-    function burn(IERC20 tokenIn, uint256 tokenInAmount, IERC20 tokenOut, address recipient) external returns(uint256 tokenOutAmount);
+    function burn(
+        IERC20 tokenIn,
+        uint256 tokenInAmount,
+        IERC20 tokenOut,
+        address recipient
+    ) external returns (uint256 tokenOutAmount);
 }
--- a/pkg/interfaces/contracts/standalone-utils/IProtocolFeeBurner.sol
+++ b/pkg/interfaces/contracts/standalone-utils/IProtocolFeeBurner.sol
@@ -10,5 +10,5 @@ interface IProtocolFeeBurner {
         uint256 feeTokenAmount,
         IERC20 targetToken,
         address recipient
-    ) external returns (uint256 targetTokenAmount);
+    ) external;
 }
--- a/pkg/interfaces/contracts/standalone-utils/IProtocolFeeBurner.sol
+++ b/pkg/interfaces/contracts/standalone-utils/IProtocolFeeBurner.sol
@@ -1,0 +1,9 @@
+// SPDX-License-Identifier: GPL-3.0-or-later
+
+pragma solidity ^0.8.24;
+
+import { IERC20 } from "@openzeppelin/contracts/token/ERC20/IERC20.sol";
+
+interface IProtocolFeeBurner {
+    function burn(IERC20 tokenIn, uint256 tokenInAmount, IERC20 tokenOut, address recipient) external returns(uint256 tokenOutAmount);
+}
--- a/pkg/interfaces/contracts/standalone-utils/IProtocolFeeSweeper.sol
+++ b/pkg/interfaces/contracts/standalone-utils/IProtocolFeeSweeper.sol
@@ -34,12 +34,7 @@ interface IProtocolFeeSweeper {
      * @param feeTokenAmount The number of feeTokens
      * @param recipient The recipient of the target tokens
      */
-    event ProtocolFeeSwept(
-        address indexed pool,
-        IERC20 indexed feeToken,
-        uint256 feeTokenAmount,
-        address recipient
-    );
+    event ProtocolFeeSwept(address indexed pool, IERC20 indexed feeToken, uint256 feeTokenAmount, address recipient);
 
     /**
      * @notice Withdraw, convert, and forward protocol fees for a given pool.
@@ -55,7 +50,7 @@ interface IProtocolFeeSweeper {
      * @dev It is not immutable in the Vault, so we need to get it every time.
      * @return protocolFeeController The address of the current `ProtocolFeeController`
      */
-    function getProtocolFeeController() public view returns (IProtocolFeeController);
+    function getProtocolFeeController() external view returns (IProtocolFeeController);
 
     /**
      * @notice Getter for the target token.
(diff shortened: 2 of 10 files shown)


pkg/solidity-utils/test/foundry/utils/BaseTest.sol
```diff
@@ -31,6 +31,8 @@ abstract contract BaseTest is Test, GasSnapshot {
     bytes32 internal constant ZERO_BYTES32 = 0x0000000000000000000000000000000000000000000000000000000000000000;
     bytes32 internal constant ONE_BYTES32 = 0x0000000000000000000000000000000000000000000000000000000000000001;
 
+    address internal constant ZERO_ADDRESS = address(0);
```
joaobrunoah, line 34: Nice!

pkg/interfaces/contracts/standalone-utils/IProtocolFeeBurner.sol
```diff
@@ -0,0 +1,32 @@
+// SPDX-License-Identifier: GPL-3.0-or-later
+
+pragma solidity ^0.8.24;
+
+import { IERC20 } from "@openzeppelin/contracts/token/ERC20/IERC20.sol";
+
+interface IProtocolFeeBurner {
+    event ProtocolFeeBurned(
```
joaobrunoah, line 8: Not sure if we want to document this event, or if the documentation of the `burn()` function is enough

pkg/standalone-utils/contracts/ProtocolFeeSweeper.sol
```diff
@@ -0,0 +1,200 @@
...
+            if (withdrawnTokenBalance == 0) {
+                continue;
+            }
+
+            // If this is already the target token (or we haven't enabled burning), just forward directly.
+            if (canBurn && feeToken != targetToken) {
+                // Allow the burner to withdraw tokens from this contract.
+                feeToken.forceApprove(address(feeBurner), withdrawnTokenBalance);
+                // This is asynchronous; the burner will complete the action and emit an event.
+                feeBurner.burn(pool, feeToken, withdrawnTokenBalance, targetToken, recipient);
```
joaobrunoah, line 88: Maybe it's okay as is, but leaving to the burner the responsibility of dealing with a failed "burn" won't allow us to fallback this to a different burner in the future.

Maybe we should think of a way to return these tokens to the protocol fee sweeper if the burner has failed and try using a different burner.

I am mentioning this here because this solution would change the sweeper's design, since only receiving new tokens and forwarding to a burner wouldn't be enough.

pkg/standalone-utils/contracts/test/ProtocolFeeBurnerMock.sol
```diff
@@ -0,0 +1,37 @@
...
+    /// @inheritdoc IProtocolFeeBurner
+    function burn(
+        address pool,
+        IERC20 feeToken,
+        uint256 feeTokenAmount,
+        IERC20 targetToken,
+        address recipient
+    ) external {
+        // Just emit the event, simulating the tokens being exchanged at 1-to-1.
+        feeToken.safeTransferFrom(address(_feeSweeper), address(this), feeTokenAmount);
```
joaobrunoah, line 31:
```suggestion
        feeToken.safeTransferFrom(msg.sender, address(this), feeTokenAmount);
```

If the burner is generically enough, msg.sender should be enough to get the fee sweeper's address. Then the burner wouldn't need to know about the _feeSweeper address.

joaobrunoah (dismissed):
LGTM!

pkg/standalone-utils/contracts/ProtocolFeeSweeper.sol
```diff
@@ -0,0 +1,200 @@
...
+            IERC20 feeToken = poolTokens[i];
+            uint256 withdrawnTokenBalance = feeToken.balanceOf(address(this)) - existingBalances[i];
+
+            // If no balance, nothing to do.
+            if (withdrawnTokenBalance == 0) {
+                continue;
+            }
+
+            // If this is already the target token (or we haven't enabled burning), just forward directly.
+            if (canBurn && feeToken != targetToken) {
```
elshan-eth, line 84: I think it should be split into two functions. One that calls burn, and one that does safeTransfer. It will be more convenient when we integrate CowSwapBurner. WDYT?

elshan-eth (approved):
LGTM!

pkg/interfaces/contracts/standalone-utils/IProtocolFeeBurner.sol
EndymionJkb: It depends on the burner design - this would be emitted by the burner, not the sweeper, since burning is asynchronous. It depends on whether the burner will internally track and retry ones that fail (maybe ultimately just forwarding them to the recipient). Or whether the tokens will be sent back to the sweeper (and perhaps recovered directly, in which case we'd want to know the amounts in the event, etc.) There are a lot of ways we could design this. Could also be tracked off-chain with events (e.g., by matching up nonces, as noted in the description).

pkg/solidity-utils/test/foundry/utils/BaseTest.sol
EndymionJkb: Yes, sorry for introducing an extraneous thing, but this is an obvious thing to define, which was being defined at the test level in a few other places.

pkg/standalone-utils/contracts/ProtocolFeeSweeper.sol
EndymionJkb: The "burn" one would still have to fall back to forwarding if the contract/target were undefined, though. If we want to collect and forward even though we've assigned a burner, I suppose we could - but if this is being called by governance, at that point they could just call withdraw directly and not have to transfer tokens twice.

The safeTransfer mode is just a fallback in case the burner's undefined. In future PRs we will likely have a choice of burners, or a mapping of pools to burners, or something like that, to make it more flexible. At that point, "no entry" in the map for a pool would mean don't use a burner and just forward it.

We're only constrained by the CowSwapBurner if we're directly reusing it. If we're rewriting it, we can make things simpler.

pkg/standalone-utils/contracts/ProtocolFeeSweeper.sol
EndymionJkb: As we discussed in the meeting, future PRs will likely have a map of pool->burner or something like that (and maybe a sweep call that takes a burner argument). One case would be the burner can't handle it (i.e., has no paths to do swaps), and another would be some kind of market failure, like the burner knows how to handle it, but there's no liquidity.

This is why the burner is in charge of pulling tokens. If it can't handle this pool, it could just leave the tokens there. Maybe the burner could return a boolean indicating whether it accepted or rejected the order. If it rejected it, the sweeper could retry (based on some internal logic we'd have to design). If it accepted it, the burner would pull the tokens, and then it would be the burner's problem.

The burner could retry a few times (e.g., in cases of low liquidity), before giving up, at which point it could either forward the fee tokens to the recipient, or return them to the sweeper, maybe emitting an event that something off-chain could detect, and then recall the sweeper with a different burner.

There are lots of ways we could do it. This is just the first iteration, meant to be as simple as possible.

pkg/standalone-utils/contracts/test/ProtocolFeeBurnerMock.sol
EndymionJkb: We could do this, but it means anyone could use these burners to do cow swap conversions from their own wallet. Is there a possible DoS attack from hundreds of accounts all calling these burner contracts?

joaobrunoah: The way it's implemented, if the sweeper has tokens in the balance, has given the approval and the burn has failed for some reason, anyone could call burn and take the tokens to himself by using the recipient (since burn is async). Using msg.sender prevents that.

joaobrunoah (commented):
I just noticed that the way sweeper is implemented allows to steal tokens if the burn() function fails. The sweeper gives approval but cannot revoke if the burn fails, and anyone can call the burner with a different recipient and use the allowance to steal tokens for himself later.

We could solve that by changing the burner, but I believe we should solve this in the level of sweeper, so the burner developer wouldn't need to worry about this allowance issue.

jubeira (commented):
Great work @EndymionJkb, very clean! Just a few minor comments, but otherwise LGTM!

pkg/standalone-utils/contracts/test/ProtocolFeeBurnerMock.sol
EndymionJkb: Hanging approvals... have caused issues before. Thanks! Good catch.

pkg/interfaces/contracts/standalone-utils/IProtocolFeeSweeper.sol
```diff
@@ -0,0 +1,107 @@
...
+    /**
+     * @notice Withdraw, convert, and forward protocol fees for a given pool.
+     * @dev This will withdraw all fee tokens from the controller to this contract, and attempt to convert and forward
+     * them to the fee recipient. There is also a single token pool withdrawal, but that is an edge case not in scope
+     * for the sweeper. Note that this requires governance to grant this contract permission to call
+     * `withdrawProtocolFees` on the `ProtocolFeeController`.
+     *
+     * @param pool The pool that incurred the fees we're withdrawing
+     */
+    function sweepProtocolFees(address pool) external;
```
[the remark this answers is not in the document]
EndymionJkb: I originally had a `sweepProtocolFees` and `sweepProtocolFeesByToken`, but removed the second one to keep it simple. Should I restore it then?

joaobrunoah (approved):
LGTM!

pkg/standalone-utils/contracts/ProtocolFeeSweeper.sol
```diff
@@ -0,0 +1,200 @@
...
+    }
+
+    function _setTargetToken(IERC20 targetToken) internal {
+        _targetToken = targetToken;
+
+        emit TargetTokenSet(targetToken);
+    }
+
+    /// @inheritdoc IProtocolFeeSweeper
+    function recoverProtocolFees(IERC20[] memory feeTokens) external {
```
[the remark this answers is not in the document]
EndymionJkb: I was thinking of that too, but didn't really see a need for it. But sure, could make the two suggested ones non-reentrant; doesn't hurt anything. There could be cases we're not seeing.

jubeira (approved):
Awesome work @EndymionJkb :100:

PR #1240, opened 2025-01-20
Add cow router type

EndymionJkb:
# Description

According to the CoW design doc, we will need a registry of CoW routers, requiring a new COW_ROUTER type. This is potentially ambiguous with the ROUTER type, so the simplest change would be to rename ROUTER to BALANCER_ROUTER, and add the new type.

`isTrustedRouter` is unchanged for now, assuming it won't be ambiguous in the context of regular hooks and pools, and we don't also need a shortcut for CoW. If we did, or think the name is now ambiguous, we could have `isTrustedBalancerRouter` and `isTrustedCowRouter`. Or just rename `isTrustedRouter` to `isTrustedBalancerRouter`.

## Type of change

- [ ] Bug fix 
- [X] New feature 
- [ ] Breaking change 
- [ ] Dependency changes
- [ ] Code refactor / cleanup
- [ ] Other

## Checklist:

### DIFF
--- a/pkg/interfaces/contracts/vault/IBalancerContractRegistry.sol
+++ b/pkg/interfaces/contracts/vault/IBalancerContractRegistry.sol
@@ -6,7 +6,8 @@ pragma solidity ^0.8.24;
 enum ContractType {
     OTHER, // a blank entry will have a 0-value type, and it's safest to return this in that case
     POOL_FACTORY,
-    ROUTER,
+    BALANCER_ROUTER,
+    COW_ROUTER,
     HOOK,
     ERC4626
 }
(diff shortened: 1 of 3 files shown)


joaobrunoah (approved):
Makes sense!

jubeira:
I'm not against this, but I can see this pattern becoming a scope creep as we add more protocols and partners.

To reduce ambiguity, why not keep these registered in the cow pool factory? That way we isolate this issue within cow-specific contracts. Thoughts?

EndymionJkb:
> I'm not against this, but I can see this pattern becoming a scope creep as we add more protocols and partners.
> 
> To reduce ambiguity, why not keep these registered in the cow pool factory? That way we isolate this issue within cow-specific contracts. Thoughts?

That would be better, yes. I guess wait for the factory and see if we can handle it there - and if so close this. We have an "Other" category as a catch-all already too.

PR #1250, opened 2025-01-22
Add events for initial aggregate pool fees

### LINKED ISSUE #1247: Pool fee events
Hello,

I'm a bit puzzled how pool fee events are supposed to work.

In order to properly track pool balances on swap and liquidity events, the aggregate swap fee percentage needs to be known. This is initialized to a default value from the ProtocolFeeController and may be updated later on. Updates to both the default values and the pool specific value emit proper events, so these are not a problem.

However if the pool is created with protocolFeeExempt set, I can't see a way to properly retrieve this information. It is not included in the pool registration event, not can it be retrieved from the protocol fee controller in a reliable fashion. While it does internally store this information in the isOverride flag, this can be changed later on if the fee is updated. The end result is that there is no way to reliably retrieve the fee information for newly created pools unless running an archive node.

I hope I'm missing something here, maybe someone can shed some light on how this is supposed to work, with the goal being able to track pool balances.

EndymionJkb:
# Description

Add events to track the initial aggregate fee percentages for pools. Otherwise, there would be no indication (off-chain) that a pool was registered as protocol fee exempt.

## Type of change

- [ ] Bug fix 
- [X] New feature 
- [ ] Breaking change 
- [ ] Dependency changes
- [ ] Code refactor / cleanup
- [ ] Other

## Checklist:

## Issue Resolution

Resolves #1247


jubeira:
Thanks @EndymionJkb; this looks good.

Adding @mendesfabio  to the loop.

joaobrunoah (approved):
LGTM! We don't set the aggregate fee percentages after initialization, so using the same events for the initial value and the setters doesn't make sense.

jubeira (approved):
LGTM!

I've added the flag too. It's a bit redundant, but I don't see how it can hurt.

PR #1245, opened 2025-01-21
CoW AMM Router

joaobrunoah:
# Description

Create CoW AMM Router with `swapExactInAndDonateSurplus`, `swapExactOutAndDonateSurplus` and `donate` features.

## Type of change

- [ ] Bug fix 
- [x] New feature 
- [ ] Breaking change 
- [ ] Dependency changes
- [ ] Code refactor / cleanup
- [ ] Optimization: [ ] gas / [ ] bytecode
- [ ] Documentation or wording changes
- [ ] Other

## Checklist:

- [x] The diff is legible and has no extraneous changes
- [x] Complex code has been commented, including external interfaces
- [x] Tests have 100% code coverage
- [x] The base branch is either `main`, or there's a description of how to merge

## Issue Resolution

Closes #1241

PR #1252, opened 2025-01-22
Cow AMM Router: withdrawProtocolFees()

### LINKED ISSUE #1242: Implement withdrawProtocolFees in CoW AMM Router
Depends on #1241 

## Contract Variables to add (all variables are internal)

- `address feeSweeper` : Address of contract that will receive the protocol fees when `withdrawProtocolFees` is called. This is important to make `withdrawProtocolFees` permissionless, and reduce the risk of informing the wrong receiver of protocol fees.

### getFeeSweeper

- return:
    - `feeSweeper`

### setFeeSweeper

- permissioned
- arguments:
    - `address newFeeSweeper` : Address of the contract that will receive fees when `withdrawProtocolFees` is called
- Logic:
    - `feeSweeper` = `newFeeSweeper` (nothing to validate, except the permission)
    - Emit event `FeeSweeperChanged` with new fee sweeper address

### withdrawProtocolFees

Withdraw protocol fees of a given `token` to `feeSweeper` contract.

- permissionless
- arguments
    - `address token` : Token which will be withdrawn
- Logic
    - Transfer `protocolFees[token]` to `feeSweeper` contract
    - `protocolFees[token]` = 0
    - Emit event `CoWProtocolFeeWithdrawn`
        - `address token`
        - `address feeSweeper`
        - `uint256 amount`

joaobrunoah:
# Description

This PR implements withdrawProtocolFees() in the Cow AMM Router. To do so, the router also needs to have a feeSweeper address, which has a permissioned setter.

This PR depends on #1245 .

## Type of change

- [ ] Bug fix 
- [x] New feature 
- [ ] Breaking change 
- [ ] Dependency changes
- [ ] Code refactor / cleanup
- [ ] Other

## Checklist:

## Issue Resolution

Closes #1242


pkg/pool-cow/contracts/CowRouter.sol
```diff
@@ -28,11 +28,18 @@ contract CowRouter is SingletonAuthentication, VaultGuard, ICowRouter {
...
     mapping(IERC20 => uint256) internal _protocolFees;
 
-    constructor(IVault vault, uint256 protocolFeePercentage) VaultGuard(vault) SingletonAuthentication(vault) {
+    constructor(
+        IVault vault,
+        uint256 protocolFeePercentage,
+        address feeSweeper
+    ) VaultGuard(vault) SingletonAuthentication(vault) {
         _protocolFeePercentage = protocolFeePercentage;
+        _feeSweeper = feeSweeper;
```
EndymionJkb, line 42: Since there is validation and an event tied to this, would use the pattern of making a `_setFeeSweeper` and calling it here and from the external `setFeeSweeper`.

Same thing with the protocol fee percentage: `_setProtocolFeePercentage` called from both places.

pkg/pool-cow/contracts/CowRouter.sol
```diff
@@ -60,6 +72,16 @@ contract CowRouter is SingletonAuthentication, VaultGuard, ICowRouter {
         emit ProtocolFeePercentageChanged(newProtocolFeePercentage);
     }
 
+    /// @inheritdoc ICowRouter
+    function setFeeSweeper(address newFeeSweeper) external authenticate {
+        if (newFeeSweeper == address(0) || newFeeSweeper == address(this)) {
```
EndymionJkb, line 77: As noted above, do we need to check for our own address here, or just zero? (In which case we wouldn't need the address argument to the error.)

pkg/pool-cow/contracts/CowRouter.sol
```diff
@@ -145,6 +167,23 @@ contract CowRouter is SingletonAuthentication, VaultGuard, ICowRouter {
...
+    /// @inheritdoc ICowRouter
+    function withdrawProtocolFees(IERC20 token) external {
+        uint256 amountToWithdraw = _protocolFees[token];
+        if (amountToWithdraw == 0) {
+            revert NoProtocolFeesToWithdraw(token);
+        }
+
+        token.safeTransfer(_feeSweeper, amountToWithdraw);
+
+        _protocolFees[token] = 0;
```
EndymionJkb, line 182:
```suggestion
        _protocolFees[token] = 0;
        
        token.safeTransfer(_feeSweeper, amountToWithdraw);
```
Classic reentrancy protection :)

pkg/pool-cow/contracts/CowRouter.sol
```diff
@@ -145,6 +167,23 @@ contract CowRouter is SingletonAuthentication, VaultGuard, ICowRouter {
         );
     }
 
+    /********************************************************
+                     Withdraw Protocol Fees
+    ********************************************************/
+    /// @inheritdoc ICowRouter
+    function withdrawProtocolFees(IERC20 token) external {
+        uint256 amountToWithdraw = _protocolFees[token];
+        if (amountToWithdraw == 0) {
```
EndymionJkb, line 176: Do we really need this? Would be simpler to not revert, and simply have: if(amountToWithdraw > 0) { <code> } and let it fall through and do nothing in the zero balance case.

EndymionJkb (commented):
Mostly language stuff, but also some structural comments.

pkg/pool-cow/contracts/CowRouter.sol
```diff
@@ -28,11 +28,18 @@ contract CowRouter is SingletonAuthentication, VaultGuard, ICowRouter {
     // Protocol fee percentage capped at 10%.
     uint256 internal constant _MAX_PROTOCOL_FEE_PERCENTAGE = 10e16;
 
+    // Address that will receive the protocol fees on withdraw.
```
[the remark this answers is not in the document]
joaobrunoah: The fee sweeper is not in the critical path of the swap or add/removeLiquidity. I think it's worth it to keep the setter for it and not make it immutable since the fee sweeper contract has a considerable chance of changing. Besides, since the router collects fees, it wouldn't be that simple to change it. It's referred as a trusted router by all CoW AMM Pools and the CoW Factory, it's a bit painful.

pkg/pool-cow/contracts/CowRouter.sol
```diff
@@ -149,6 +175,19 @@ contract CowRouter is SingletonAuthentication, VaultGuard, ICowRouter {
         );
     }
 
+    /********************************************************
+                     Withdraw Protocol Fees
+    ********************************************************/
+    /// @inheritdoc ICowRouter
```
EndymionJkb, line 181:
```suggestion
    ********************************************************/

    /// @inheritdoc ICowRouter
```
Blank line after block headers

pkg/pool-cow/test/foundry/CowRouter.t.sol
```diff
@@ -604,6 +604,110 @@ contract CowRouterTest is BaseCowTest {
...
+
+        vm.prank(admin);
+        cowRouter.setFeeSweeper(newFeeSweeper);
+
+        assertEq(cowRouter.getFeeSweeper(), newFeeSweeper, "Fee sweeper was not set properly");
+    }
+
+    /********************************************************
+                          setFeeSweeper()
+    ********************************************************/
```
EndymionJkb, line 623:
```suggestion
```

pkg/pool-cow/test/foundry/CowRouter.t.sol
```diff
@@ -604,6 +604,110 @@ contract CowRouterTest is BaseCowTest {
         assertEq(cowRouter.getProtocolFeePercentage(), newProtocolFeePercentage, "Protocol Fee Percentage is not set");
     }
 
+    /********************************************************
+                          getFeeSweeper()
+    ********************************************************/
+    function testGetFeeSweeper() public {
```
EndymionJkb, line 610:
```suggestion
                          Fee Sweeper
    ********************************************************/
    function testGetFeeSweeper() public {
```
Even in tests, there should be a blank line after section headers (here and elsewhere). I'd just say "Fee Sweeper" here if it's a header; we don't need headers for every function, just every section.

pkg/pool-cow/test/foundry/CowRouter.t.sol
```diff
@@ -604,6 +604,110 @@ contract CowRouterTest is BaseCowTest {
...
+
+        vm.prank(admin);
+        vm.expectEmit();
+        emit ICowRouter.FeeSweeperChanged(newFeeSweeper);
+        cowRouter.setFeeSweeper(newFeeSweeper);
+
+        assertEq(cowRouter.getFeeSweeper(), newFeeSweeper, "Fee sweeper was set properly");
+    }
+
+    /********************************************************
```
EndymionJkb, line 647: Probably don't need a section header here, as there's only one function (unless you're going to add more later?). Actually, it would be a good idea to keep the section header (something like "Withdraw Protocol Fees", and separate this one long test out into a few individual tests (e.g., event, token with zero balance, if you send the router "excess" tokens it still only withdraws the calculated amount, etc.).

EndymionJkb (approved):
Shaping up well! Mostly doc comments, and some suggestions on the withdrawal tests.

EndymionJkb:
Per discussion this morning, we should also note in the description that this assumes "aggregation" - and we shouldn't merge this part until the Fee Sweeper is token-based vs. pool-based; otherwise it won't be able to process these "extra" tokens sent to it.

jubeira (approved):
LGTM, pretty clean! Good job.

jubeira:
> and we shouldn't merge this part until the Fee Sweeper is token-based vs. pool-based; otherwise it won't be able to process these "extra" tokens sent to it.

This is correct but we're still a bit far from actually deploying this. For the sake of making progress without growing the PR queue too much I think it's fine to proceed.

PR #1253, opened 2025-01-23
Add support for multiple Protocol Fee Burners

EndymionJkb:
# Description

Built off #1251, this adds a fee burner argument to the sweep call, and replaces the single fee burner storage with an allowlist of valid burners.

## Type of change

- [ ] Bug fix 
- [X] New feature 
- [ ] Breaking change 
- [ ] Dependency changes
- [ ] Code refactor / cleanup
- [ ] Optimization: [ ] gas / [ ] bytecode
- [ ] Documentation or wording changes
- [ ] Other

## Checklist:

- [X] The diff is legible and has no extraneous changes
- [X] Complex code has been commented, including external interfaces
- [X] Tests have 100% code coverage
- [X] The base branch is either `main`, or there's a description of how to merge

PR #1254, opened 2025-01-23
Aggregated fee sweeper

EndymionJkb:
# Description

Built on #1253 (support for multiple burners), this version "aggregates" collection from multiple pools. Instead of taking a pool argument and doing the withdrawal in the sweeper (requiring permission), it assumes the off-chain process has both collected and withdrawn the protocol fees to the sweeper (i.e, called withdrawal functions with the fee sweeper as the recipient), so that the token balances are in the sweeper.

It can then check these balances against a threshold, and sweep all (or optionally only some) of the token balance through a burner to a final recipient. This is more efficient if there are many pools with the same fee tokens (especially low liquidity pools), since the burns can be done less frequently, and with higher balances.

## Type of change

- [ ] Bug fix 
- [ ] New feature 
- [ ] Breaking change 
- [ ] Dependency changes
- [X] Code refactor / cleanup
- [ ] Other

## Checklist:

### DIFF
--- a/pkg/interfaces/contracts/standalone-utils/IProtocolFeeSweeper.sol
+++ b/pkg/interfaces/contracts/standalone-utils/IProtocolFeeSweeper.sol
@@ -23,12 +23,11 @@ interface IProtocolFeeSweeper {
     /**
      * @notice Emitted when a fee token is transferred directly, vs. calling the burner.
      * @dev This can happen if no target token or burner contract was specified, or the fee token is the target token.
-     * @param pool The pool on which the fee was collected
      * @param feeToken The token the fee was collected in (also the target token in this case; no swap necessary)
      * @param feeTokenAmount The number of feeTokens
      * @param recipient The recipient of the fee tokens
      */
-    event ProtocolFeeSwept(address indexed pool, IERC20 indexed feeToken, uint256 feeTokenAmount, address recipient);
+    event ProtocolFeeSwept(IERC20 indexed feeToken, uint256 feeTokenAmount, address recipient);
 
     /**
      * @notice Emitted when a burner is added to the protocol fee burner allowlist.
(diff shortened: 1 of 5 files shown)


pkg/interfaces/contracts/standalone-utils/IProtocolFeeSweeper.sol
```diff
@@ -72,26 +71,34 @@ interface IProtocolFeeSweeper {
     error ProtocolFeeBurnerNotAdded(address protocolFeeBurner);
 
     /**
-     * @notice Withdraw, convert, and forward protocol fees for a given pool and token.
-     * @dev This will withdraw the fee token from the controller to this contract, and attempt to convert and forward
-     * the proceeds to the fee recipient. Note that this requires governance to grant this contract permission to call
-     * `withdrawProtocolFeesForToken` on the `ProtocolFeeController`.
+     * @notice Cannot request burning more than the balance of the contract.
+     * @param feeToken The feeToken being burned
```
joaobrunoah, line 75:
```suggestion
     * @param feeToken The token being burned
```

joaobrunoah (approved):
LGTM!

pkg/interfaces/contracts/standalone-utils/IProtocolFeeSweeper.sol
EndymionJkb: This is fixed in the underlying PR.

pkg/standalone-utils/contracts/ProtocolFeeSweeper.sol
```diff
@@ -126,11 +121,11 @@ contract ProtocolFeeSweeper is IProtocolFeeSweeper, SingletonAuthentication, Ree
         }
     }
 
-    function _transferFeeToken(address pool, IERC20 feeToken, uint256 withdrawnTokenBalance) private {
+    function _transferFeeToken(IERC20 feeToken, uint256 workingBalance) private {
```
dveth2, line 124: `workingBalance` => `balance` ?

PR #1259, opened 2025-01-24
Filter surge cases for unbalanced liquidity

jubeira:
# Description

PoC: allow unbalanced liquidity for stable surge, blocking operations that would make the pool surge past the threshold (proportional excluded).

Using `after` hooks because it's easier to reconstruct original balances without caring about the operation kind. Also prevents reloading balances unnecessarily at the vault.

## Type of change

- [ ] Bug fix 
- [x] New feature 
- [ ] Breaking change 
- [ ] Dependency changes
- [ ] Code refactor / cleanup
- [ ] Other

## Checklist:

### DIFF
--- a/pkg/pool-hooks/contracts/StableSurgeHook.sol
+++ b/pkg/pool-hooks/contracts/StableSurgeHook.sol
@@ -54,7 +54,7 @@ contract StableSurgeHook is BaseHooks, VaultGuard, SingletonAuthentication {
     uint256 private immutable _defaultSurgeThresholdPercentage;
 
     // Store the current threshold and max fee for each pool.
-    mapping(address pool => SurgeFeeData data) private _surgeFeePoolData;
+    mapping(address pool => SurgeFeeData data) internal _surgeFeePoolData;
 
     /**
      * @notice A new `StableSurgeHook` contract has been registered successfully.
--- a/pkg/pool-hooks/contracts/StableSurgeHook.sol
+++ b/pkg/pool-hooks/contracts/StableSurgeHook.sol
@@ -8,10 +8,12 @@ import { IBasePoolFactory } from "@balancer-labs/v3-interfaces/contracts/vault/I
 import { IHooks } from "@balancer-labs/v3-interfaces/contracts/vault/IHooks.sol";
 import { IVault } from "@balancer-labs/v3-interfaces/contracts/vault/IVault.sol";
 import {
+    AddLiquidityKind,
     LiquidityManagement,
     TokenConfig,
     PoolSwapParams,
     HookFlags,
+    RemoveLiquidityKind,
     SwapKind,
     Rounding
 } from "@balancer-labs/v3-interfaces/contracts/vault/VaultTypes.sol";
@@ -112,6 +114,8 @@ contract StableSurgeHook is BaseHooks, VaultGuard, SingletonAuthentication {
     /// @inheritdoc IHooks
     function getHookFlags() public pure override returns (HookFlags memory hookFlags) {
         hookFlags.shouldCallComputeDynamicSwapFee = true;
+        hookFlags.shouldCallAfterAddLiquidity = true;
+        hookFlags.shouldCallAfterRemoveLiquidity = true;
     }
 
     /**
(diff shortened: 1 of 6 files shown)


pkg/pool-hooks/contracts/StableSurgeHook.sol
```diff
@@ -187,7 +191,100 @@ contract StableSurgeHook is BaseHooks, VaultGuard, SingletonAuthentication {
...
+        address,
+        address pool,
+        AddLiquidityKind kind,
+        uint256[] memory amountsInScaled18,
+        uint256[] memory amountsInRaw,
+        uint256,
+        uint256[] memory balancesScaled18,
+        bytes memory
+    ) public view override returns (bool success, uint256[] memory hookAdjustedAmountsInRaw) {
+        hookAdjustedAmountsInRaw = amountsInRaw; // don't adjust amounts in.
```
EndymionJkb, line 227: Might get optimized by the compiler, but this might be copying twice (here and in the return). Other hooks don't have this assignment (or the named return value at all), and just return `amountsInRaw`. Every path does a return here, so not declaring `hookAdjustedAmountsInRaw` and just returning `amountsInRaw` everywhere seems cleaner.

pkg/pool-hooks/contracts/StableSurgeHook.sol
```diff
@@ -187,7 +191,100 @@ contract StableSurgeHook is BaseHooks, VaultGuard, SingletonAuthentication {
...
+        }
+
+        // Rebuild old balances before adding liquidity.
+        uint256[] memory oldBalancesScaled18 = new uint256[](balancesScaled18.length);
+        for (uint256 i = 0; i < balancesScaled18.length; ++i) {
+            oldBalancesScaled18[i] = balancesScaled18[i] - amountsInScaled18[i];
+        }
+
+        SurgeFeeData memory surgeFeeData = _surgeFeePoolData[pool];
+        uint256 newTotalImbalance = StableSurgeMedianMath.calculateImbalance(balancesScaled18);
```
EndymionJkb, line 241: This is repeated below, and could be rolled into `_isSurging` if you passed in the pool and balancesScaled18.

pkg/pool-hooks/contracts/StableSurgeHook.sol
```diff
@@ -187,7 +191,100 @@ contract StableSurgeHook is BaseHooks, VaultGuard, SingletonAuthentication {
...
+        address,
+        address pool,
+        RemoveLiquidityKind kind,
+        uint256,
+        uint256[] memory amountsOutScaled18,
+        uint256[] memory amountsOutRaw,
+        uint256[] memory balancesScaled18,
+        bytes memory
+    ) public view override returns (bool success, uint256[] memory hookAdjustedAmountsOutRaw) {
+        hookAdjustedAmountsOutRaw = amountsOutRaw; // don't adjust amounts out.
```
EndymionJkb, line 264: Same as above; could remove `hookAdjustedAmountsOutRaw` and just return `amountsOutRaw`

pkg/pool-hooks/contracts/StableSurgeHook.sol
```diff
@@ -234,37 +331,22 @@ contract StableSurgeHook is BaseHooks, VaultGuard, SingletonAuthentication {
...
-        uint256 oldTotalImbalance = StableSurgeMedianMath.calculateImbalance(params.balancesScaled18);
+        bool isSurging = _isSurging(
+            surgeFeeData,
+            params.balancesScaled18,
+            newTotalImbalance
+        );
 
-        SurgeFeeData storage surgeFeeData = _surgeFeePoolData[pool];
-        if (newTotalImbalance <= oldTotalImbalance || newTotalImbalance <= surgeFeeData.thresholdPercentage) {
+        if (isSurging == false) {
```
EndymionJkb, line 349: Consider positive logic:
```
if (isSurging) {
    // surgeFee = staticFee + (maxFee - staticFee) * (pctImbalance - pctThreshold) / (1 - pctThreshold).
    //
    // As you can see from the formula, if it’s unbalanced exactly at the threshold, the last term is 0,
    // and the fee is just: static + 0 = static fee.
    // As the unbalanced proportion term approaches 1, the fee surge approaches: static + max - static ~= max fee.
    // This formula linearly increases the fee from 0 at the threshold up to the maximum fee.
    // At 35%, the fee would be 1% + (0.95 - 0.01) * ((0.35 - 0.3)/(0.95-0.3)) = 1% + 0.94 * 0.0769 ~ 8.2%.
    // At 50% unbalanced, the fee would be 44%. At 99% unbalanced, the fee would be ~94%, very close to the maximum.
    return
        staticFeePercentage +
        (surgeFeeData.maxSurgeFeePercentage - staticFeePercentage).mulDown(
            (newTotalImbalance - surgeFeeData.thresholdPercentage).divDown(
                uint256(surgeFeeData.thresholdPercentage).complement()
            )
        );
}

return staticFeePercentage;
```

pkg/pool-hooks/contracts/StableSurgeHook.sol
```diff
@@ -285,6 +367,25 @@ contract StableSurgeHook is BaseHooks, VaultGuard, SingletonAuthentication {
...
+            return false;
+        }
+
+        uint256 oldTotalImbalance = StableSurgeMedianMath.calculateImbalance(currentBalances);
+
+        if (newTotalImbalance <= oldTotalImbalance || newTotalImbalance <= surgeFeeData.thresholdPercentage) {
+            return false;
+        }
+
+        return true;
```
EndymionJkb, line 386:
```suggestion
        return newTotalImbalance > oldTotalImbalance && newTotalImbalance > surgeFeeData.thresholdPercentage;
```

jubeira: yeah I left it as it was to minimize changes, but this way is probably better.

pkg/pool-hooks/contracts/StableSurgeHook.sol
jubeira: The problem is that I need `newTotalImbalance` in the dynamic swap fee computation, and I don't want a function that returns 2 values.

This way I incur in some repetition, but otherwise it's more readable. I'll do one more pass.

jubeira: I'm on the fence about this one. It's just 2 extra lines, and we're already reusing `isSurging` everywhere.

I think I prefer this inline as opposed to one more level down the call trace. I like pasta al dente.

pkg/pool-hooks/contracts/StableSurgeHook.sol
```diff
@@ -190,6 +194,91 @@ contract StableSurgeHook is BaseHooks, VaultGuard, SingletonAuthentication {
...
+    function getSurgeFeePercentage(
+        PoolSwapParams calldata params,
+        address pool,
+        uint256 staticSwapFeePercentage
+    ) public view returns (uint256) {
+        uint256 numTokens = params.balancesScaled18.length;
+
+        uint256 amountCalculatedScaled18 = StablePool(pool).onSwap(params);
+
+        uint256[] memory newBalances = new uint256[](numTokens);
```
EndymionJkb, line 206:
```suggestion
        uint256 amountCalculatedScaled18 = StablePool(pool).onSwap(params);

        uint256[] memory newBalances = new uint256[](params.balancesScaled18.length);
```
Could get rid of this, as it's only used once. Also defined far from use.

EndymionJkb (approved):
Looks good!

joaobrunoah (approved):
LGTM!

PR #1260, opened 2025-01-27
Lbp one weight change

mkflow27:
See below for the original text.

This version is in the same spirit, keeping things simple for aggregators by limiting the scope to token sales, and essentially fixing all parameters on deployment: the sale has a fixed start and end time, with a single weight change schedule. Liquidity can only be added before the start, and only removed after the end. Swaps can only occur during the sale.

There is a flag indicating whether the project token can be sold "back" into the pool. If not, it is a "one way" distribution. Everything is immutable. The only things that can change unpredictably are the actual balances (due to swaps), and the swap fee (which can be changed by the owner).

# Goal of the draft:
- I would like to get feedback on the approach to the feature implementations. Once they are agreed upon I will move on to cleaning up the tests.

# Description

This pr adds changes to the LBP. Namely:
- Move weight change setting to the constructor
- Introduce `bootstrapToken` which allows for the pool creator to determine if the bootstrap token can be sold&bought or only bought.
- introduce `allowRemovalOnlyAfterWeightChange` which allows for the pool creator to determine if removing liquidity is allowed during pool lifecycle or only after weight change has ended
- Allow swaps to only be possible during & after weight change

Overall the goal of this pr is to deliver the following feature requests:
- Support only 1 weight change, defined on construction, this would make it much easier for aggregators to know the weight, without the need to pool for events. The start time can be at a point in the future
- Remove swaps enabled, swaps are turned on during the weight change, otherwise they are off. 
- A flag, if set, the liquidity can only be removed once the weight change has ended
- A flag if set, the token can only be bought and not sold

## Type of change

- [ ] Bug fix 
- [x] New feature 
- [ ] Breaking change 
- [ ] Dependency changes
- [ ] Code refactor / cleanup
- [ ] Other

## Checklist:

## Issue Resolution


pkg/pool-weighted/contracts/lbp/LBPool.sol
```diff
@@ -283,13 +303,46 @@ contract LBPool is WeightedPool, Ownable2Step, BaseHooks {
...
-    function _setSwapEnabled(bool swapEnabled) private {
-        _poolState.swapEnabled = swapEnabled;
+    function _getTokenSwapAllowed(PoolSwapParams memory params) private view returns (bool) {
+        if (restrictSaleOfBootstrapToken) {
+            if (params.kind == SwapKind.EXACT_IN) {
+                return params.indexIn == _bootstrapTokenIndex ? false : true;
+            } else {
+                // exact out swap
+                return params.indexOut == _bootstrapTokenIndex ? false : true;
+            }
```
jubeira, line 318: Hmm I think this is not correct.

You should just check that the index out is not the bootstrap token index. It really doesn't matter which end is fixed and which end is calculated; you only care about not pulling the bootstrap token from the vault.

```suggestion
            // Do not allow pulling bootstrap token out.
            return params.indexOut != _bootstrapTokenIndex;
```

pkg/pool-weighted/contracts/lbp/LBPool.sol
```diff
@@ -302,6 +355,16 @@ contract LBPool is WeightedPool, Ownable2Step, BaseHooks {
...
     ) internal virtual {
+        InputHelpers.ensureInputLengthMatch(_NUM_TOKENS, startWeights.length);
+        InputHelpers.ensureInputLengthMatch(_NUM_TOKENS, endWeights.length);
+
+        if (endWeights[0] < _MIN_WEIGHT || endWeights[1] < _MIN_WEIGHT) {
+            revert MinWeight();
+        }
+        if (endWeights[0] + endWeights[1] != FixedPoint.ONE) {
+            revert NormalizedWeightInvariant();
+        }
```
jubeira, line 366: I think this is all checked inside the `WeightedPool` constructor. This was only needed for the external method, given that parameters needed validation. But if we're only calling this inside the constructor it shouldn't be needed.

pkg/pool-weighted/contracts/lbp/LBPool.sol
```diff
@@ -254,11 +248,37 @@ contract LBPool is WeightedPool, Ownable2Step, BaseHooks {
...
+        RemoveLiquidityKind,
+        uint256,
+        uint256[] memory,
+        uint256[] memory,
+        bytes memory
+    ) public view override onlyVault returns (bool success) {
+        if (router != _trustedRouter) {
+            revert RouterNotTrusted();
+        }
+
```
jubeira, line 268:
```suggestion
```

I don't think this is needed; you're not doing anything about the sender here.

jubeira (commented):
Hey @mkflow27, thanks for putting this together!

Looks good overall; here's a first pass with some comments, mostly small things where there's room for improvement and some questions.

Besides that, the main question I have is whether we want all pools to be like this (i.e. a single weight update at construction time vs configurable weight updates). If we want this to be more flexible and have _some_ pools with a single update and _some_ pools with configurable updates, then we might need to add a flag and rethink a few things.

Same for disabling swaps. Do we just want them enabled after the update starts and that's it?

pkg/pool-weighted/contracts/lbp/LBPool.sol
EndymionJkb: Agree - we don't know what their process is. It's possible the owner is an EOA or contract that mints the project tokens or something, that will send the BPT somewhere (e.g., to a multi-sig) for withdrawal, as that's when there is a lot of value and it matters.

pkg/pool-weighted/contracts/lbp/LBPool.sol
```diff
@@ -27,6 +28,12 @@ import { WeightedPool } from "../WeightedPool.sol";
  */
 contract LBPool is WeightedPool, Ownable2Step, BaseHooks {
     using SafeCast for *;
+    address public bootstrapToken;
```
EndymionJkb, line 31: We typically make these private and add getters. Apparently there is no way to change these, so they could all be immutable. Instead of individual getters, you could define and implement ImmutableData and DynamicData interfaces (like Weighted and Stable pools do)...

And here we have a little problem, since the WeightedPoolImmutable data is wrong - and it's not virtual. So I think we need to make them virtual and revert (I guess DynamicData is still fine).

pkg/pool-weighted/contracts/lbp/LBPool.sol
```diff
@@ -85,12 +85,22 @@ contract LBPool is WeightedPool, Ownable2Step, BaseHooks {
     /// @dev Indicates that the `owner` has disabled swaps.
     error SwapsDisabled();
 
+    /// @dev Indicates that removing liquidity is not allowed.
+    error RemovingLiquidityNotAllowed();
```
EndymionJkb, line 89: Also as a style not, type definitions go on top, then state variables, then events, then errors, then functions.

EndymionJkb (commented):
I'm proposing some structural changes. Let me know what you think.

pkg/pool-weighted/contracts/lbp/LBPool.sol
```diff
@@ -283,13 +303,46 @@ contract LBPool is WeightedPool, Ownable2Step, BaseHooks {
     }
 
     function _getPoolSwapEnabled() private view returns (bool) {
-        return _poolState.swapEnabled;
+        // Swaps are only enabled, once the weight change has started.
+        (uint256 startTime, , ) = getGradualWeightUpdateParams();
+        return block.timestamp >= startTime;
```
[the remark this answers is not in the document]
EndymionJkb: It seems like if this is "automatic" like this, it should only allow trading during the sale - so the check should be that the time is between the start and end. If the point of this was to prevent LBP owners from prematurely terminating the sale - so we're saying it's truly a fixed length sale - then you shouldn't be able to trade after it either.

pkg/pool-weighted/contracts/lbp/LBPool.sol
```diff
@@ -254,11 +248,37 @@ contract LBPool is WeightedPool, Ownable2Step, BaseHooks {
         return IRouterCommon(router).getSender() == owner();
     }
 
+    /**
```
EndymionJkb, line 251: Also, how serious are we about making the sale *completely* deterministic? It is still possible for the owner to change the price or sell more tokens by adding liquidity during the sale (a feature of the original LBP, though rarely used).

If we want it to be completely predictable, we could block adding liquidity after the sale starts. (You might then have a dead contract if you failed to initialize it before the start of the sale. Ok I guess as long as funds can't be locked. Worst case, they have to wait for the end of the "sale" to withdraw again.)

pkg/pool-weighted/contracts/lbp/LBPool.sol
jubeira: Let's leave it anyways. Doesn't hurt really, it's only called once.

joaobrunoah:
Should we use the Balancer Contract Registry to identify a trusted router in the LBP?

pkg/pool-weighted/test/LBPool.test.ts
jubeira, line 1: Why are these tests deleted?

mkflow27: I initially did not have enough time to refactor the typescript tests. So I removed them with the intention of adding them during my work on this pr.

joaobrunoah (approved):
LGTM! I've removed a few dependencies/errors used in the old version, but deprecated in the new one.

jubeira (approved):
Just a few comments, otherwise LGTM! Good job everyone.

pkg/pool-weighted/contracts/lbp/LBPool.sol
```diff
@@ -18,55 +29,45 @@ import { BaseHooks } from "@balancer-labs/v3-vault/contracts/BaseHooks.sol";
...
+
+    uint256 private immutable _startTime;
+    uint256 private immutable _endTime;
+
+    uint256 private immutable _projectTokenStartWeight;
+    uint256 private immutable _reserveTokenStartWeight;
+    uint256 private immutable _projectTokenEndWeight;
+    uint256 private immutable _reserveTokenEndWeight;
+
+    // If true, project tokens can only be the token out in the context of a swap.
```
EndymionJkb, line 66:
```suggestion
    // If true, project tokens can only be bought, not sold back to the pool (i.e., they cannot be the `tokenIn`
    // of a swap)
```

EndymionJkb (approved):
LGTM! Just some comment clarifications

jubeira:
Created ticket for the failure: #1275

PR #1282, opened 2025-02-11
Make CI go faster

jubeira:
# Description

Use every available CPU core when running all base scripts across every package.

## Type of change

- [ ] Bug fix 
- [ ] New feature 
- [ ] Breaking change 
- [ ] Dependency changes
- [ ] Code refactor / cleanup
- [x] Other

## Checklist:

## Issue Resolution

N/A

### DIFF
--- a/package.json
+++ b/package.json
@@ -21,13 +21,13 @@
     "Nicolás Venturo <nico@balancerlabs.dev>"
   ],
   "scripts": {
-    "build": "yarn workspaces foreach --parallel --verbose run build",
-    "lint": "NODE_NO_WARNINGS=1 yarn workspaces foreach --parallel --verbose run lint",
-    "test": "yarn workspaces foreach --parallel --verbose run test",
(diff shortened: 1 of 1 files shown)


EndymionJkb (approved):
Was it only using 10% of the brain? :) Looks like it defaults to 8 cores, and there is some risk of contention/corruption if we turn this on. Worth a try, though.

jubeira:
well maybe CI doesn't change that much, and also this just affects lint and build as the others are separated anyways.

But locally I have 16 threads so I do see a difference :)

joaobrunoah (approved):
Locally it'll make difference, which is super cool!

I believe in the CI it didn't do a difference because each job runs inside its own docker, so it's not possible to parallelize the tests, but maybe there's a way to improve the docker job in the github CI.

jubeira:
I'd leave it like this for now. I'm happy with the local linter upgrade :)

PR #1283, opened 2025-02-12
Add project/reserve token information to LBPools

### LINKED ISSUE #1278: LBPs should have a way to differentiate project/reserve
Input from internal users: kind of in the "duh" category :) Surprised we didn't think of this.

Add to the immutable data, and perhaps explicit getters.

EndymionJkb:
# Description

We explicitly store the project and reserve tokens, but don't have getters for them. It's fairly obvious to a human which is which, but automated systems might also need to know this; for instance, aggregators who must know which direction to block if `blockProjectTokenSwapsIn` is true.

Both on- and off-chain processes might conceivably need this, so this PR:
1) Adds an LBP-specific event with the project and reserve tokens named
2) Adds explicit getters for the project and reserve tokens, in case somebody just wants that, and doesn't need the full immutable data (including token decimals, etc.)
3) Adds project/reserve information to the immutable data getter, so that processes already calling that can either use the existing tokens array (e.g., if they don't care which is which and just need the tokens), or use the added indexes to pull the token(s) they need out of the array, without needing to make any more calls

## Type of change

- [ ] Bug fix 
- [X] New feature 
- [ ] Breaking change 
- [ ] Dependency changes
- [ ] Code refactor / cleanup
- [ ] Other

## Checklist:

## Issue Resolution

Resolves #1278

### DIFF
--- a/pkg/pool-weighted/contracts/lbp/LBPool.sol
+++ b/pkg/pool-weighted/contracts/lbp/LBPool.sol
@@ -140,6 +140,8 @@ contract LBPool is ILBPool, WeightedPool, Ownable2Step, BaseHooks {
 
         (_projectTokenIndex, _reserveTokenIndex) = lbpParams.projectToken < lbpParams.reserveToken ? (0, 1) : (1, 0);
 
+        emit LBPoolCreated(lbpParams.projectToken, lbpParams.reserveToken);
+
         // Preserve event compatibility with previous LBP versions.
         uint256[] memory startWeights = new uint256[](_TWO_TOKENS);
         uint256[] memory endWeights = new uint256[](_TWO_TOKENS);
--- a/pkg/pool-weighted/contracts/lbp/LBPool.sol
+++ b/pkg/pool-weighted/contracts/lbp/LBPool.sol
@@ -140,8 +140,6 @@ contract LBPool is ILBPool, WeightedPool, Ownable2Step, BaseHooks {
 
         (_projectTokenIndex, _reserveTokenIndex) = lbpParams.projectToken < lbpParams.reserveToken ? (0, 1) : (1, 0);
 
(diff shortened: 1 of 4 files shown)


pkg/pool-weighted/contracts/lbp/LBPool.sol
```diff
@@ -254,18 +264,21 @@ contract LBPool is ILBPool, WeightedPool, Ownable2Step, BaseHooks {
...
-        data.startWeights[_projectTokenIndex] = _projectTokenStartWeight;
-        data.startWeights[_reserveTokenIndex] = _reserveTokenStartWeight;
+        data.startWeights[data.projectTokenIndex] = _projectTokenStartWeight;
+        data.startWeights[data.reserveTokenIndex] = _reserveTokenStartWeight;
 
         data.endWeights = new uint256[](_TWO_TOKENS);
-        data.endWeights[_projectTokenIndex] = _projectTokenEndWeight;
-        data.endWeights[_reserveTokenIndex] = _reserveTokenEndWeight;
+        data.endWeights[data.projectTokenIndex] = _projectTokenEndWeight;
+        data.endWeights[data.reserveTokenIndex] = _reserveTokenEndWeight;
```
jubeira, line 281: This is not storage, so I don't think it makes much of a difference to use `_projectTokenIndex` and `_reserveTokenIndex`. Either way should be fine.

jubeira (approved):
LGTM! Good catch @EndymionJkb

pkg/pool-weighted/contracts/lbp/LBPool.sol
EndymionJkb: Yeah; on second thought, I think I will change it back to use the immutables directly. There should be no performance difference (just switching one compile-time constant for another), but using the immutable indexes directly is clearer.

elshan-eth (approved):
LGTM!

PR #1294, opened 2025-02-14
Allow swapFeeManager to also update the A factor

### LINKED ISSUE #1288: Stable pools: allow swap fee manager to modify amp factor

EndymionJkb:
# Description

It is a common pattern for the swap fee manager to also need to adjust the amplification parameter. This PR grants this permission non-exclusively (i.e., governance can still do it).

## Type of change

- [ ] Bug fix 
- [X] New feature 
- [ ] Breaking change 
- [ ] Dependency changes
- [ ] Code refactor / cleanup
- [ ] Other

## Checklist:

## Issue Resolution

Resolves #1288.

### DIFF
--- a/pkg/pool-stable/contracts/StablePool.sol
+++ b/pkg/pool-stable/contracts/StablePool.sol
@@ -126,7 +126,6 @@ contract StablePool is IStablePool, BalancerPoolToken, BasePoolAuthentication, P
             // Otherwise, defer to governance.
             _authenticateCaller();
         }
-
         _;
     }
 
--- a/pkg/pool-stable/contracts/StablePool.sol
+++ b/pkg/pool-stable/contracts/StablePool.sol
@@ -45,6 +45,22 @@ contract StablePool is IStablePool, BalancerPoolToken, BasePoolAuthentication, P
     using FixedPoint for uint256;
     using SafeCast for *;
 
+    /**
+     * @notice Parameters used to deploy a new Stable Pool.
+     * @param name ERC20 token name
+     * @param symbol ERC20 token symbol
+     * @param amplificationParameter Controls the "flatness" of the invariant curve. higher values = lower slippage,
+     * and assumes prices are near parity. lower values = closer to the constant product curve (e.g., more like a
+     * weighted pool). This has higher slippage, and accommodates greater price volatility
+     * @param version The stable pool version
+     */
+    struct NewPoolParams {
+        string name;
+        string symbol;
+        uint256 amplificationParameter;
+        string version;
+    }
+
     // This contract uses timestamps to slowly update its Amplification parameter over time. These changes must occur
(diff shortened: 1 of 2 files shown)


pkg/pool-stable/contracts/StablePool.sol
```diff
@@ -101,19 +117,16 @@ contract StablePool is IStablePool, BalancerPoolToken, BasePoolAuthentication, P
...
-        string symbol;
-        uint256 amplificationParameter;
-        string version;
+    modifier authenticateByRole() {
+        // Allow if this is the swapFeeManager.
+        if (msg.sender != _vault.getPoolRoleAccounts(address(this)).swapFeeManager) {
+            // Otherwise, defer to governance.
+            _authenticateCaller();
+        }
+        _;
```
jubeira, line 129: This pattern is slightly different from the one we use for swap fees though.

In this case governance can always allow the permissioned function to be protected even if there's a designated swap fee manager. For the swap fee percentage in the vault admin, if the pool has a designated swap fee manager then governance is out of the loop.

I'm not against this pattern, but I'd double check whether it's what we want or not.

jubeira (commented):
Thanks @EndymionJkb!

Looks good, but there's a subtlety I'd like to iron out before proceeding. Let's double check and circle back on this.

pkg/pool-stable/contracts/StablePool.sol
EndymionJkb: Yes; I noted this in the docs, thinking it was being done by governance now, and they wanted the swap manager to also be able to do it as a convenience. We need to check with whoever wanted this change to make sure we have the correct requirements.

jubeira (approved):
Double-checked with maxis, this is fine.

pkg/pool-stable/test/foundry/StablePool.t.sol
```diff
@@ -118,6 +120,46 @@ contract StablePoolTest is BasePoolTest, StablePoolContractsDeployer {
...
+        IStablePool(pool).startAmplificationParameterUpdate(newAmplificationParameter, endTime);
+
+        (, bool isUpdating, ) = IStablePool(pool).getAmplificationParameter();
+        assertTrue(isUpdating, "Amplification update not started");
+
+        IStablePool(pool).stopAmplificationParameterUpdate();
+        vm.stopPrank();
+
+        (, isUpdating, ) = IStablePool(pool).getAmplificationParameter();
+        assertFalse(isUpdating, "Amplification update not stopped");
```
jubeira, line 160: Maybe now give permissions via governance to bob and trigger an update again?

jubeira (commented):
Small suggestion for tests

elshan-eth (approved):
LGTM!

PR #1295, opened 2025-02-14
Create lib to handle Weth in routers

joaobrunoah:
# Description

Some routers (like CowRouter) will need to handle Native ETH, but don't need extra features from RouterCommon (like Permit2 or some transient states).

This PR decouples the logic to handle Weth from RouterCommon, so it can be reused by other routers.

## Type of change

- [ ] Bug fix 
- [ ] New feature 
- [ ] Breaking change 
- [ ] Dependency changes
- [x] Code refactor / cleanup
- [ ] Other

## Checklist:

### DIFF
--- a/pkg/vault/contracts/RouterCommon.sol
+++ b/pkg/vault/contracts/RouterCommon.sol
@@ -26,6 +26,7 @@ import {
 } from "@balancer-labs/v3-solidity-utils/contracts/helpers/TransientStorageHelpers.sol";
 
 import { VaultGuard } from "./VaultGuard.sol";
+import { RouterWethLib } from "./lib/RouterWethLib.sol";
 
 /**
  * @notice Abstract base contract for functions shared among all Routers.
@@ -36,7 +37,7 @@ import { VaultGuard } from "./VaultGuard.sol";
 abstract contract RouterCommon is IRouterCommon, VaultGuard, ReentrancyGuardTransient, Version {
     using Address for address payable;
     using StorageSlotExtension for *;
-    using SafeERC20 for IWETH;
+    using RouterWethLib for IWETH;
     using SafeCast for *;
 
     // NOTE: If you use a constant, then it is simply replaced everywhere when this constant is used by what is written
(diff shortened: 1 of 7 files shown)


EndymionJkb (approved):
Great simplification!

pkg/vault/contracts/RouterCommon.sol
```diff
@@ -264,31 +263,12 @@ abstract contract RouterCommon is IRouterCommon, VaultGuard, ReentrancyGuardTran
...
-     * @dev Returns excess ETH back to the contract caller. Checks for sufficient ETH balance are made right before
-     * each deposit, ensuring it will revert with a friendly custom error. If there is any balance remaining when
-     * `_returnEth` is called, return it to the sender.
-     *
-     * Because the caller might not know exactly how much ETH a Vault action will require, they may send extra.
-     * Note that this excess value is returned *to the contract caller* (msg.sender). If caller and e.g. swap sender
-     * are not the same (because the caller is a relayer for the sender), then it is up to the caller to manage this
-     * returned ETH.
-     */
     function _returnEth(address sender) internal {
```
jubeira, line 266: I think I'd leave this one out, or maybe put the whole function in another library? (that'd be out of scope for this PR).

This function doesn't really use WETH, and as the deleted comment states it's cheaper to do it this way than reading transient storage unconditionally. That's why I'm not super convinced.

It is true that we will likely need this elsewhere though. But maybe instead of WETH we can do it with the `isReturnEthLockedSlot` (weird, but it might work and we wouldn't need to actually read the slot)

pkg/vault/test/foundry/RouterCommon.t.sol
jubeira, line 1: I guess this should be mostly enough, but can we also have very simple unit tests for this new lib?

jubeira (commented):
Thanks @joaobrunoah, looks good for the most part! I have a few observations; other than that this makes sense.

pkg/vault/contracts/RouterCommon.sol
joaobrunoah: Makes sense, I'll remove the refactor of the function `_returnEth` from this PR.

elshan-eth (approved):
Maybe small unit test would be good for this lib but in general LGTM!

jubeira (approved):
LGTM!

PR #1298, opened 2025-02-17
Change MAX_AMP to 50000

elshan-eth:
# Description

This PR changes MAX_AMP for stable math and removes unnecessary note

## Type of change

- [ ] Bug fix 
- [ ] New feature 
- [ ] Breaking change 
- [ ] Dependency changes
- [ ] Code refactor / cleanup
- [ ] Other

## Checklist:

### DIFF
--- a/pkg/solidity-utils/contracts/math/StableMath.sol
+++ b/pkg/solidity-utils/contracts/math/StableMath.sol
@@ -31,22 +31,13 @@ library StableMath {
     uint256 public constant MAX_STABLE_TOKENS = 5;
 
     uint256 internal constant MIN_AMP = 1;
-    uint256 internal constant MAX_AMP = 5000;
+    uint256 internal constant MAX_AMP = 50000;
     uint256 internal constant AMP_PRECISION = 1e3;
 
     // Invariant growth limit: non-proportional add cannot cause the invariant to increase by more than this ratio.
     uint256 internal constant MIN_INVARIANT_RATIO = 60e16; // 60%
     // Invariant shrink limit: non-proportional remove cannot cause the invariant to decrease by less than this ratio.
     uint256 internal constant MAX_INVARIANT_RATIO = 500e16; // 500%
-
-    // Note on unchecked arithmetic:
-    // This contract performs a large number of additions, subtractions, multiplications and divisions, often inside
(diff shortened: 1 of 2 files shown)


pkg/solidity-utils/contracts/math/StableMath.sol
```diff
@@ -31,23 +31,14 @@ library StableMath {
...
     uint256 internal constant MIN_INVARIANT_RATIO = 60e16; // 60%
     // Invariant shrink limit: non-proportional remove cannot cause the invariant to decrease by less than this ratio.
     uint256 internal constant MAX_INVARIANT_RATIO = 500e16; // 500%
 
-    // Note on unchecked arithmetic:
-    // This contract performs a large number of additions, subtractions, multiplications and divisions, often inside
-    // loops. Since many of these operations are gas-sensitive (as they happen e.g. during a swap), it is important to
-    // not make any unnecessary checks. We rely on a set of invariants to avoid having to use checked arithmetic,
-    // including:
-    //  - the amplification parameter is bounded by MAX_AMP * AMP_PRECISION, which fits in 23 bits
```
EndymionJkb, line 47: It's 26 bits now, but I guess we never directly do this; just not sure why this was there in the first place (from v2 days).

EndymionJkb (commented):
There's still a failing test; need to increase the length of the change to trigger the right error.

EndymionJkb (approved):
I guess this is fine if all the fuzz tests are passing, etc.

joaobrunoah (commented):
I believe we don't have a test that puts the pool at the edge. To be comfortable with this change, I think we should have a test of a stable pool with 5 tokens, and see how much balance we can have on each token before the math reverts.

With the current tests, we test only 2 tokens, so the invariant is a low number. Anyway, checking the invariant calculation, it doesn't look like it will overflow, but I'd do this test just in case.

jubeira (approved):
So apparently there's no particular issue with this right?

Taking a second look,
- The comment was stale, so there's no issue with the amount of bits used to store amp
- In `computeInvariant`, the amplification factor is multiplied by the sum of the balances and by the previous invariant. None of the two values should be big enough to make this overflow by increasing max amp 10x.
- In `computeBalance` the amp factor is always in the denominator, so there won't be any overflow because of this change.

LGTM!

PR #1299, opened 2025-02-18
Protocol Fee Controller migration contract

### LINKED ISSUE #1290: Governance script for Protocol Fee Controller V2
- Start with admin role
- Grant role to itself
- Grant role to upgrade fee controller / set defaults to itself
- Set defaults
- Upgrade controller
- Revoke all permissions (upgrade fee controller, set defaults)
- Revoke admin role

EndymionJkb:
# Description

Based on #1307, this is the monorepo part of the ProtocolFeeController migration. It introduces a /governance-scripts package, and adds contracts used for the migration.

Turns out this is actually quite tricky! Essentially, there is no way to copy the "hidden" state that is set only on registration through the Vault or by pool creators. See #1307 for the ProtocolFeeController changes necessary to accomplish this (i.e., a `migratePool` function).

We'd thought the problem was only pool creators, but it's actually worse. We cannot call the `updateProtocolSwap/YieldFeePercentage` functions to "copy" pool state, because those call into the Vault to update the aggregate percentage - and this function can only be called by the current fee controller: not one we're migrating to.

So the final solution is to migrate to a new Protocol Fee Controller that has the ability to copy hidden state, with one twist. We needed to add new getters in order to copy the pool creator percentages, so these do not exist in the current (original) ProtocolFeeController. The first migration (very soon) has no pools with pool creators, so it's fine, but we need to tolerate failure so that the first migration works. (If any pool creators are missed, the pool creators can simply set the percentages again in the new fee controller, so it's not catastrophic. The fact that there is a pool creator cannot be lost, as that's stored in the Vault.)

So in the first migration, all the pool creator getter calls will fail (try... catch, so it doesn't revert), which is fine because they're all zero anyway. However, in future migrations they will work and correctly copy the pool creator fees.

This adds a new package, so if you see "hardhat not found" or something like that, run "yarn" at the top level to update the dependencies.

## Type of change

- [ ] Bug fix 
- [X] New feature 
- [ ] Breaking change 
- [ ] Dependency changes
- [ ] Code refactor / cleanup
- [ ] Other

## Checklist:

## Issue Resolution

Resolves #1290


pkg/governance-scripts/contracts/ProtocolFeeControllerMigration.sol
```diff
@@ -0,0 +1,153 @@
...
+     */
+    error InvalidFeeController();
+
+    /// @notice Migration can only be performed once.
+    error AlreadyMigrated();
+
+    constructor(IVault _vault, IProtocolFeeController _newFeeController) {
+        oldFeeController = _vault.getProtocolFeeController();
+
+        // Ensure valid fee controllers. Also ensure that we are not trying to operate on the current fee controller.
```
EndymionJkb, line 63: Could avoid all this by simply deploying a new ProtocolFeeController here (the only argument is the Vault), but we don't usually do that in migrations. Would be a simplification, though.

pkg/vault/contracts/ProtocolFeeController.sol
```diff
@@ -412,6 +450,59 @@ contract ProtocolFeeController is
         // Allow tracking pool fee percentages in all cases (e.g., when the pool is protocol-fee exempt).
         emit InitialPoolAggregateSwapFeePercentage(pool, aggregateSwapFeePercentage, protocolFeeExempt);
         emit InitialPoolAggregateYieldFeePercentage(pool, aggregateYieldFeePercentage, protocolFeeExempt);
+
+        if (poolCreator != address(0)) {
+            emit PoolWithCreatorRegistered(pool, poolCreator, protocolFeeExempt);
+        }
```
jubeira, line 456: !!!!
doesn't this brick pool registration when there's a pool creator? This is set by the vault: https://github.com/balancer/balancer-v3-monorepo/blob/main/pkg/vault/contracts/VaultExtension.sol#L260-L261

jubeira: Btw how does this not make any test fail?

pkg/governance-scripts/contracts/ProtocolFeeControllerMigrationV2.sol
```diff
@@ -0,0 +1,111 @@
...
+        // At the end of this process, governance must still withdraw any leftover protocol fees from the old
+        // controller (i.e., that have been collected but not withdrawn). Pool creators likewise would still need to
+        // withdraw any leftover pool creator fees from the old controller.
+        for (uint256 i = 0; i < pools.length; ++i) {
+            // This function is not in the public interface.
+            ProtocolFeeController(address(newFeeController)).migratePool(pools[i]);
+        }
+    }
+
+    function finalizeMigration() external virtual authenticate {
```
jubeira, line 89: Hm.

This is a bit tricky. In V2 we just granted admin access and then we just cycled through the stages in the script until the end.

In this case we also need permission to call `finalizeMigration`. I think that bit kinda defeats the point of the script.

What's the worst thing that can happen if we don't migrate some pool (with or without pool creator?). I'm leaning towards reviving the contract from V2 with stages, populating the inputs in the deployment script, and use it like that.

Then, migrate global percentages would be stage one, migrating (all) the hardcoded pools would be stage 2, and then finalize migration the stage 3.

jubeira (commented):
Thanks for putting this together so quickly @EndymionJkb.

Overall I think it looks good. I do have a few high level comments:
- I think reviving the script to migrate in stages would be a good idea here. Just set everything offchain, then hit migrate. Governance only grants admin permissions, and the script can only do what was indicated during deployment time. No possibility to leave permissions hanging, nor requirements around granting more permissions.
- I think there's an issue with the code introduced in `registerPool`. Not sure why it didn't make any test fail.
- The implementation for `migratePool` makes sense. Perhaps we can migrate the bulk of the pools as part of the script but leave that one open (i.e. permissionless). Given that it can only port pools from the old controller only once the behavior seems to be pretty well constrained.

pkg/governance-scripts/contracts/ProtocolFeeControllerMigrationV2.sol
EndymionJkb: Ok; was trying to avoid that, but it makes sense. Will port it over then.

pkg/vault/contracts/ProtocolFeeController.sol
EndymionJkb: I think I don't understand what you mean here; it's just emitting an extra event now.

pkg/governance-scripts/contracts/ProtocolFeeControllerMigrationV2.sol
EndymionJkb: Per discussion, looks like we need to keep the current design. The coordinator works for things that are deterministic (e.g., ordered set of immutable tasks, like upgrading a gauge adder). This depends on a dynamic set of pools that would surely change over the course of the governance vote, so it would be out-of-date by the time we could run it.

pkg/vault/contracts/ProtocolFeeController.sol
jubeira: lol I'm too used to the `if / revert` pattern; my brain read `revert` instead of `emit`.

Do we need to emit this conditionally? Why not just emit an event with the pool creator address?

EndymionJkb:
> I was thinking that if there are fees collected but not withdrawn in the old controller, probably the best way to get them is to just grant the permission to a multisig to pull them out. WDYT? We should probably include that detail in the notion docs as part of the 'user manual' for this migration.

I added a section on for post-migration considerations. The account that withdraws would already have permission on the old controller; could just leave it alone. The account could withdraw from both controllers if necessary. Nothing more can be collected post-migration, so the old controller would just be inert (to governance) after that. Pool creators could still withdraw forever, as they don't need permission.

joaobrunoah (approved):
LGTM!

PR #1305, opened 2025-02-18
Use swapFeeManager as owner in mev hook

### LINKED ISSUE #1300: Use swap fee manager as owner
- For amp factor (exclusive role)
- For mev tax hook (exclusive role, same pattern as stable surge)

elshan-eth:
# Description

This PR changes the authorization logic for mev hook: if swapFeeManager is set, use it as owner instead of governance.

## Type of change

- [ ] Bug fix 
- [ ] New feature 
- [ ] Breaking change 
- [ ] Dependency changes
- [ ] Code refactor / cleanup
- [ ] Other

## Checklist:

## Issue Resolution

Closes #1300

### DIFF
--- a/pkg/pool-hooks/test/foundry/MevCaptureHook.t.sol
+++ b/pkg/pool-hooks/test/foundry/MevCaptureHook.t.sol
@@ -5,7 +5,12 @@ pragma solidity ^0.8.24;
 import "forge-std/Test.sol";
 
 import { IAuthentication } from "@balancer-labs/v3-interfaces/contracts/solidity-utils/helpers/IAuthentication.sol";
-import { PoolSwapParams, MAX_FEE_PERCENTAGE } from "@balancer-labs/v3-interfaces/contracts/vault/VaultTypes.sol";
+import {
+    PoolSwapParams,
+    MAX_FEE_PERCENTAGE,
+    PoolRoleAccounts
+} from "@balancer-labs/v3-interfaces/contracts/vault/VaultTypes.sol";
+import { IVaultExplorer } from "@balancer-labs/v3-interfaces/contracts/vault/IVaultExplorer.sol";
 import { IMevCaptureHook } from "@balancer-labs/v3-interfaces/contracts/pool-hooks/IMevCaptureHook.sol";
 import { IHooks } from "@balancer-labs/v3-interfaces/contracts/vault/IHooks.sol";
 import { IVault } from "@balancer-labs/v3-interfaces/contracts/vault/IVault.sol";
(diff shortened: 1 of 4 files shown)


jubeira (commented):
LGTM; very clean!

This pattern is starting to get used everywhere, so it makes sense to put it in `SingletonAuthentication`.

jubeira (approved):

pkg/vault/contracts/SingletonAuthentication.sol
```diff
@@ -16,6 +16,11 @@ import { Authentication } from "@balancer-labs/v3-solidity-utils/contracts/helpe
 abstract contract SingletonAuthentication is Authentication {
     IVault private immutable _vault;
 
+    modifier onlySwapFeeManagerOrAuthentication(address pool) {
```
jubeira, line 19: Maybe `onlySwapFeeManagerOrAuthenticate` / `onlySwapFeeManagerXorAuthenticate` sounds a bit better (xor is what we are actually doing, although the name doesn't sound great. Leaving that to @EndymionJkb :) )

EndymionJkb: If we don't want to use exclusiveRole terminology, and have to make it swap-manager specific, maybe `onlySwapFeeManagerOrGovernance`, or `onlySwapFeeManagerIfDefined`?

I'd still prefer to have the `_ensureAuthenticatedByExclusiveRole` function parameterized by the role, and then give the modifiers more specific names, and have them call the generic role authentication (identical to VaultAdmin). It would be most consistent across the whole codebase. Otherwise we're making up different patterns and naming conventions when there's already a suitable one there.

EndymionJkb (commented):
Per comments on the other PR, I still suggest using a standard generic function `_ensureAuthenticatedByExclusiveRole` in the base contract (maybe SingletonAuthentication; wherever it makes sense), and then have either local modifiers, or if they're really all going to be swapFeeManager, the modifier could go up there, too, but separate from the generic authentication logic that just takes a "roleAddress" - in case we have something else that uses a different role later.

pkg/pool-hooks/test/foundry/MevCaptureHook.t.sol
```diff
@@ -5,7 +5,12 @@ pragma solidity ^0.8.24;
 import "forge-std/Test.sol";
 
 import { IAuthentication } from "@balancer-labs/v3-interfaces/contracts/solidity-utils/helpers/IAuthentication.sol";
-import { PoolSwapParams, MAX_FEE_PERCENTAGE } from "@balancer-labs/v3-interfaces/contracts/vault/VaultTypes.sol";
+import {
+    PoolSwapParams,
+    MAX_FEE_PERCENTAGE,
+    PoolRoleAccounts
+} from "@balancer-labs/v3-interfaces/contracts/vault/VaultTypes.sol";
+import { IVaultExplorer } from "@balancer-labs/v3-interfaces/contracts/vault/IVaultExplorer.sol";
```
joaobrunoah, line 13: I'd avoid using IVaultExplorer, and use IVaultExtension instead, for documentation. IVaultExplorer is a tool created to make it easy for users to navigate the vault through etherscan, I don't think we should use it on tests.

joaobrunoah (approved):
LGTM!

PR #1306, opened 2025-02-18
Use swapFeeManager as owner in stable pool

elshan-eth:
# Description

This PR changes the authorization logic for stable pool: if swapFeeManager is set, use it as owner instead of governance.

## Type of change

- [ ] Bug fix 
- [ ] New feature 
- [ ] Breaking change 
- [ ] Dependency changes
- [ ] Code refactor / cleanup
- [ ] Other

## Checklist:


pkg/pool-stable/contracts/StablePool.sol
```diff
@@ -116,19 +116,6 @@ contract StablePool is IStablePool, BalancerPoolToken, BasePoolAuthentication, P
...
-     * @dev Unlike the swap fee percentage setting permission, this is non-exclusive.
-     */
-    modifier authenticateByRole() {
-        // Allow if this is the swapFeeManager.
-        if (msg.sender != _vault.getPoolRoleAccounts(address(this)).swapFeeManager) {
-            // Otherwise, defer to governance.
-            _authenticateCaller();
-        }
-        _;
-    }
```
EndymionJkb, line 130:
```
modifier authenticateByExclusiveRole() {
    address roleAddress = _vault.getPoolRoleAccounts(address(this)).swapFeeManager;

    if (roleAddress == address(0)) {
        // Defer to governance if no role assigned.
        _authenticateCaller();
    } else if (msg.sender != roleAddress) {
        revert SenderNotAllowed();
    }
    _;
}
```

EndymionJkb: I would just follow the pattern established in `VaultAdmin`, and make the minimal change of replacing `authenticateByRole` with `authenticateByExclusiveRole`. It's specific to StablePool, so consider leaving it here vs. putting something up in BasePool that is swapFee-specific.

EndymionJkb (commented):
See comments for a simpler change

pkg/pool-stable/contracts/StablePool.sol
EndymionJkb: Ahh... forgot you're sharing it with the MEV hook. Still, I'd say use standard terminology, and maybe define this up in BasePool, and have the local modifiers call it. That would be cleanest.

```
function _ensureAuthenticatedByExclusiveRole(address pool, address roleAddress) private view {
    if (roleAddress == address(0)) {
        // Defer to governance if no role assigned.
        _ensureAuthenticated(pool);
    } else if (msg.sender != roleAddress) {
        revert SenderNotAllowed();
    }
}
```

jubeira: @EndymionJkb mev hook is not a pool, so it's slightly different. I think the current change is fine (and in this case we always use swap fee manager, so it's fine to have it in the name of the modifier)

pkg/pool-stable/test/foundry/StablePool.t.sol
```diff
@@ -172,6 +172,16 @@ contract StablePoolTest is BasePoolTest, StablePoolContractsDeployer {
...
+    function testStartAmplificationParameterUpdateIsPermissioned() public {
+        vm.expectRevert(IAuthentication.SenderNotAllowed.selector);
+        IStablePool(pool).startAmplificationParameterUpdate(DEFAULT_AMP_FACTOR * 2, block.timestamp + 5000 days);
+    }
+
+    function testStopAmplificationParameterUpdateIsPermissioned() public {
+        vm.expectRevert(IAuthentication.SenderNotAllowed.selector);
+        IStablePool(pool).stopAmplificationParameterUpdate();
+    }
+
```
jubeira, line 184: We should add a test where the swap fee manager performs the permissioned functions

Also maybe show that when there's a swap fee manager, the usual governance permission doesn't work

pkg/pool-utils/contracts/BasePoolAuthentication.sol
```diff
@@ -10,6 +10,14 @@ import { Authentication } from "@balancer-labs/v3-solidity-utils/contracts/helpe
 abstract contract BasePoolAuthentication is Authentication {
     IVault private immutable _vault;
 
+    /**
+     * @dev Allow only the swapFeeManager or an authenticated user to call the function.
+     */
+    modifier onlySwapFeeManagerOrAuthentication(address pool) {
```
jubeira, line 16: Same as in the other PR, maybe `onlySwapFeeManagerOrAuthenticate` / `onlySwapFeeManagerXorAuthenticate`

jubeira (approved):
LGTM, just minor comments around tests / naming

EndymionJkb (approved):
Some minor comment/import/test issues, but looks correct. Unfortunate that we're adding to the contract hierarchy depth, but we do have slightly different cases.

PR #1311, opened 2025-02-19
Add vault getter to aggregator router

jubeira:
# Description

For regular routers it's not much of a utility since if you interact with the router you don't even need to know about the vault.

But in the aggregator router you need to pay the vault directly, so a getter can be handy.

## Type of change

- [ ] Bug fix 
- [x] New feature 
- [ ] Breaking change 
- [ ] Dependency changes
- [ ] Code refactor / cleanup
- [ ] Other

## Checklist:

## Issue Resolution

N/A

### DIFF
--- a/pkg/vault/contracts/AggregatorRouter.sol
+++ b/pkg/vault/contracts/AggregatorRouter.sol
@@ -32,6 +32,11 @@ contract AggregatorRouter is IAggregatorRouter, SenderGuard, VaultGuard, Reentra
         // solhint-disable-previous-line no-empty-blocks
     }
 
+    /// @inheritdoc IAggregatorRouter
+    function getVault() public view returns (IVault) {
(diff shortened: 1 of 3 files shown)


elshan-eth (approved):
LGTM!

pkg/vault/contracts/AggregatorRouter.sol
```diff
@@ -32,6 +32,11 @@ contract AggregatorRouter is IAggregatorRouter, SenderGuard, VaultGuard, Reentra
         // solhint-disable-previous-line no-empty-blocks
     }
 
+    /// @inheritdoc IAggregatorRouter
+    function getVault() public view returns (IVault) {
```
EndymionJkb, line 36: I guess it's public vs. external so that it can be used in descendant contracts?

PR #1317, opened 2025-02-22
ProtocolFeeController v2 (working migration edition)

EndymionJkb:
# Description

So... turns out the original plan for two migration scripts (a first simple and a later more comprehensive one) doesn't actually work. We need the full migration even for the first time; the "update" functions must be called by the current fee controller, so can't be used during a migration.

It's actually simpler this way, as there's only one migration script now. The only catch is we have to tolerate the pool creator fee getters failing the first time, as they don't exist in the original fee controller. This is fine the first time, as there are no overrides or pool creators (both of which can be recreated in the new fee controller using permissioned calls from governance and the pool creators respectively, in case we miss one).

For the second and all subsequent migrations, they will work and copy the entire state.

I noticed that since we now derived from `SingletonAuthentication`, we have both `vault()` and `getVault()` defined, which of course do the same thing. We could theoretically remove `vault()` from the interface -- except it's being called by Gyro and CoW pools, so now we're stuck with it.

## Type of change

- [ ] Bug fix 
- [ ] New feature 
- [ ] Breaking change 
- [ ] Dependency changes
- [X] Code refactor / cleanup
- [ ] Other

## Checklist:

### DIFF
--- a/pkg/vault/contracts/ProtocolFeeController.sol
+++ b/pkg/vault/contracts/ProtocolFeeController.sol
@@ -463,8 +463,22 @@ contract ProtocolFeeController is
             isOverride: yieldFeeIsOverride
         });
 
-        _poolCreatorSwapFeePercentages[pool] = oldFeeController.getPoolCreatorSwapFeePercentage(pool);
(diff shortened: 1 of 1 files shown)


jubeira (approved):
LGTM!

joaobrunoah (approved):
LGTM!

PR #1320, opened 2025-02-25
DRAFT: Updating q

elshan-eth:
# Description

## Type of change

- [ ] Bug fix 
- [ ] New feature 
- [ ] Breaking change 
- [ ] Dependency changes
- [ ] Code refactor / cleanup
- [ ] Other

## Checklist:

### DIFF
--- a/pkg/pool-aclamm/contracts/AclAmmPool.sol
+++ b/pkg/pool-aclamm/contracts/AclAmmPool.sol
@@ -39,7 +39,14 @@ contract AclAmmPool is BalancerPoolToken, PoolInfo, Version, IBasePool, BaseHook
     uint256[] private _virtualBalances;
 
     uint256 private _c;
+
     uint256 private _sqrtQ0;
+    uint256 private _initialSqrtQ0;
+    uint256 private _targetSqrtQ0;
+
+    uint256 private _startChangingQ0Timestamp;
+    uint256 private _endChangingQ0Timestamp;
+
     uint256 private _centernessMargin;
 
     /// @dev Struct with data for deploying a new AclAmmPool.
@@ -88,7 +95,7 @@ contract AclAmmPool is BalancerPoolToken, PoolInfo, Version, IBasePool, BaseHook
             request.balancesScaled18,
             _virtualBalances,
             _c,
-            _sqrtQ0,
+            _updateAndGetSqrtQ0(),
             _lastTimestamp,
             _centernessMargin
         );
@@ -138,7 +145,7 @@ contract AclAmmPool is BalancerPoolToken, PoolInfo, Version, IBasePool, BaseHook
     /// @inheritdoc IHooks
     function onBeforeInitialize(uint256[] memory balancesScaled18, bytes memory) public override returns (bool) {
         _lastTimestamp = block.timestamp;
-        _virtualBalances = AclAmmMath.initializeVirtualBalances(balancesScaled18, _sqrtQ0);
+        _virtualBalances = AclAmmMath.initializeVirtualBalances(balancesScaled18, _updateAndGetSqrtQ0());
         return true;
     }
 
@@ -170,7 +177,7 @@ contract AclAmmPool is BalancerPoolToken, PoolInfo, Version, IBasePool, BaseHook
             balancesScaled18,
             _virtualBalances,
             _c,
-            _sqrtQ0,
+            _calculateCurrentQ0(),
             _lastTimestamp,
             _centernessMargin
         );
@@ -178,6 +185,44 @@ contract AclAmmPool is BalancerPoolToken, PoolInfo, Version, IBasePool, BaseHook
 
     function getLastTimestamp() external view returns (uint256) {
         return _lastTimestamp;
+    }
+
+    function setNewQ0(uint256 newQ0, uint256 startTime, uint256 endTime) external {
+        _initialSqrtQ0 = _updateAndGetSqrtQ0();
+        _targetSqrtQ0 = newQ0;
+        _startChangingQ0Timestamp = startTime;
+        _endChangingQ0Timestamp = endTime;
+    }
+
+    function _updateAndGetSqrtQ0() internal returns (uint256) {
+        uint256 storedSqrtQ0 = _sqrtQ0;
+        uint256 currentQ0 = _calculateCurrentQ0(storedSqrtQ0);
+        if (currentQ0 != storedSqrtQ0) {
+            _setSqrtQ0(currentQ0);
+        }
+
+        return currentQ0;
+    }
+
+    function _calculateCurrentQ0() internal view returns (uint256) {
+        return _calculateCurrentQ0(_sqrtQ0);
+    }
+
+    function _calculateCurrentQ0(uint256 storedSqrtQ0) internal view returns (uint256) {
+        uint256 currentTime = block.timestamp;
+        uint256 endTime = _endChangingQ0Timestamp;
+        uint256 startTime = _startChangingQ0Timestamp;
+        uint256 targetSqrtQ0 = _targetSqrtQ0;
+
+        if (currentTime > endTime && storedSqrtQ0 != targetSqrtQ0) {
+            return targetSqrtQ0;
+        } else if (currentTime < startTime) {
+            return storedSqrtQ0;
+        }
+
+        return
+            ((endTime - currentTime) * _initialSqrtQ0 + (currentTime - startTime) * targetSqrtQ0) /
+            (endTime - startTime);
     }
 
     function _setIncreaseDayRate(uint256 increaseDayRate) internal {
--- a/pkg/pool-aclamm/contracts/AclAmmPool.sol
+++ b/pkg/pool-aclamm/contracts/AclAmmPool.sol
@@ -75,7 +75,7 @@ contract AclAmmPool is BalancerPoolToken, PoolInfo, Version, IBasePool, BaseHook
                 balancesScaled18,
                 _virtualBalances,
                 _c,
-                _sqrtQ0,
+                _calculateCurrentQ0(),
                 _lastTimestamp,
                 _centernessMargin,
                 rounding


pkg/pool-aclamm/contracts/AclAmmPool.sol
```diff
@@ -180,6 +189,47 @@ contract AclAmmPool is BalancerPoolToken, PoolInfo, Version, IBasePool, BaseHook
         return _lastTimestamp;
     }
 
+    function setNewQ0(uint256 newQ0, uint256 startTime, uint256 endTime) external {
```
joaobrunoah, line 192: it should receive the new sqrtQ0, so you don't need to calculate the sqrt later. Especially because this is an 18 decimal number, Math.sqrt won't work.

pkg/pool-aclamm/contracts/AclAmmPool.sol
```diff
@@ -39,7 +41,14 @@ contract AclAmmPool is BalancerPoolToken, PoolInfo, Version, IBasePool, BaseHook
     uint256[] private _virtualBalances;
 
     uint256 private _c;
+
     uint256 private _sqrtQ0;
+    uint256 private _initSqrtQ0;
+    uint256 private _targetQ0;
+
+    uint256 private _startChangingQ0Timestamp;
+    uint256 private _endChangingQ0Timestamp;
```
joaobrunoah: StablePool has an AmplificationState, like this:

```
struct AmplificationState {
    uint64 startValue;
    uint64 endValue;
    uint32 startTime;
    uint32 endTime;
}
```

I believe this could resemble that struct. We don't need 3 sqrt!0, only two: startQ0 and endQ0.

When the pool is initialized, we can store the current sqrtQ0 as startQ0 and endQ0, and the endTime as the current timestamp.

pkg/pool-aclamm/contracts/AclAmmPool.sol
```diff
@@ -180,6 +189,47 @@ contract AclAmmPool is BalancerPoolToken, PoolInfo, Version, IBasePool, BaseHook
...
+
+    function _updateAndGetSqrtQ0() internal returns (uint256) {
+        uint256 storedSqrtQ0 = _sqrtQ0;
+        uint256 currentSqrtQ0 = _calculateCurrentSqrtQ0(storedSqrtQ0);
+        if (currentSqrtQ0 != storedSqrtQ0) {
+            _setSqrtQ0(currentSqrtQ0);
+        }
+
+        return currentSqrtQ0;
+    }
```
joaobrunoah, line 207: We don't need to update the _sqrtQ0 on every interaction. We can just calculate the current Q0 based on start and end values and times.

We should only change the storage when a pool manager wants to start updating the Q0 to another value, so the `Q0State` (Similar to StablePool's AmplificationState) needs to be updated.

pkg/pool-aclamm/contracts/AclAmmPool.sol
```diff
@@ -180,6 +189,47 @@ contract AclAmmPool is BalancerPoolToken, PoolInfo, Version, IBasePool, BaseHook
...
+        } else if (currentTime > endTime) {
+            return storedSqrtQ0;
+        }
+
+        return
+            Math.sqrt(
+                ((endTime - currentTime) * _initSqrtQ0 + (currentTime - startTime) * targetSqrtQ0) /
+                    (endTime - startTime)
+            );
+    }
```
joaobrunoah: There's no need for the _calculateCurrentSqrtQ0 to receive arguments. The argument will always be the current state of the Q0.

```suggestion
    function _calculateCurrentSqrtQ0() internal view returns (uint256) {
        uint256 currentTime = block.timestamp;
        uint256 endTime = _endChangingQ0Timestamp;
        uint256 startTime = _startChangingQ0Timestamp;
        uint256 initialSqrtQ0 = _sqrtQ0;
        uint256 targetSqrtQ0 = _targetSqrtQ0; // The pool manager will inform the sqrtQ0

        // If current time is bigger than end time, always return the targetQ0.
        if (currentTime > endTime {
            return targetSqrtQ0;
        }

        // No need to use FixedPoint math in the formula below; there won't be overflows.
        return
                ((endTime - currentTime) * initialSqrtQ0 + (currentTime - startTime) * targetSqrtQ0) /
                    (endTime - startTime);
    }
```

pkg/pool-aclamm/contracts/AclAmmPool.sol
joaobrunoah, line 207:
```suggestion
```

Actually, I just realised that this function is not needed.

pkg/pool-aclamm/contracts/AclAmmPool.sol
```diff
@@ -138,7 +147,7 @@ contract AclAmmPool is BalancerPoolToken, PoolInfo, Version, IBasePool, BaseHook
     /// @inheritdoc IHooks
     function onBeforeInitialize(uint256[] memory balancesScaled18, bytes memory) public override returns (bool) {
         _lastTimestamp = block.timestamp;
-        _virtualBalances = AclAmmMath.initializeVirtualBalances(balancesScaled18, _sqrtQ0);
+        _virtualBalances = AclAmmMath.initializeVirtualBalances(balancesScaled18, _updateAndGetSqrtQ0());
```
joaobrunoah, line 150:
```suggestion
        _virtualBalances = AclAmmMath.initializeVirtualBalances(balancesScaled18, _calculateCurrentSqrtQ0());
```

pkg/interfaces/contracts/pool-aclamm/IAclAmmPool.sol
```diff
@@ -0,0 +1,34 @@
+// SPDX-License-Identifier: GPL-3.0-or-later
```
joaobrunoah, line 1: Nice!

pkg/pool-aclamm/test/foundry/AclAmmMath.t.sol
```diff
@@ -103,4 +103,48 @@ contract AclAmmMathTest is Test {
...
+    ) public view {
+        endTime = bound(endTime, 2, type(uint64).max);
+        startTime = bound(startTime, 1, endTime - 1);
+        currentTime = bound(currentTime, startTime, endTime);
+
+        console.log("currentTime: %s", currentTime);
+        console.log("startSqrtQ0: %s", startSqrtQ0);
+        console.log("endSqrtQ0: %s", endSqrtQ0);
+        console.log("startTime: %s", startTime);
+        console.log("endTime: %s", endTime);
```
joaobrunoah, line 122: CL

pvt/helpers/src/math/aclAmm.ts
```diff
@@ -0,0 +1,19 @@
+import { BigNumberish } from 'ethers';
```
joaobrunoah, line 1: Cool!

pkg/pool-aclamm/test/AclAmmMath.test.ts
```diff
@@ -0,0 +1,49 @@
...
+import { expect } from 'chai';
+import { bn, fp } from '@balancer-labs/v3-helpers/src/numbers';
+import { expectEqualWithError } from '@balancer-labs/v3-helpers/src/test/relativeError';
+import { random } from 'lodash';
+import { calculateSqrtQ0 } from '@balancer-labs/v3-helpers/src/math/aclAmm';
+
+const MAX_RELATIVE_ERROR = 0.0001; // Max relative error
+
+// TODO: Test this math by checking extremes values for the amplification field (0 and infinite)
+// to verify that it equals constant sum and constant product (weighted) invariants.
```
joaobrunoah, line 12:
```suggestion
```

We can remove this

joaobrunoah (approved):
Just some minor comments, but LGTM! There's a logic when pool is "in range" and q0 is updating, that one is not set, but we can do that in the new repo, this PR already adds a lot by creating the interface and updating Q0 when the pool is "out of range".

pkg/pool-aclamm/contracts/lib/AclAmmMath.sol
```diff
@@ -112,6 +126,20 @@ library AclAmmMath {
             }
 
             changed = true;
+        } else if (isInRange && currentTime > sqrtQ0State.startTime && currentTime <= sqrtQ0State.endTime) {
```
joaobrunoah, line 129:
```suggestion
        } else if (currentTime > sqrtQ0State.startTime && currentTime <= sqrtQ0State.endTime) {
```

There's no need to check the inRange variable here, it's already checked in the line 107

pkg/pool-aclamm/contracts/lib/AclAmmMath.sol
```diff
@@ -84,13 +94,17 @@ library AclAmmMath {
...
+        SqrtQ0State memory sqrtQ0State
     ) internal view returns (uint256[] memory virtualBalances, bool changed) {
         // TODO Review rounding
 
         virtualBalances = new uint256[](balancesScaled18.length);
 
-        if (isPoolInRange(balancesScaled18, lastVirtualBalances, centernessMargin) == false) {
+        bool isInRange = isPoolInRange(balancesScaled18, lastVirtualBalances, centernessMargin);
+
+        if (isInRange == false) {
```
joaobrunoah, line 107:
```suggestion
        if (isPoolInRange(balancesScaled18, lastVirtualBalances, centernessMargin) == false) {
```

If we don't check isInRange in line 129, this code can be reduced to this.

pkg/pool-aclamm/contracts/lib/AclAmmMath.sol
```diff
@@ -112,6 +126,20 @@ library AclAmmMath {
...
+            uint256 currentSqrtQ0 = calculateSqrtQ0(
+                currentTime,
+                sqrtQ0State.startSqrtQ0,
+                sqrtQ0State.endSqrtQ0,
+                sqrtQ0State.startTime,
+                sqrtQ0State.endTime
+            );
+
+            virtualBalances[0] = rACenter.divDown(currentSqrtQ0 - FixedPoint.ONE);
+            virtualBalances[1] = rBCenter.divDown(currentSqrtQ0 - FixedPoint.ONE);
```
joaobrunoah, line 142: As I said in the call, we probably will need to review this math, but I'll do that and then we can reimplement in another task. Right now it's fine as is.

pkg/pool-aclamm/contracts/AclAmmPool.sol
```diff
@@ -170,22 +179,51 @@ contract AclAmmPool is BalancerPoolToken, PoolInfo, Version, IBasePool, BaseHook
...
+    /// @inheritdoc IAclAmmPool
+    function setSqrtQ0(
+        uint256 newSqrtQ0,
+        uint256 startTime,
+        uint256 endTime
+    ) external onlySwapFeeManagerOrGovernance(address(this)) {
+        _sqrtQ0State.startSqrtQ0 = _calculateCurrentSqrtQ0();
+        _sqrtQ0State.endSqrtQ0 = newSqrtQ0;
+        _sqrtQ0State.startTime = startTime;
+        _sqrtQ0State.endTime = endTime;
```
joaobrunoah, line 209: We should emit an event in this function, every time we change the sqrtQ0. The event should have start and end time and values.

Besides, I'd keep an internal _setSqrtQ0 function, and call it from the external setSqrtQ0, so we can use the internal function in the constructor as well.

pkg/pool-aclamm/contracts/AclAmmPool.sol
```diff
@@ -35,29 +45,24 @@ contract AclAmmPool is BalancerPoolToken, PoolInfo, Version, IBasePool, BaseHook
...
+    )
+        BalancerPoolToken(vault, params.name, params.symbol)
+        PoolInfo(vault)
+        BasePoolAuthentication(vault, msg.sender)
+        Version(params.version)
+    {
         _setIncreaseDayRate(params.increaseDayRate);
-        _setSqrtQ0(params.sqrtQ0);
+
+        _sqrtQ0State.endSqrtQ0 = params.sqrtQ0;
```
joaobrunoah, line 65: We should use the internal _setSqrtQ0 here, and pass 0 as start and endTime (or current timestamp), as well as params.sqrtQ0 as start and end value.

Doing that, we make sure the same event is emitted, which is better for offchain calculations.

pkg/pool-aclamm/contracts/lib/AclAmmMath.sol
```diff
@@ -145,6 +173,23 @@ library AclAmmMath {
...
     }
 
+    function calculateSqrtQ0(
+        uint256 currentTime,
+        uint256 startSqrtQ0,
+        uint256 endSqrtQ0,
+        uint256 startTime,
+        uint256 endTime
+    ) internal pure returns (uint256) {
+        if (currentTime >= endTime) {
```
joaobrunoah, line 183: In this function, we should probably check if currentTime < startTime and return startQ0. This condition should be tested.

pkg/pool-aclamm/test/foundry/AclAmmPoolVirtualBalances.t.sol
```diff
@@ -0,0 +1,258 @@
...
+        }
+
+        uint256[] memory newInitialBalances = new uint256[](2);
+        if (diffCoefficient > 0) {
+            newInitialBalances[0] = initialABalance * uint256(diffCoefficient);
+            newInitialBalances[1] = initialBBalance * uint256(diffCoefficient);
+        } else {
+            newInitialBalances[0] = initialABalance / uint256(-diffCoefficient);
+            newInitialBalances[1] = initialBBalance / uint256(-diffCoefficient);
+        }
```
joaobrunoah, line 58: It's not clear what you're trying to accomplish here. Can you add a comment?

To me, looks like you could just make a bound in the newInitialBalance[0], from initialABalance/100 to initialABalance * 100. (Same with newInitialBalance[1]). Something like the below. Does it work?

The nice part of doing it this way is that the coefficient of A and B balances are different, so you test more scenarios, but not sure if that's what you'd want.

```suggestion
    function testWithDifferentInitialBalances_Fuzz(uint256 _initialABalance, uint256 _initialBBalance) public {
        uint256[] memory newInitialBalances = new uint256[](2);
        newInitialBalances[0] = bound(_initialABalance, initialABalance/100, initialABalance*100);
        newInitialBalances[1] = bound(_initialBBalance, initialBBalance/100, initialBBalance*100);
```

pkg/pool-aclamm/test/foundry/AclAmmPoolVirtualBalances.t.sol
```diff
@@ -0,0 +1,258 @@
...
+            false,
+            new bytes(0)
+        );
+
+        uint256 invariantAfter = _getCurrentInvariant();
+        assertLt(invariantAfter, invariantBefore, "Invariant should decrease");
+
+        // uint256[] memory curentVirtualBalances = AclAmmPool(pool).getLastVirtualBalances();
+        // assertEq(curentVirtualBalances[0], virtualBalances[0], "Virtual A balances don't equal");
+        // assertEq(curentVirtualBalances[1], virtualBalances[1], "Virtual B balances don't equal");
```
joaobrunoah, line 234: Are these comments on purpose? Can we remove them?
If not, I'd leave a TODO here with some explanation on future work

pkg/pool-aclamm/test/foundry/AclAmmPoolVirtualBalances.t.sol
```diff
@@ -0,0 +1,258 @@
...
+            new bytes(0)
+        );
+
+        uint256 invariantAfter = _getCurrentInvariant();
+
+        assertGt(invariantAfter, invariantBefore, "Invariant should increase");
+
+        // uint256[] memory curentVirtualBalances = AclAmmPool(pool).getLastVirtualBalances();
+        // assertEq(curentVirtualBalances[0], virtualBalances[0], "Virtual A balances don't equal");
+        // assertEq(curentVirtualBalances[1], virtualBalances[1], "Virtual B balances don't equal");
```
joaobrunoah, line 210: Are these comments on purpose? Can we remove them?
If not, I'd leave a TODO here with some explanation on future work

joaobrunoah (approved):
Small comments, but overall it's good to merge. We need to figure out the logic of in range, but we can do that later.

pkg/pool-aclamm/test/foundry/AclAmmMath.t.sol
```diff
@@ -103,4 +103,42 @@ contract AclAmmMathTest is Test {
...
+        uint256 startSqrtQ0 = 100;
+        uint256 endSqrtQ0 = 200;
+        uint256 startTime = 0;
+        uint256 endTime = 50;
+        uint256 currentTime = 100;
+
+        uint256 sqrtQ0 = AclAmmMath.calculateSqrtQ0(currentTime, startSqrtQ0, endSqrtQ0, startTime, endTime);
+
+        assertEq(sqrtQ0, endSqrtQ0, "SqrtQ0 should be equal to endSqrtQ0");
+    }
```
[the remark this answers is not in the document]
elshan-eth: overflow error

pkg/pool-aclamm/contracts/AclAmmPool.sol
elshan-eth: I created another event for constructor 

`event AclAmmPoolInitialized(uint256 increaseDayRate, uint256 sqrtQ0, uint256 centernessMargin);`

pkg/pool-aclamm/test/foundry/AclAmmPoolVirtualBalances.t.sol
elshan-eth: Juani asked me to create a test that compares the virtual balances of two pools with different balances but the same q. The diffCoefficient allows for fuzz testing with different values. The values should maintain the same ratio, which is why diffCoefficient is used.

PR #1321, opened 2025-02-26
Fix documentation about salt used by factories

joaobrunoah:
# Description

The ECLP factory mentioned that salt was passed to create3, but it's used by create2. 
The stable pool only describes that salt is used for deployment and does not specify whether it's for create2 or create3, which seems more future-proof.

So, this PR adopts the Stable Pool Factory comment as the standard.

## Type of change

- [ ] Bug fix 
- [ ] New feature 
- [ ] Breaking change 
- [ ] Dependency changes
- [ ] Code refactor / cleanup
- [x] Documentation or wording changes
- [ ] Other

## Checklist:


EndymionJkb (approved):
Looks good! Yes, we keep flip-flopping on create2 vs. create3; best to keep the comments agnostic so that we don't need to maintain them.

jubeira (approved):
LGTM!

PR #1324, opened 2025-02-27
Refactor ERC165 import

EndymionJkb:
# Description

We were seeing a compilation error during contract verification, due to missing file references in the build-info. We are deliberately pruning the build-info in this step. The main purpose of this is to enforce the file order; otherwise, the build-info will differ slightly with each compilation, making it extremely difficult to review deployment preparation PRs, as you can't easily tell whether the build-info is identical. Also, it makes the file smaller by getting rid of mocks and other files that are used for testing but not in production. See [trimmedBuildInfoInput](https://github.com/balancer/balancer-deployments/blob/346f78f1df91b7c910adda8a8e258861b8d452d4/src/verifier.ts#L121:L139) in the deployments repo.

This is similar to Elshan's PR, in that it removes the ERC165 base contract from the burner contract (which I think is a good simplification regardless), but it does not duplicate the IERC165 interface locally. It just references the "missing" import in the file directly, so that it doesn't get trimmed. This allows the contract to be verified (see https://sepolia.etherscan.io/address/0x4636355cD210B651a2FA509288210f13CB2a3933#code).

## Type of change

- [ ] Bug fix 
- [ ] New feature 
- [ ] Breaking change 
- [X] Dependency changes
- [ ] Code refactor / cleanup
- [ ] Other

## Checklist:


elshan-eth (approved):
LGTM!

joaobrunoah (approved):
LGTM!

PR #1326, opened 2025-02-27
Remove unused state variable in StableSurgeHook

### LINKED ISSUE #1325: Missing state updates or optimizable state variables
Hi, we are a research group on programming languages and software engineering. We recently conducted a systematic study on the causes, effects, and fixes of the inconsistent state update vulnerability in solidity. We are attempting to build a tool to detect bugs about state updates based on our findings. We have tried our prototype tool on some popular Github solidity repositories, and for your repository, we found that there are missing state updates or gas consumption that can be optimized.

The point is that when we declare a state variable in the contract, if the variable is not reassigned throughout the project, it may be a missed state update, including balance, order number, counter, contract status flag, etc. Of course, it may also be a state variable with special purposes that does not need to be changed, such as maximum supply, contract administrator address, configuration information, etc. When declaring these state variables that do not need to be changed, the _constant_ or _immutable_ modifier should be used as required, which will save gas.

For your repository, we found the following state variable that may need attention. If you are not forgetting to update it, perhaps you should declare it with the _constant_ or _immutable_ modifier:

[StableSurgeHook.sol](https://github.com/balancer/balancer-v3-monorepo/blob/main/pkg/pool-hooks/contracts/StableSurgeHook.sol)
**State variable:** _surgeThresholdPercentage
```solidity
mapping(address pool => uint256 threshold) private _surgeThresholdPercentage;
```

Do you find our results useful? Your reply and invaluable suggestions would be greatly appreciated, and are vital for improving our tool. Thanks a lot for your time!
(The reason why _constant_ and _immutable_ modifiers can save gas is that they do not consume storage space within the EVM. Their values are compiled directly into smart contract bytecode, which reduces the gas cost of storage. This storage method also avoids the SLOAD operation that reads EVM storage (costs about 100 gas in EVM). The main difference between _constant_ and _immutable_ variables is that the value of _immutable_ variables can be set in the constructor, and _immutable_ variables may cost more gas than _constant_ variables. In addition, there is a slight difference in the variable types they support. The official documentation describes more details: https://docs.soliditylang.org/en/latest/contracts.html)

EndymionJkb:
# Description

Somehow this was missed in reviews... We replaced `_surgeThresholdPercentage ` with `_surgeFeePoolData`, but never removed the old mapping. It's not hurting anything (and we already deployed it like that), but might as well clean it up for the future. Thanks, @mingbaile!

## Type of change

- [ ] Bug fix 
- [ ] New feature 
- [ ] Breaking change 
- [ ] Dependency changes
- [X] Code refactor / cleanup
- [ ] Other

## Checklist:

## Issue Resolution

Resolves #1325

### DIFF
--- a/pkg/pool-hooks/contracts/StableSurgeHook.sol
+++ b/pkg/pool-hooks/contracts/StableSurgeHook.sol
@@ -88,9 +88,6 @@ contract StableSurgeHook is BaseHooks, VaultGuard, SingletonAuthentication {
         _;
     }
 
-    // Store the current threshold for each pool.
-    mapping(address pool => uint256 threshold) private _surgeThresholdPercentage;
-
     constructor(
         IVault vault,
         uint256 defaultMaxSurgeFeePercentage,


jubeira (approved):
LGTM!

joaobrunoah (approved):
LGTM!

PR #1330, opened 2025-03-05
Add initial global protocol fees to the ProtocolFeeController

EndymionJkb:
# Description

We had a request to incorporate default fees into the protocol fee controller, so that we don't need to use governance to set them after deployment.

Another issue related to new chains is the fact that the Protocol Fee Controller used to be deployed with the Vault in the Vault factory. This means we'd have to redeploy the Vault factory every time we migrated the fee controller. To avoid this, we pass the fee controller address to the Vault factory, so that it can be deployed externally and these two deployments are decoupled.

## Type of change

- [ ] Bug fix 
- [X] New feature 
- [ ] Breaking change 
- [ ] Dependency changes
- [ ] Code refactor / cleanup
- [ ] Other

## Checklist:

### DIFF
--- a/pkg/vault/contracts/VaultFactory.sol
+++ b/pkg/vault/contracts/VaultFactory.sol
@@ -6,6 +6,7 @@ import { Ownable2Step } from "@openzeppelin/contracts/access/Ownable2Step.sol";
 import { Ownable } from "@openzeppelin/contracts/access/Ownable.sol";
 import { Create2 } from "@openzeppelin/contracts/utils/Create2.sol";
 
+import { IProtocolFeeController } from "@balancer-labs/v3-interfaces/contracts/vault/IProtocolFeeController.sol";
 import { IAuthorizer } from "@balancer-labs/v3-interfaces/contracts/vault/IAuthorizer.sol";
 import { IVault } from "@balancer-labs/v3-interfaces/contracts/vault/IVault.sol";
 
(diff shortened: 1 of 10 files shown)


jubeira (approved):
LGTM, looks pretty reasonable to me.

This is kinda sensitive so the second review should be as thorough as possible too.

jubeira:
great job @EndymionJkb !

pkg/vault/contracts/VaultFactory.sol
```diff
@@ -107,13 +111,14 @@ contract VaultFactory is ReentrancyGuardTransient, Ownable2Step {
             revert InvalidBytecode("VaultExtension");
         }
 
-        address vaultAddress = getDeploymentAddress(salt);
-        if (targetAddress != vaultAddress) {
+        // If this passes, we know we can use the targetAddress for the vaultAddress.
+        if (targetAddress != getDeploymentAddress(salt)) {
```
joaobrunoah, line 115:
```suggestion
        // If this passes, we know we can use the targetAddress for the vaultAddress.
        if (targetAddress != getVaultDeploymentAddress(salt)) {
```

Probably, if we change the function name, the comment is not needed.

joaobrunoah (approved):
LGTM!

pkg/vault/contracts/VaultFactory.sol
EndymionJkb: Yes; that's what I had to do to solve the stack-too-deep error

PR #1331, opened 2025-03-20
Changing BigNumberish to bigint in math js tests

elshan-eth:
# Description

This PR removes the unnecessary cast BigNumbership to bigint in math js tests.

## Type of change

- [ ] Bug fix 
- [ ] New feature 
- [ ] Breaking change 
- [ ] Dependency changes
- [ ] Code refactor / cleanup
- [ ] Other

## Checklist:


EndymionJkb (approved):
Looks good. This PR is just scoped to the pool math functions, so I think it's fine.

The original purpose of using BigNumberish was to be flexible and allow strings and numerics to be interpreted without explicit casting.

If we no longer need that and want to banish it entirely, we could change this more extensively, but I suspect tests would start breaking if we did a global search/replace.

I'd say it's best to do it piecemeal, as needed.

jubeira (approved):
LGTM!

PR #1335, opened 2025-03-26
Wrapped BPT

### LINKED ISSUE #1332: BPT wrapper
Create a BPT wrapper that takes BPT 1:1, minting only when the vault is locked.

Suggested architecture, open to improvements:
- Wrapper factory (similar to gauge factories) deploys wrapper tokens (1 wrapper contract per BPT). BPT must be registered and shall be immutable in the wrapper. Consider using a transparent proxy pattern to make deployment of the wrappers cheaper.
- Tokens can be minted by depositing BPT while checking that the vault remains locked (otherwise revert)

Extras: add liquidity and wrap in router :) (or even use the factory as a router?)

elshan-eth:
# Description

The current BPT can be minted on credit as long as the vault is unlocked. This means we cannot use BPT as a token in a pool or in flash loan protocols. This PR fixes that and introduces a wrapped BPT (wBPT), which allows minting only when the Vault is locked.

## Type of change

- [ ] Bug fix 
- [ ] New feature 
- [ ] Breaking change 
- [ ] Dependency changes
- [ ] Code refactor / cleanup
- [ ] Other

## Checklist:

## Issue Resolution

Closes #1332

### DIFF
--- a/pkg/interfaces/contracts/vault/IWrappedBalancerPoolTokenFactory.sol
+++ b/pkg/interfaces/contracts/vault/IWrappedBalancerPoolTokenFactory.sol
@@ -1,0 +1,24 @@
+// SPDX-License-Identifier: GPL-3.0-or-later
+
+pragma solidity ^0.8.24;
+
+interface IWrappedBalancerPoolTokenFactory {
+    /// @notice The wrapped BPT already exists
+    error WrappedBPTAlreadyExists(address wrappedToken);
+
+    /// @notice The Balancer pool token has not been initialized
+    error BalancerPoolTokenNotInitialized();
+
+    /// @notice Wrapped Token was created
+    event WrappedTokenCreated(address indexed bpt, address wrappedToken);
+
+    /// @notice Creates a wrapped token for the given Balancer pool token
+    /// @param bpt The Balancer pool token to wrap
+    /// @return address The wrapped token address
+    function createWrappedToken(address bpt) external returns (address);
+
+    /// @notice Gets the wrapped token for the given Balancer pool token
+    /// @param bpt The Balancer pool token
+    /// @return address The wrapped token address
+    function getWrappedToken(address bpt) external view returns (address);
+}
--- a/pkg/interfaces/contracts/vault/IWrappedBalancerPoolTokenFactory.sol
+++ b/pkg/interfaces/contracts/vault/IWrappedBalancerPoolTokenFactory.sol
@@ -2,6 +2,7 @@
 
 pragma solidity ^0.8.24;
 
+/// @notice Interface for factory contract for creating wrapped Balancer pool tokens
 interface IWrappedBalancerPoolTokenFactory {
     /// @notice The wrapped BPT already exists
     error WrappedBPTAlreadyExists(address wrappedToken);
--- a/pkg/interfaces/contracts/vault/IWrappedBalancerPoolTokenFactory.sol
+++ b/pkg/interfaces/contracts/vault/IWrappedBalancerPoolTokenFactory.sol
@@ -11,15 +11,19 @@ interface IWrappedBalancerPoolTokenFactory {
     error BalancerPoolTokenNotInitialized();
 
     /// @notice Wrapped Token was created
-    event WrappedTokenCreated(address indexed bpt, address wrappedToken);
+    event WrappedTokenCreated(address indexed balancerPoolToken, address wrappedToken);
 
-    /// @notice Creates a wrapped token for the given Balancer pool token
-    /// @param bpt The Balancer pool token to wrap
-    /// @return address The wrapped token address
-    function createWrappedToken(address bpt) external returns (address);
+    /**
+     * @notice Creates a wrapped token for the given Balancer pool token
+     * @param balancerPoolToken The Balancer pool token to wrap
+     * @return address The wrapped token address
+     */
+    function createWrappedToken(address balancerPoolToken) external returns (address);
 
-    /// @notice Gets the wrapped token for the given Balancer pool token
-    /// @param bpt The Balancer pool token
-    /// @return address The wrapped token address
-    function getWrappedToken(address bpt) external view returns (address);
+    /**
+     * @notice Gets the wrapped token for the given Balancer pool token
+     * @param balancerPoolToken The Balancer pool token
+     * @return address The wrapped token address
+     */
+    function getWrappedToken(address balancerPoolToken) external view returns (address);
 }
--- a/pkg/vault/contracts/WrappedBalancerPoolToken.sol
+++ b/pkg/vault/contracts/WrappedBalancerPoolToken.sol
@@ -11,24 +11,26 @@ import { IVault } from "@balancer-labs/v3-interfaces/contracts/vault/IVault.sol"
 import { IWrappedBalancerPoolToken } from "@balancer-labs/v3-interfaces/contracts/vault/IWrappedBalancerPoolToken.sol";
 
 /**
- * @notice ERC20 wrapper for Balancer Pool Token (BPT), allowing users to deposit BPT
- * and receive 1:1 wrapped tokens, or burn wrapped tokens to redeem the original BPT.
+ * @notice ERC20 wrapper for Balancer Pool Token (BPT).
+ * @dev This allows users to deposit BPT and receive wrapped tokens 1:1, or burn wrapped tokens to redeem the original
+ * amount of BPT.
+ *
  * Minting and burning are only allowed when the Vault is locked.
  */
 contract WrappedBalancerPoolToken is IWrappedBalancerPoolToken, ERC20, ERC20Permit {
     using SafeERC20 for *;
 
-    IERC20 public immutable bpt;
+    IERC20 public immutable balancerPoolToken;
     IVault public immutable vault;
 
     constructor(
         IVault vault_,
-        IERC20 bpt_,
+        IERC20 balancerPoolToken_,
         string memory name,
         string memory symbol
     ) ERC20(name, symbol) ERC20Permit(name) {
         vault = vault_;
-        bpt = bpt_;
+        balancerPoolToken = balancerPoolToken_;
     }
 
     modifier onlyIfVaultLocked() {
@@ -40,7 +42,7 @@ contract WrappedBalancerPoolToken is IWrappedBalancerPoolToken, ERC20, ERC20Perm
 
     /// @inheritdoc IWrappedBalancerPoolToken
     function mint(uint256 amount) public onlyIfVaultLocked {
-        bpt.safeTransferFrom(msg.sender, address(this), amount);
+        balancerPoolToken.safeTransferFrom(msg.sender, address(this), amount);
 
         _mint(msg.sender, amount);
     }
@@ -60,6 +62,6 @@ contract WrappedBalancerPoolToken is IWrappedBalancerPoolToken, ERC20, ERC20Perm
     function _burnAndTransfer(address account, uint256 value) internal {
         _burn(account, value);
 
-        bpt.transfer(account, value);
+        balancerPoolToken.transfer(account, value);
     }
 }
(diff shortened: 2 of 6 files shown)


pkg/interfaces/contracts/vault/IWrappedBalancerPoolTokenFactory.sol
```diff
@@ -0,0 +1,25 @@
...
+
+    /// @notice The Balancer pool token has not been initialized
+    error BalancerPoolTokenNotInitialized();
+
+    /// @notice Wrapped Token was created
+    event WrappedTokenCreated(address indexed bpt, address wrappedToken);
+
+    /// @notice Creates a wrapped token for the given Balancer pool token
+    /// @param bpt The Balancer pool token to wrap
+    /// @return address The wrapped token address
```
jubeira, line 18: nit: use `/** */` for multiline comments

pkg/vault/contracts/WrappedBalancerPoolToken.sol
```diff
@@ -0,0 +1,65 @@
...
+
+/**
+ * @notice ERC20 wrapper for Balancer Pool Token (BPT), allowing users to deposit BPT
+ * and receive 1:1 wrapped tokens, or burn wrapped tokens to redeem the original BPT.
+ * Minting and burning are only allowed when the Vault is locked.
+ */
+contract WrappedBalancerPoolToken is IWrappedBalancerPoolToken, ERC20, ERC20Permit {
+    using SafeERC20 for *;
+
+    IERC20 public immutable bpt;
```
jubeira, line 21: I'd say let's use `balancerPoolToken` here as the name

pkg/vault/contracts/WrappedBalancerPoolToken.sol
```diff
@@ -0,0 +1,65 @@
...
+    modifier onlyIfVaultLocked() {
+        if (vault.isUnlocked()) {
+            revert VaultIsUnlocked();
+        }
+        _;
+    }
+
+    /// @inheritdoc IWrappedBalancerPoolToken
+    function mint(uint256 amount) public onlyIfVaultLocked {
+        bpt.safeTransferFrom(msg.sender, address(this), amount);
```
jubeira, line 43: Note: BPTs can't be _completely_ trusted.

It'd be a very weird edge case, since this contract is just to wrap certain BPTs that work in lending protocols, and as such they should just be standard implementations. I guess if there's something malicious about the transfer function the wrapper would just not make any sense either right? In other words, maybe this is just fine as is, and there's no need to add any extra protections.

pkg/vault/contracts/WrappedBalancerPoolToken.sol
```diff
@@ -0,0 +1,65 @@
...
+
+import { IVault } from "@balancer-labs/v3-interfaces/contracts/vault/IVault.sol";
+import { IWrappedBalancerPoolToken } from "@balancer-labs/v3-interfaces/contracts/vault/IWrappedBalancerPoolToken.sol";
+
+/**
+ * @notice ERC20 wrapper for Balancer Pool Token (BPT), allowing users to deposit BPT
+ * and receive 1:1 wrapped tokens, or burn wrapped tokens to redeem the original BPT.
+ * Minting and burning are only allowed when the Vault is locked.
+ */
+contract WrappedBalancerPoolToken is IWrappedBalancerPoolToken, ERC20, ERC20Permit {
```
jubeira, line 18: The most important invariant of this contract is that every time we mint / burn, `bpt.balanceOf(this) == totalSupply`. 

Do you think it's worth adding that (redundant) check after each operation? E.g.

```
modifier fullyBacked() {
    _;
    if (bpt.balanceOf(address(this)) < totalSupply()) {
        revert NotFullyBacked();
    }
}
```

pkg/vault/test/foundry/WrappedBalancerPoolToken.t.sol
```diff
@@ -0,0 +1,97 @@
...
+
+        uint256 balanceBefore = bpt.balanceOf(lp);
+
+        vm.startPrank(lp);
+        bpt.approve(address(wBPT), DEFAULT_AMOUNT);
+        wBPT.mint(DEFAULT_AMOUNT);
+        vm.stopPrank();
+
+        assertEq(bpt.balanceOf(lp), balanceBefore - DEFAULT_AMOUNT, "Invalid BPT balance");
+        assertEq(wBPT.balanceOf(lp), DEFAULT_AMOUNT, "Invalid wBPT balance");
```
jubeira, line 39: Also check that `wBPT.balanceOf(bpt) == wBPT.totalSupply()`

pkg/vault/test/foundry/WrappedBalancerPoolToken.t.sol
```diff
@@ -0,0 +1,97 @@
...
+        vm.startPrank(lp);
+        bpt.approve(address(wBPT), DEFAULT_AMOUNT);
+        wBPT.mint(DEFAULT_AMOUNT);
+
+        uint256 balanceBefore = bpt.balanceOf(lp);
+        wBPT.burn(DEFAULT_AMOUNT);
+        vm.stopPrank();
+
+        assertEq(bpt.balanceOf(lp), balanceBefore + DEFAULT_AMOUNT, "Invalid BPT balance");
+        assertEq(wBPT.balanceOf(lp), 0, "Invalid wBPT balance");
```
jubeira, line 61: Same as above; check total supply vs locked bpt inside wBPT.

pkg/vault/test/foundry/WrappedBalancerPoolToken.t.sol
```diff
@@ -0,0 +1,97 @@
...
+        uint256 balanceBefore = bpt.balanceOf(lp);
+
+        vm.prank(lp);
+        wBPT.approve(alice, DEFAULT_AMOUNT);
+
+        vm.prank(alice);
+        wBPT.burnFrom(lp, DEFAULT_AMOUNT);
+
+        assertEq(bpt.balanceOf(lp), balanceBefore + DEFAULT_AMOUNT, "Invalid BPT balance");
+        assertEq(wBPT.balanceOf(lp), 0, "Invalid wBPT balance");
```
jubeira, line 88: (same as above)

pkg/vault/contracts/WrappedBalancerPoolToken.sol
jubeira: note the use of `<` instead of `!=` as anyone can donate BPT to the wrapper contract. The point is that locked assets should be _at least_ 1:1 with the total supply; if we have more assets than total supply that's also fine.

jubeira (approved):
Great job @elshan-eth ! Very clean.

Just a few questions / extras. The extra check might be too much, but perhaps it's worth explicitly enforcing the invariant.

pkg/vault/contracts/WrappedBalancerPoolToken.sol
EndymionJkb: I think it is worth it for piece of mind, given that this shouldn't be a "critical path" function, so we don't care too much about gas (and the check is pretty cheap anyway). We'd just have to make sure legitimate deposits can't get locked somehow.

pkg/vault/contracts/WrappedBalancerPoolToken.sol
```diff
@@ -0,0 +1,65 @@
...
+    IVault public immutable vault;
+
+    constructor(
+        IVault vault_,
+        IERC20 bpt_,
+        string memory name,
+        string memory symbol
+    ) ERC20(name, symbol) ERC20Permit(name) {
+        vault = vault_;
+        bpt = bpt_;
```
EndymionJkb, line 31: Should we check that this is actually a BPT in the Vault? Redundant if it's deployed the intended way (from the factory), but somebody could just deploy it standalone.

Presumably lending protocols would verify that the Wrapped BPT token was deployed by the canonical factory anyway, like exchanges should do now for pools.

```
if (BalancerPoolToken(address(bpt_)).getVault() != vault_) {
    revert TokenIsNotBPT();
}
```

pkg/vault/contracts/WrappedBalancerPoolTokenFactory.sol
```diff
@@ -0,0 +1,49 @@
...
+
+import { WrappedBalancerPoolToken } from "./WrappedBalancerPoolToken.sol";
+
+/// @notice Factory contract for creating wrapped Balancer pool tokens
+contract WrappedBalancerPoolTokenFactory is IWrappedBalancerPoolTokenFactory {
+    IVault internal immutable _vault;
+    mapping(address => address) internal _wrappedTokens;
+
+    constructor(IVault vault_) {
+        _vault = vault_;
```
EndymionJkb, line 21:
```suggestion
    constructor(IVault vault) {
        _vault = vault;
```
Don't need the suffix in this case.

pkg/vault/contracts/WrappedBalancerPoolTokenFactory.sol
```diff
@@ -0,0 +1,49 @@
...
+    }
+
+    /// @inheritdoc IWrappedBalancerPoolTokenFactory
+    function createWrappedToken(address bpt) external returns (address) {
+        address wrappedToken = _wrappedTokens[bpt];
+        if (wrappedToken != address(0)) {
+            revert WrappedBPTAlreadyExists(wrappedToken);
+        }
```
joaobrunoah, line 29: What happens if the factory needs to be deprecated/updated? I wonder if this register shouldn't stay in the registry contract.

joaobrunoah (approved):
A minor comment, but LGTM!

pkg/vault/contracts/WrappedBalancerPoolTokenFactory.sol
```diff
@@ -0,0 +1,49 @@
...
+
+        if (!_vault.isPoolInitialized(address(bpt))) {
```
EndymionJkb, line 31: Could also check `isPoolRegistered` to make sure it's even a valid BPT for this Vault:

```
if (_vault.isPoolRegistered(bpt) == false) {
    revert TokenIsNotBPT();
} else if (_vault.isPoolInitialized(bpt) == false) {
    revert BalancerPoolTokenNotInitialized();
}
```
Also, the readability convention is "x == false" vs. "!x".

jubeira: Registration is a necessary condition for initialization, so this is a bit redundant if it's just to add a different error message.
I'd just check initialization.

pkg/vault/contracts/WrappedBalancerPoolToken.sol
jubeira: you can check whether it's registered here actually

jubeira: (more than `getVault`, just check if initialized here, as `bpt` can also lie about the vault, but not vice versa).

pkg/vault/contracts/WrappedBalancerPoolToken.sol
elshan-eth: I think this check is pointless, because if we can prove through tests that 1 wBPT = 1 BPT, then there won't be any cases where it's possible to break it.

Since BPT must be owned by the Vault, the only way to create malicious BPTs would be to make a pool where BPTs can be freely minted, but cannot be burned if they belong to another participant. That means WBPT would never actually reach the scenario where this fullyBacked condition fails.

pkg/vault/contracts/WrappedBalancerPoolToken.sol
elshan-eth: I think that if a contract wasn’t deployed through the factory, we can’t trust it at all. Yes, we could verify that the bytecode is a 1:1 match, but in that case, there’s no point in not using the factory for deployment. I’d prefer to keep the factory as the single source of truth for determining which wBPT is canonical.

PR #1341, opened 2025-04-01
Add disable flag for wrapped bpt factory

elshan-eth:
# Description

This PR adds the ability to stop the factory in case we suddenly decide that we need a new one. The chances of that happening are minimal, but I think it doesn't hurt to have this option.

## Type of change

- [ ] Bug fix 
- [X] New feature 
- [ ] Breaking change 
- [ ] Dependency changes
- [ ] Code refactor / cleanup
- [ ] Other

## Checklist:

### DIFF
--- a/pkg/vault/contracts/WrappedBalancerPoolTokenFactory.sol
+++ b/pkg/vault/contracts/WrappedBalancerPoolTokenFactory.sol
@@ -24,7 +24,7 @@ contract WrappedBalancerPoolTokenFactory is IWrappedBalancerPoolTokenFactory, Si
 
     modifier notDisabled() {
         if (_isDisabled) {
-            revert FactoryPaused();
+            revert Disabled();
         }
         _;
     }
@@ -56,8 +56,10 @@ contract WrappedBalancerPoolTokenFactory is IWrappedBalancerPoolTokenFactory, Si
     }
 
     /// @inheritdoc IWrappedBalancerPoolTokenFactory
-    function setDisabled(bool disabled) external {
-        _isDisabled = disabled;
+    function disable() external notDisabled authenticate {
+        _isDisabled = true;
+
+        emit FactoryDisabled();
     }
 
     /// @inheritdoc IWrappedBalancerPoolTokenFactory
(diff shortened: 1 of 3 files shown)


EndymionJkb (commented):
Before I look in detail, do we want this to work like the pool factories (IBasePoolFactory)? That's a one-way switch: `isDisabled` queries the state, and you call `disable` to emit the `FactoryDisabled` event and make `create` revert with `Disabled`.

If we want a one-way switch, I think it should parallel `IBasePoolFactory`. If we want it reversible, the example of that is in `VaultAdmin`, where we have `disableQuery` and `disableQueryPermanently`.

It seems like that's overkill for this, though, since unlike with the Vault, we can redeploy this. It seems like a one-way switch is good enough, like the pools have. I can't think of a reason why it would be reversed. And if it's reversible, what if you disabled it, redeployed, then re-enabled the original one?

elshan-eth:
@EndymionJkb, I agree with you, one-way switch sounds better. Thanks!

pkg/vault/contracts/WrappedBalancerPoolTokenFactory.sol
```diff
@@ -10,31 +10,39 @@ import {
...
+    modifier notDisabled() {
+        if (_isDisabled) {
+            revert Disabled();
+        }
+        _;
     }
 
     /// @inheritdoc IWrappedBalancerPoolTokenFactory
-    function createWrappedToken(address balancerPoolToken) external returns (address) {
+    function createWrappedToken(address balancerPoolToken) external notDisabled authenticate returns (address) {
```
jubeira, line 33: I think I'd keep wrapper creation permissionless; it's pretty harmless really. Do you see any reason to make it permissioned?

jubeira (commented):
I'm not against this change, but I'm a bit curious about the reasoning behind it.

As it is, the wrappers are pretty harmless; they only matter to the integrations that use them to build something, and presumably they will only accept certain wrappers (i.e. a lending protocol will only allowlist a few wrappers to be used as collateral, not anything that comes out of this factory). In that sense, I don't quite see what we would be preventing by preventing new wrappers from being created.

It'd be another story if the `disable` flag also prevented new mints for existing tokens, but I'm not sure we want to have that kind of power.

PR #1343, opened 2025-04-02
Protocol fee helper

EndymionJkb:
# Description

This follows the pattern of the PauseHelper, but works with protocol fees.

I noticed a typo/inconsistency in the PauseHelper (already merged, but not deployed), so fixed that in both.

I don't really like the "pagination" interface, and argued during review that it should match `IBasePoolFactory`:
    function getPoolCount();
    function getPoolsInRange(uint256 start, uint256 count);
    function getPools();

Instead of how it is:
    function getPoolsCount(); // modified to `getPoolCount`
    function hasPool(address pool);
    function getPools(uint256 from, uint256 to);

Since now we have two similar but slightly different interfaces. Adding `hasPool` is fine. If we think there will be too many pools to support `getPools` returning the set directly, we could leave that off and rename `getPools` to `getPoolsInRange` for consistency.

I didn't change anything (except `getPoolsCount` -> `getPoolCount`), but we could discuss it before deployment. (Not very important of course, but consistency across interfaces does improve devUX.)

## Type of change

- [ ] Bug fix 
- [X] New feature 
- [ ] Breaking change 
- [ ] Dependency changes
- [ ] Code refactor / cleanup
- [ ] Other

## Checklist:

- [X] The diff is legible and has no extraneous changes (see above)


pkg/standalone-utils/contracts/ProtocolFeeHelper.sol
```diff
@@ -0,0 +1,115 @@
...
+    constructor(IVault vault) SingletonAuthentication(vault) {
+        // solhint-disable-previous-line no-empty-blocks
+    }
+
+    /***************************************************************************
+                                    Manage Pools
+    ***************************************************************************/
+
+    /// @inheritdoc IProtocolFeeHelper
+    function addPools(address[] calldata newPools) external authenticate {
```
jubeira, line 32: Maybe I'd check that the pool address is a registered pool here

jubeira (approved):
LGTM! Just one suggestion to make `addPools` a bit more robust.

PR #1344, opened 2025-04-02
Adds flags to disable benchmark tests

joaobrunoah:
# Description

This change introduces flags (disableUnbalancedLiquidityTests, disableDonationTests) to disable specific benchmark tests. These flags allow for more granular control over which tests are executed during benchmarking.

This PR is a requirement to create benchmark tests on the ReClamm repo.

## Type of change

- [ ] Bug fix 
- [x] New feature 
- [ ] Breaking change 
- [ ] Dependency changes
- [ ] Code refactor / cleanup
- [ ] Other

## Checklist:


pvt/benchmarks/src/PoolBenchmark.behavior.ts
```diff
@@ -48,12 +48,14 @@ export type TestsAddLiquidityHooks = {
 
 export type TestSettings = {
   offNestedPoolTests: boolean;
```
EndymionJkb, line 50: I never noticed this; it's kind of bizarre English. Should be either "turnOffXXXTests" or "XXXTestsDisabled"; something like that.

joaobrunoah: I renamed to disable...Tests, is that ok?

pvt/benchmarks/src/PoolBenchmark.behavior.ts
```diff
@@ -443,6 +493,10 @@ export class Benchmark {
       });
 
       it(`measures gas (unbalanced) (${poolTag})`, async () => {
+        if (this._settings.disableUnbalancedLiquidityTests) {
+          return;
```
jubeira, line 497: here and elsewhere: I think you can do `this.skip()` inside `it` in mocha: https://stackoverflow.com/a/32820119

That will show the test as skipped when you run it instead as successful, which reflects what we're doing a bit better.

jubeira (commented):
Just a few comments, but I think the overall idea is fine :+1:

pvt/benchmarks/src/PoolBenchmark.behavior.ts
```diff
@@ -442,179 +492,191 @@ export class Benchmark {
         await saveSnap(this._testDirname, `[${this._poolType} - ${poolTag}] add liquidity proportional`, [receipt]);
       });
 
-      it(`measures gas (unbalanced) (${poolTag})`, async () => {
-        const exactAmountsIn = Array(poolInfo.poolTokens.length)
-          .fill(TOKEN_AMOUNT)
-          .map((amount, index) => BigInt(amount / BigInt(index + 1)));
+      (this._settings.disableUnbalancedLiquidityTests ? it.skip : it)(
```
jubeira, line 495: ah, `skip` didn't work inside `it`, did it?

This is fine but it's a bit complicated to read. The answer linked above mentions that if you declare the function explicitly it should work:

```
it(`measures gas (unbalanced) (${poolTag})`, async function {
    if (this._settings.disableUnbalancedLiquidityTests) {
        this.skip();
    }
    ...
});
```

pvt/benchmarks/src/PoolBenchmark.behavior.ts
```diff
@@ -47,13 +47,28 @@ export type TestsAddLiquidityHooks = {
...
+  disableNestedPoolTests?: boolean;
+  disableUnbalancedLiquidityTests?: boolean;
+  disableDonationTests?: boolean;
+  enableCustomSwapTests?: boolean;
+};
+
+export type TestCustomSwapsParams = {
+  router: Router;
+  batchRouter: BatchRouter;
+  alice: SignerWithAddress;
```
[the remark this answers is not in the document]
joaobrunoah: I removed the struct altogether, it wasn't working properly. Now I use the functions from Benchmark class directly.

joaobrunoah:
> Probably the best we can do. Just one minor nit question; I'd say after addressing Juani's skip suggestion it's good to go.
> 
> I'm getting an internal compiler error when I try to run benchmarks, though. Apparently just me?

Fixed!

jubeira (approved):
LGTM, looks clean like this! I'd just suggest adding a comment about the new closure; otherwise it's good to go.

elshan-eth (approved):
LGTM!

Issue #1345, opened 2025-04-04
Factor out common interface for pause and related helpers


EndymionJkb:
We have the pause helper, protocol fee helper, and swap fee helper, which are all very similar. We can factor out a lot of the common code and make a common interface for helper contracts of this type.

I'd also like to change the iteration interface to match `IBasePoolFactory` in the process; up for discussion.

PR #1346, opened 2025-04-04
Swap fee helper

### LINKED ISSUE #1337: Swap fee manager
Besides #897, we could use a swap fee manager contract following the same pattern as the pause helper in #1211

EndymionJkb:
# Description

Following the pattern of the pause helper contract, this is that for setting swap fees. The slight twist is that this only works for pools that *don't* have a swap manager role set, since that role is exclusive and cuts out governance.

It occurs to me that since we now have three of these things, we could probably factor out some common errors and events, and have a common interface except for the permissioned functions. See issue #1345.

## Type of change

- [ ] Bug fix 
- [X] New feature 
- [ ] Breaking change 
- [ ] Dependency changes
- [ ] Code refactor / cleanup
- [ ] Other

## Checklist:

## Issue Resolution

Closes #1337


pkg/standalone-utils/contracts/PoolSwapFeeHelper.sol
```diff
@@ -0,0 +1,106 @@
...
+        for (uint256 i = from; i < to; i++) {
+            pools[i - from] = _pools.at(i);
+        }
+    }
+
+    /***************************************************************************
+                                Internal functions                                
+    ***************************************************************************/
+
+    function _ensureKnownPool(address pool) internal view {
```
jubeira, line 101: how about `ensureAddedPool` / `ensurePoolAdded`? Would be consistent with `addPools`.

jubeira (approved):
LGTM, very clean!

PR #1352, opened 2025-04-16
Replace `isPoolInitialized` for `isPoolRegistered` in BPT wrapper factory

jubeira:
# Description

While it's technically a bit better to use the `initialized` flag as you really can't mint any wrapped BPT without the pool being initialized, it's harder to verify the wrapper contract onchain automatically like this, as we'd need the address of an initialized pool on every network.

By using the `registered` flag, we can just get pool mock addresses, deploy a mock bpt wrapper and verify it automatically in the deployment script.

I don't think this adds any risk, as you still need to initialize the pool at some point to be able to mint wrapped shares anyways. This just allows creating the wrapper contract earlier on.

## Type of change

- [ ] Bug fix 
- [ ] New feature 
- [ ] Breaking change 
- [ ] Dependency changes
- [ ] Code refactor / cleanup
- [x] Other

## Checklist:

## Issue Resolution

N/A

### DIFF
--- a/pkg/interfaces/contracts/vault/IWrappedBalancerPoolTokenFactory.sol
+++ b/pkg/interfaces/contracts/vault/IWrappedBalancerPoolTokenFactory.sol
@@ -12,8 +12,8 @@ interface IWrappedBalancerPoolTokenFactory {
      */
     error WrappedBPTAlreadyExists(address wrappedToken);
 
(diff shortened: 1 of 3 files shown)


EndymionJkb (approved):
LGTM. I think we talked about this at the time, though we weren't thinking of deployment issues.

PR #1353, opened 2025-04-16
Add aggregator batch router

### LINKED ISSUE #1333: Aggregator batch router
Modify batch router with construction-time constants to decide how to settle.

In 'aggregator mode', just assume that the aggregator already paid amounts in upfront (or max amount in for exact out, then return the difference); skip all permit2 transfers. In regular / permit2 mode, do what we do now.

elshan-eth:
# Description

This PR adds the Aggregator Batch Router where the sender pays upfront.

## Type of change

- [ ] Bug fix 
- [ ] New feature 
- [ ] Breaking change 
- [ ] Dependency changes
- [ ] Code refactor / cleanup
- [ ] Other

## Checklist:

## Issue Resolution

Closes #1333


pkg/vault/contracts/AggregatorBatchRouter.sol
```diff
@@ -0,0 +1,476 @@
...
+            amountsOut[i] = _currentSwapTokenOutAmounts().tGet(tokensOut[i]);
+        }
+    }
+
+    function _computePathAmountsOut(
+        SwapExactInHookParams calldata params
+    ) internal returns (uint256[] memory pathAmountsOut) {
+        pathAmountsOut = new uint256[](params.paths.length);
+
+        if (EVMCallModeHelpers.isStaticCall() == false) {
```
EndymionJkb, line 158: Should comment this block, as this is the different part compared to the regular BatchRouter.

EndymionJkb (commented):
First pass. It looks "symmetrical," but I will have to draw it out and really trace it through to be sure it's doing the right thing.

Would it be possible to share test code (e.g., batchRouterBehavior.ts), so it's clear that all the same tests pass between the Batch and AggregatorBatch?

I also wonder if we could likewise move some shared code up into `BatchRouterCommon`? That would highlight the differences, and increase conference that there aren't subtle/unintended copy/paste changes.

More internal comments would help generally.

elshan-eth:
@EndymionJkb, We can extract the common parts of the code and make the tests more generic, but it will require additional time investment.

@EndymionJkb @jubeira, Do you think it's worth it?

pkg/vault/contracts/AggregatorBatchRouter.sol
```diff
@@ -0,0 +1,476 @@
...
+                ),
+                (uint256[], address[], uint256[])
+            );
+    }
+
+    /// @inheritdoc IBatchRouter
+    function swapExactOut(
+        SwapPathExactAmountOut[] memory paths,
+        uint256 deadline,
+        bool wethIsEth,
```
jubeira, line 90: (same as exact in)

pkg/vault/contracts/AggregatorBatchRouter.sol
```diff
@@ -0,0 +1,476 @@
...
+    function _computePathAmountsOut(
+        SwapExactInHookParams calldata params
+    ) internal returns (uint256[] memory pathAmountsOut) {
+        pathAmountsOut = new uint256[](params.paths.length);
+
+        if (EVMCallModeHelpers.isStaticCall() == false) {
+            for (uint256 i = 0; i < params.paths.length; ++i) {
+                SwapPathExactAmountIn memory path = params.paths[i];
+                _currentSwapTokensIn().add(address(path.tokenIn));
+                _currentSwapTokenInAmounts().tAdd(address(path.tokenIn), path.exactAmountIn);
```
jubeira, line 160: I'd clarify that these are the amounts expected to be paid by the sender upfront, and that we're registering them as settled here.

pkg/vault/contracts/AggregatorBatchRouter.sol
```diff
@@ -0,0 +1,476 @@
...
+            address[] memory tokensIn = _currentSwapTokensIn().values();
+            for (uint256 i = 0; i < tokensIn.length; ++i) {
+                uint256 amount = _currentSwapTokenInAmounts().tGet(tokensIn[i]);
+                uint256 tokenInCredit = _vault.settle(IERC20(tokensIn[i]), amount);
+                if (tokenInCredit < amount) {
+                    revert InsufficientPayment();
+                }
+
+                _currentSwapTokenInAmounts().tSet(tokensIn[i], 0);
+            }
```
jubeira: This logic looks fine.

An alternative would be to just call settle for the token in for each of the paths though. The current approach is better if you have many paths using the same token in, as you minimize the calls to `vault.settle`. But if you don't have that situation, it might be better to just call `vault.settle` for each token in and call it a day. You wouldn't need to update token in amounts, tokens in, etc.

WDYT?

pkg/vault/contracts/AggregatorBatchRouter.sol
```diff
@@ -0,0 +1,476 @@
...
+                            kind: SwapKind.EXACT_IN,
+                            direction: step.pool == address(stepTokenIn)
+                                ? WrappingDirection.UNWRAP
+                                : WrappingDirection.WRAP,
+                            wrappedToken: IERC4626(step.pool),
+                            amountGivenRaw: stepExactAmountIn,
+                            limitRaw: minAmountOut
+                        })
+                    );
+                } else {
```
jubeira, line 213: Maybe we should require that `step.pool != stepTokenIn` and `step.pool != step.tokenOut` if it's not a buffer operation? Those are really not supported and will revert when calling `swap` with a not-so-nice error.

pkg/vault/contracts/AggregatorBatchRouter.sol
```diff
@@ -0,0 +1,476 @@
...
+                            limitRaw: stepMaxAmountIn,
+                            userData: params.userData
+                        })
+                    );
+                }
+
+                if (stepLocals.isLastStep) {
+                    pathAmountsIn[i] = amountIn;
+                    _currentSwapTokensOut().add(address(stepTokenIn));
+                    _currentSwapTokenOutAmounts().tAdd(address(stepTokenIn), path.maxAmountIn - amountIn);
```
jubeira, line 379: I'd clarify that this is to reimburse the difference between max amount in and actual amount in, and we're just adding it to the token out amounts to process altogether at settlement time (which I think is a good idea).

jubeira (commented):
First pass done!

Overall it makes sense. I've left a bunch of comment to polish some rough edges and add some extra docs here and there; there's a lot going on and I'd rather have some extra notes around.

I'll do a second pass and look at the tests after those are addressed. Good job!

pkg/vault/contracts/AggregatorBatchRouter.sol
elshan-eth: Hmm, then we would need to limit the case when multiple same tokens are used, otherwise there will be an error if we try to settle the same token twice.

pkg/vault/contracts/AggregatorBatchRouter.sol
```diff
@@ -0,0 +1,476 @@
...
+
+    /***************************************************************************
+                                       Swaps
+    ***************************************************************************/
+
+    /// @inheritdoc IBatchRouter
+    function swapExactIn(
+        SwapPathExactAmountIn[] memory paths,
+        uint256 deadline,
+        bool wethIsEth,
```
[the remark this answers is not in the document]
elshan-eth: Then I would need to create a separate interface IAggregatorBatchRouter, and use all the structures through IBatchRouter. I'm not sure if this is a good solution, so I would just add a revert if this flag is used.

EndymionJkb (approved):
Love the test simplification! 🚢

PR #1354, opened 2025-04-17
Alternate Balancer Contract Registry Initializer

EndymionJkb:
# Description

This is an alternate version of the initializer which requires explicit permissions to be granted (only two), vs. the historical pattern of granting admin permission that is self-revoked after the migration (or initialization in this case).

## Type of change

- [ ] Bug fix 
- [ ] New feature 
- [ ] Breaking change 
- [ ] Dependency changes
- [X] Code refactor / cleanup
- [ ] Other

## Checklist:


pkg/governance-scripts/contracts/BalancerContractRegistryInitializer.sol
```diff
@@ -111,11 +99,5 @@ contract BalancerContractRegistryInitializer {
         for (uint256 i = 0; i < _aliasNames.length; ++i) {
             balancerContractRegistry.addOrUpdateBalancerContractAlias(_aliasNames[i], _aliasAddresses[i]);
         }
-
-        // Renounce all roles.
-        _authorizer.renounceRole(registerContractRole, address(this));
-        _authorizer.renounceRole(addAliasRole, address(this));
```
jubeira, line 117: we should still renounce these two; I wouldn't keep them active.

pkg/governance-scripts/contracts/BalancerContractRegistryInitializer.sol
```diff
@@ -78,17 +77,6 @@ contract BalancerContractRegistryInitializer {
...
-        // Grant permissions to register contracts and add aliases.
-        bytes32 registerContractRole = IAuthentication(address(balancerContractRegistry)).getActionId(
-            IBalancerContractRegistry.registerBalancerContract.selector
-        );
-        bytes32 addAliasRole = IAuthentication(address(balancerContractRegistry)).getActionId(
-            IBalancerContractRegistry.addOrUpdateBalancerContractAlias.selector
-        );
-
-        _authorizer.grantRole(registerContractRole, address(this));
-        _authorizer.grantRole(addAliasRole, address(this));
```
jubeira, line 90: We start with the assumption that the permissions have already been granted. If they are not, the script will fail with `SENDER_NOT_ALLOWED`.

That's kinda fine; another option would be to check that the script effectively has the permissions before executing them and reverting with a custom error. I don't have a strong opinion.

jubeira (commented):
Thanks @EndymionJkb; just a few comments, otherwise looks good!

pkg/governance-scripts/contracts/BalancerContractRegistryInitializer.sol
EndymionJkb: I guess it depends on their model. Would they grant, execute and revoke in one transaction? If the contract revokes them, we'd need them to know they only need to grant. I can make it do that.

pkg/governance-scripts/contracts/BalancerContractRegistryInitializer.sol
EndymionJkb: If we're adopting this as a new pattern, a custom error isn't a bad idea.

pkg/governance-scripts/contracts/BalancerContractRegistryInitializer.sol
```diff
@@ -70,14 +73,21 @@ contract BalancerContractRegistryInitializer {
...
+     * to be done externally.
+     */
     function initializeBalancerContractRegistry() external {
         // Explicitly ensure this can only be called once.
         if (_initialized) {
             revert AlreadyInitialized();
         }
 
-        _initialized = true;
-
```
jubeira, line 80: this really doesn't change much.

Yes, it saves some gas if you can't perform one of the actions, but as a general rule it's not important to optimize for that case. It's better to keep it clear and straightforward.

In this case the change is also fine, no need to change it back.

pkg/governance-scripts/contracts/BalancerContractRegistryInitializer.sol
```diff
@@ -86,11 +96,25 @@ contract BalancerContractRegistryInitializer {
...
+        uint256 numPoolFactories = _poolFactoryNames.length;
+        uint256 numRouters = _routerNames.length;
+        uint256 numAliases = _aliasNames.length;
+
+        if (
+            ((numRouters > 0 || numPoolFactories > 0) &&
+                _authorizer.canPerform(registerContractRole, address(this), address(balancerContractRegistry)) ==
+                false) ||
+            (numAliases > 0 &&
+                _authorizer.canPerform(addAliasRole, address(this), address(balancerContractRegistry)) == false)
```
jubeira, line 109: Do you think it's worth adding a getter that returns `true` if both permissions are granted and `false` otherwise?

you can always check directly on the authorizer though, maybe too much.

jubeira (approved):
LGTM!

pkg/governance-scripts/contracts/BalancerContractRegistryInitializer.sol
EndymionJkb: It's possible they don't need to grant both permissions. Maybe they don't want to use aliases (or not on a certain chain). Or maybe they *only* need aliases, because they already manually registered the contracts before this was deployed.

PR #1358, opened 2025-04-23
Fix E2E Swap test

joaobrunoah:
# Description

This PR fixes two E2E Swap errors:
1. When the swap is Exact Out, in a Do/Undo test, sometimes the fee will be large and the exact amount in calculated in the Do path won't be available in the pool when executing the Undo part of the test.
2. In the Fee specific test, for Exact In, sometimes the amount out calculated is 0. These tests should be ignored.

Moreover, the ReClamm requires some tests to have a slightly bigger error tolerance.

## Type of change

- [x] Bug fix 
- [ ] New feature 
- [ ] Breaking change 
- [ ] Dependency changes
- [ ] Code refactor / cleanup
- [ ] Other

## Checklist:

### DIFF
--- a/pkg/vault/test/foundry/E2eSwap.t.sol
+++ b/pkg/vault/test/foundry/E2eSwap.t.sol
@@ -801,6 +801,8 @@ contract E2eSwapTest is BaseVaultTest {
         }
 
         {
+            // `exactAmountInDo` could be bigger than the actual balance of the pool, since the first swap paid fees
+            // and pool creator is 100% (no LP fees).
             BaseVaultTest.Balances memory balancesMiddle = getBalances(sender, Rounding.ROUND_DOWN);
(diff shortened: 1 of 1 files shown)


elshan-eth (approved):

pkg/vault/test/foundry/E2eSwap.t.sol
```diff
@@ -798,6 +800,11 @@ contract E2eSwapTest is BaseVaultTest {
             );
         }
 
+        {
+            BaseVaultTest.Balances memory balancesMiddle = getBalances(sender, Rounding.ROUND_DOWN);
```
jubeira, line 804: Let's add a small comment here explaining why we need this (fees / need enough tokens to do the roundrip, etc)

jubeira (approved):
LGTM, good catch! Just one minor comment

PR #1362, opened 2025-04-25
Factor common code out of helpers

### LINKED ISSUE #1345: Factor out common interface for pause and related helpers

EndymionJkb:
# Description

We have three nearly identical helpers that share most of their code. This factors out everything common, so that the helpers really just add their particular function, and it's really easy to make more if we need to.

Could probably factor things out of tests, too, but leaving them to ensure this doesn't break anything.

## Type of change

- [ ] Bug fix 
- [ ] New feature 
- [ ] Breaking change 
- [ ] Dependency changes
- [X] Code refactor / cleanup
- [ ] Other

## Checklist:

## Issue Resolution

Resolves #1345.

### DIFF
--- a/pkg/standalone-utils/contracts/PoolPauseHelper.sol
+++ b/pkg/standalone-utils/contracts/PoolPauseHelper.sol
@@ -8,14 +8,12 @@ import { IPoolPauseHelper } from "@balancer-labs/v3-interfaces/contracts/standal
 import { IVaultAdmin } from "@balancer-labs/v3-interfaces/contracts/vault/IVaultAdmin.sol";
 import { IVault } from "@balancer-labs/v3-interfaces/contracts/vault/IVault.sol";
 
-import { SingletonAuthentication } from "@balancer-labs/v3-vault/contracts/SingletonAuthentication.sol";
+import { PoolHelperCommon } from "./PoolHelperCommon.sol";
 
-contract PoolPauseHelper is IPoolPauseHelper, SingletonAuthentication {
+contract PoolPauseHelper is IPoolPauseHelper, PoolHelperCommon {
     using EnumerableSet for EnumerableSet.AddressSet;
 
-    EnumerableSet.AddressSet private _pausablePools;
-
-    constructor(IVault vault) SingletonAuthentication(vault) {
+    constructor(IVault vault) PoolHelperCommon(vault) {
         // solhint-disable-previous-line no-empty-blocks
     }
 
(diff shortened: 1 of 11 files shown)


pkg/standalone-utils/contracts/PoolPauseHelper.sol
```diff
@@ -8,85 +8,29 @@ import { IPoolPauseHelper } from "@balancer-labs/v3-interfaces/contracts/standal
...
     /// @inheritdoc IPoolPauseHelper
     function pausePools(address[] memory pools) public authenticate {
         uint256 length = pools.length;
         for (uint256 i = 0; i < length; i++) {
             address pool = pools[i];
-            if (_pausablePools.contains(pool) == false) {
-                revert PoolNotInPausableSet(pool);
+            if (_pools.contains(pool) == false) {
+                revert PoolNotInSet(pool);
             }
```
jubeira, line 31:
```suggestion
            _ensurePoolAdded(pool);
```

pkg/standalone-utils/contracts/PoolSwapFeeHelper.sol
```diff
@@ -5,102 +5,38 @@ pragma solidity ^0.8.24;
...
-            address pool = pools[i];
-            _ensureKnownPool(pool);
-            _pools.remove(pool);
-
-            emit PoolRemovedFromSwapFeeSet(pool);
-        }
-    }
-
     /// @inheritdoc IPoolSwapFeeHelper
     function setStaticSwapFeePercentage(address pool, uint256 swapFeePercentage) public authenticate {
```
jubeira, line 25: (for another PR): should we have a setter for multiple pools / multiple fee percentages as well? (parallel arrays)

pkg/standalone-utils/contracts/ProtocolFeeHelper.sol
```diff
@@ -9,60 +9,24 @@ import { IProtocolFeeController } from "@balancer-labs/v3-interfaces/contracts/v
...
-import { SingletonAuthentication } from "@balancer-labs/v3-vault/contracts/SingletonAuthentication.sol";
+import { PoolHelperCommon } from "./PoolHelperCommon.sol";
 
-contract ProtocolFeeHelper is IProtocolFeeHelper, SingletonAuthentication {
+contract ProtocolFeeHelper is IProtocolFeeHelper, PoolHelperCommon {
     using EnumerableSet for EnumerableSet.AddressSet;
 
-    EnumerableSet.AddressSet private _pools;
-
     modifier withKnownPool(address pool) {
```
jubeira, line 17:
```suggestion
    modifier withPoolAdded(address pool) {
```

pkg/standalone-utils/test/foundry/PoolPauseHelper.t.sol
```diff
@@ -67,11 +68,10 @@ contract PoolPauseHelperTest is BaseVaultTest {
     function testDoubleAddOnePool() public {
         assertEq(pauseHelper.getPoolCount(), 0, "Initial pool count non-zero");
 
-        address[] memory pools = new address[](2);
-        pools[0] = address(0x1);
-        pools[1] = address(0x1);
+        address[] memory pools = _generatePools(2);
+        pools[1] = pools[0];
```
jubeira, line 72: these would be `address(0)`; is that ok? maybe set `pools[0]` to `0x1`?

jubeira (approved):
LGTM! Looks much better like this.

Just a few minor comments, otherwise it's good to :ship:

pkg/standalone-utils/contracts/PoolSwapFeeHelper.sol
EndymionJkb: Sounds like a question to ask the people who want to use this. Would it be more common to set a bunch of pools to the same fee, so only the first argument would be an array? Is there a use case for setting a bunch of pools to random different values (maybe some kind of Gauntlet-like fee setter)? If those pools are all from the same factory, you would probably use the factory-based fee setter.

Open to all opens, but we should probably verify there's a use case before we just add stuff.

pkg/standalone-utils/contracts/ProtocolFeeHelper.sol
EndymionJkb: I don't know... maybe this one's ok as is. "withPoolAdded" sounds weird, even though it would make the name more consistent. Maybe withValidPool? "Add" ensures it is registered - plus passes any additional validations specific to that helper - so this means it's been validated and added to the set before use.

pkg/standalone-utils/contracts/PoolSwapFeeHelper.sol
jubeira: the use case would be pools that are managed by a third party via the helper, but somewhat manually (as opposed to ezkl where we're setting the fee pool by pool).

EndymionJkb: Should we have a setter with an array of pools and a single value, and also a setter with parallel arrays?

pkg/standalone-utils/contracts/ProtocolFeeHelper.sol
jubeira: I don't have a strong opinion; it's just that I got used to naming things 'consistently' by sticking to the same word even if it sounds dumb as opposed to calling the same thing by two names.

`addPool` adds a pool, so the condition to check is whether the pool was added, not registered / valid --> `withPoolAdded` (or `withAddedPool`?). It doesn't sound that nice, but it sticks to the same word and suggests that you first need to call `addPool` to call the function affected by the modifier in a very direct manner.

I'm also fine with this option though; the contract is not complex enough to cause that much confusion around this/

PR #1365, opened 2025-04-25
Add ERC4626 to ProtocolFeeSweeper

### LINKED ISSUE #1355: Consider adding `unwrap` functionality to cow burner
Unwrap before burn on demand: caller would specify a flag, and the burner would treat the token as an ERC4626, unwrap, and then proceed with the burn with the underlying token.

elshan-eth:
# Description

This PR adds an unwrap functionality for ERC4626 tokens. The unwrap happens before interacting with the fee burner, so a boolean flag was added to sweepProtocolFeesForToken to let the admin choose whether to burn the ERC4626 token or unwrap it first. The unwrap is done directly via the token to save gas.

It would’ve been possible to avoid modifying sweepProtocolFeesForToken, but in this case I think that unwrapping the token is a responsibility of the ProtocolFeeSweeper, since the Burner is only responsible for burning tokens.

## Type of change

- [ ] Bug fix 
- [ ] New feature 
- [ ] Breaking change 
- [ ] Dependency changes
- [ ] Code refactor / cleanup
- [ ] Other

## Checklist:

## Issue Resolution

Closes #1355

### DIFF
--- a/pkg/interfaces/contracts/standalone-utils/IProtocolFeeSweeper.sol
+++ b/pkg/interfaces/contracts/standalone-utils/IProtocolFeeSweeper.sol
@@ -95,13 +95,15 @@ interface IProtocolFeeSweeper {
      * @param minTargetTokenAmountOut The minimum number of target tokens to be received
      * @param deadline Deadline for the burn operation (swap), after which it will revert
      * @param feeBurner The protocol fee burner to be used (or the zero address to fall back on direct transfer)
+     * @param shouldUnwrap Whether to unwrap the target token before burning
      */
     function sweepProtocolFeesForToken(
         address pool,
         IERC20 feeToken,
         uint256 minTargetTokenAmountOut,
         uint256 deadline,
-        IProtocolFeeBurner feeBurner
+        IProtocolFeeBurner feeBurner,
+        bool shouldUnwrap
     ) external;
 
     /**
--- a/pkg/standalone-utils/contracts/ProtocolFeeSweeper.sol
+++ b/pkg/standalone-utils/contracts/ProtocolFeeSweeper.sol
@@ -3,6 +3,7 @@
 pragma solidity ^0.8.24;
 
 import { SafeERC20 } from "@openzeppelin/contracts/token/ERC20/utils/SafeERC20.sol";
+import { IERC4626 } from "@openzeppelin/contracts/interfaces/IERC4626.sol";
 import { IERC20 } from "@openzeppelin/contracts/token/ERC20/IERC20.sol";
 
 import { IProtocolFeeSweeper } from "@balancer-labs/v3-interfaces/contracts/standalone-utils/IProtocolFeeSweeper.sol";
@@ -58,7 +59,8 @@ contract ProtocolFeeSweeper is IProtocolFeeSweeper, SingletonAuthentication, Ree
         IERC20 feeToken,
         uint256 minTargetTokenAmountOut,
         uint256 deadline,
-        IProtocolFeeBurner feeBurner
+        IProtocolFeeBurner feeBurner,
+        bool shouldUnwrap
     ) external nonReentrant onlyFeeRecipientOrGovernance {
         bool feeBurnerProvided = _getValidFeeBurner(feeBurner);
 
@@ -85,6 +87,13 @@ contract ProtocolFeeSweeper is IProtocolFeeSweeper, SingletonAuthentication, Ree
                 if (feeToken == targetToken) {
                     _transferFeeToken(pool, feeToken, withdrawnBalance);
                 } else {
+                    if (shouldUnwrap) {
+                        IERC4626 erc4626Token = IERC4626(address(feeToken));
+
+                        feeToken = IERC20(erc4626Token.asset());
+                        withdrawnBalance = erc4626Token.redeem(withdrawnBalance, address(this), address(this));
+                    }
+
                     // We must revert if this allowance is not entirely consumed by the burner, to avoid exploitable
                     // "hanging approvals." The order may be asynchronous, but it must pull the tokens immediately.
                     feeToken.forceApprove(address(feeBurner), withdrawnBalance);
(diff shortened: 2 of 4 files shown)


EndymionJkb (commented):
I agree it's better to handle this at the ProtocolFeeSweeper level, but wonder if changing this interface will be too disruptive.

How common is this case? Could we just add a new `sweepProtocolFeesForWrappedToken` function that takes an IERC4626 instead of an IERC20 as the feeToken, then unwraps and calls the regular one (which we could make public)?

pkg/interfaces/contracts/standalone-utils/IProtocolFeeSweeper.sol
```diff
@@ -104,6 +104,23 @@ interface IProtocolFeeSweeper {
         IProtocolFeeBurner feeBurner
     ) external;
 
+    /**
+     * @notice The same as `sweepProtocolFeesForToken`, but for wrapped tokens.
+     *
```
EndymionJkb, line 109:
```suggestion
```
There's only a gap for multi-line notice/dev blocks.

EndymionJkb (approved):
I can think of two ways to do this, and don't have a really strong opinion. This way is fine.

Another way would be to make the standard function public, make the `sweepProtocolFeesForWrappedToken` function take an `IERC4626 wrappedFeeToken` explicitly, and in its implementation get the IERC20 feeToken and unwrap, then call the standard function directly with the underlying token.

Then we don't need to change the standard function at all (well, other than the visibility), and the signature of the new one matches what you're doing. I think I would slightly prefer that, but up to you.

elshan-eth:
@EndymionJkb, I didn’t fully understand how this would work, because if we unwrap the token before the `sweepProtocolFeesForToken`, we won’t pass through this part of the code:
```
bool feeBurnerProvided = _getValidFeeBurner(feeBurner);

uint256 existingBalance = feeToken.balanceOf(address(this));

// Withdraw protocol fees to this contract. Note that governance will need to grant this contract permission
// to call `withdrawProtocolFeesForToken` on the `ProtocolFeeController.
IProtocolFeeController feeController = getProtocolFeeController();

// Transfer any pending fees from the Vault to the ProtocolFeeController, and allocate the protocol and pool
// creator portions.
feeController.collectAggregateFees(pool);

// Withdraw the protocol portion of the fees to this contract (requires permission).
feeController.withdrawProtocolFeesForToken(pool, address(this), feeToken);

uint256 withdrawnBalance = feeToken.balanceOf(address(this)) - existingBalance;
```

pkg/standalone-utils/contracts/ProtocolFeeSweeper.sol
```diff
@@ -85,6 +108,13 @@ contract ProtocolFeeSweeper is IProtocolFeeSweeper, SingletonAuthentication, Ree
                 if (feeToken == targetToken) {
                     _transferFeeToken(pool, feeToken, withdrawnBalance);
```
jubeira, line 109: maybe require `shouldUnwrap == false` in this case?

It's an edge case, but `shouldUnwrap` doesn't make sense in this context, and it can happen in practice.

pkg/standalone-utils/test/foundry/ProtocolFeeSweeper.t.sol
```diff
@@ -263,6 +292,35 @@ contract ProtocolFeeSweeperTest is BaseVaultTest {
...
+        vm.prank(admin);
+        feeSweeper.sweepProtocolFeesForWrappedToken(waDAIPool, waDAI, 0, MAX_UINT256, feeBurner);
+
+        assertEq(waDAI.balanceOf(address(feeController)), 0, "waDAI not withdrawn");
+        assertEq(dai.balanceOf(address(feeController)), 0, "DAI not withdrawn");
+        assertEq(usdc.balanceOf(address(feeController)), 0, "USDC not withdrawn");
+        assertEq(waDAI.balanceOf(address(feeSweeper)), 0, "Final sweeper waDAI balance non-zero");
+        assertEq(dai.balanceOf(address(feeSweeper)), 0, "Final sweeper DAI balance non-zero");
+        assertEq(usdc.balanceOf(address(feeSweeper)), 0, "Final sweeper USDC balance non-zero");
+        // DAI should have been converted to USDC, so we should have twice the DEFAULT_AMOUNT of it.
```
jubeira, line 318: Maybe check this?

```
assertEq(finalAmount, DEFAULT_AMOUNT * 2, "Incorrect USDC amount forwarded");
```

jubeira (approved):
LGTM, very clean! Nice job @elshan-eth

pkg/standalone-utils/contracts/ProtocolFeeSweeper.sol
```diff
@@ -60,6 +61,28 @@ contract ProtocolFeeSweeper is IProtocolFeeSweeper, SingletonAuthentication, Ree
...
+    }
+
+    /// @inheritdoc IProtocolFeeSweeper
+    function sweepProtocolFeesForWrappedToken(
+        address pool,
+        IERC20 feeToken,
+        uint256 minTargetTokenAmountOut,
+        uint256 deadline,
+        IProtocolFeeBurner feeBurner
+    ) external nonReentrant onlyFeeRecipientOrGovernance {
```
jubeira, line 74: Late comment: let's test `onlyFeeRecipientOrGovernance` modifier too

pkg/standalone-utils/contracts/ProtocolFeeSweeper.sol
elshan-eth: I think we set the `targetToken` ourselves, and if we want to prevent its use with ERC4626 in this case, we should add a check for `targetToken` inside `setTargetToken`. WDYT?

EndymionJkb: It doesn't really hurt anything though, does it? Worst case you'd just have to unwrap it manually at the destination. There could be a case where you just want the wrapped token directly, so I don't think we should categorically prohibit ERC4626 as targets.

pkg/standalone-utils/contracts/ProtocolFeeSweeper.sol
```diff
@@ -60,6 +61,28 @@ contract ProtocolFeeSweeper is IProtocolFeeSweeper, SingletonAuthentication, Ree
         uint256 deadline,
         IProtocolFeeBurner feeBurner
     ) external nonReentrant onlyFeeRecipientOrGovernance {
+        _sweepProtocolFeesForToken(pool, feeToken, minTargetTokenAmountOut, deadline, feeBurner, false);
+    }
+
+    /// @inheritdoc IProtocolFeeSweeper
+    function sweepProtocolFeesForWrappedToken(
+        address pool,
+        IERC20 feeToken,
```
EndymionJkb, line 70:
```suggestion
        IERC4626 wrappedFeeToken,
```
Setting the type explicitly here would be self-documenting, and doesn't change anything else (you can still pass an IERC4626 to an IERC20 argument)

PR #1371, opened 2025-05-02
Add ERC4626 cow burner

jubeira:
# Description

An alternative approach to #1365.
This is not my personal preferred approach from an architecture's standpoint, but it's easier to wire up governance-wise and we might be able to use it in the meantime.

Instead of unwrapping at the sweeper, we can deploy a new cow burner that unwraps before posting the order.
It has a few caveats: it must pull the tokens from the sweeper before creating the order and handle a few edge cases. But other than that it's rather reasonable.

Tests pending.

## Type of change

- [ ] Bug fix 
- [x] New feature 
- [ ] Breaking change 
- [ ] Dependency changes
- [ ] Code refactor / cleanup
- [ ] Other

## Checklist:

## Issue Resolution

N/A

### DIFF
--- a/pkg/standalone-utils/contracts/ERC4626CowSwapFeeBurner.sol
+++ b/pkg/standalone-utils/contracts/ERC4626CowSwapFeeBurner.sol
@@ -53,13 +53,13 @@ contract ERC4626CowSwapFeeBurner is CowSwapFeeBurner {
 
     /**
      * @notice Treats `feeToken` as an ERC4626, redeems `exactFeeTokenAmountIn`, swaps the underlying asset for
-     * `targetToken`, and send proceeds to the `recipient`.
+     * `targetToken`, and sends the proceeds to the `recipient`.
      * @dev Assumes the sweeper has granted allowance for the fee tokens to the burner prior to the call.
      * @param pool The pool the fees came from (only used for documentation in the event)
      * @param feeToken The token collected from the pool
      * @param exactFeeTokenAmountIn The number of fee tokens collected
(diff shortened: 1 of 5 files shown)


pkg/standalone-utils/contracts/ERC4626CowSwapFeeBurner.sol
```diff
@@ -0,0 +1,107 @@
...
+        // This case is not handled by the internal `_burn` function, but it's valid: we can consider that the token
+        // has already been converted to the correct token, so we just forward the result and finish.
+        if (feeToken == targetToken) {
+            // We apply the slippage check, but not deadline as the order settlement is instant in this case.
+            if (exactFeeTokenAmountIn < minTargetTokenAmountOut) {
+                revert AmountOutBelowMin(targetToken, exactFeeTokenAmountIn, minTargetTokenAmountOut);
+            }
+
+            feeToken.safeTransfer(recipient, exactFeeTokenAmountIn);
+            return;
```
EndymionJkb, line 93: Consider if/else vs. early return, since there's only one line after it.

EndymionJkb (approved):
LGTM! The fee sweeper already does the approvals, so it looks like it should work (and with the deployed one). Just a couple comment nits.

pkg/standalone-utils/contracts/ERC4626CowSwapFeeBurner.sol
jubeira: that's fair

PR #1372, opened 2025-05-05
Adjust e2e swap tests to allow fuzz params to edit limits

jubeira:
# Description

For e2e swap tests, in some cases `fuzzSwapParams` might edit balances or parameters that affect swap limits.
`testDoUndoExactInBase` and `testDoUndoExactOutBase` currently force the balances and the maximum amounts after fuzzing pool params, which is not ideal. 

For Reclamm in particular, virtual balances (i.e. pool state) are tied to pool balances, so if the test forces pool balances without considering virtual balances, something will go out of sync. This change allows keeping pool state aligned with fuzzed pool balances by changing the order of execution and allowing overrides.

## Type of change

- [x] Bug fix 
- [ ] New feature 
- [ ] Breaking change 
- [ ] Dependency changes
- [ ] Code refactor / cleanup
- [ ] Other

## Checklist:

## Issue Resolution

N/A

### DIFF
--- a/pkg/pool-gyro/test/foundry/E2eSwap2CLP.t.sol
+++ b/pkg/pool-gyro/test/foundry/E2eSwap2CLP.t.sol
@@ -87,7 +87,13 @@ contract E2eSwapGyro2CLPTest is E2eSwapTest, Gyro2ClpPoolDeployer {
         maxSwapAmountTokenB = poolInitAmountTokenB.mulDown(50e16);
     }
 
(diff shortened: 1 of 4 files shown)


elshan-eth (commented):
Looks good. I think we could make it more structured and readable if we make `fuzzParams` and `setBalance` two separate override methods, but we can think about it on the optimization stage.

I continue to check this PR with e2e reclamm.

joaobrunoah (approved):
To me, it looks good for now. I agree that we can optimize it even further, but looks like it does the job now. I've tried the changes in the e2e branch of ReClamm and it worked, so looks good.

PR #1387, opened 2025-05-22
Add minAmountOut for unwrapping in ERC4626CowBurner

### LINKED ISSUE #1386: Protect erc4626 router against low value redeems
- Encode redeem min amount out in `minAmountOut` parameter
- Apply the encoded amount to redeem (or > 0 if nothing was encoded)
- Use the lower half of the word for the min amount out for the cow order (i.e. current `minAmountOut`)

elshan-eth:
# Description

This PR adds support for using minAmountOut when unwrapping ERC4626 tokens in ERC4626CowBurner. The issue is that ERC4626 can have different implementations and may sometimes don't return tokens. For example, we encountered this problem with rstETH. As a temporary solution to this issue, we decided to use a single variable for two values in order to avoid changing the interface.

## Type of change

- [ ] Bug fix 
- [X] New feature 
- [ ] Breaking change 
- [ ] Dependency changes
- [ ] Code refactor / cleanup
- [ ] Other

## Checklist:

## Issue Resolution

Closes #1386

### DIFF
--- a/pkg/standalone-utils/contracts/ERC4626CowSwapFeeBurner.sol
+++ b/pkg/standalone-utils/contracts/ERC4626CowSwapFeeBurner.sol
@@ -13,6 +13,7 @@ import { IVault } from "@balancer-labs/v3-interfaces/contracts/vault/IVault.sol"
 import {
     ReentrancyGuardTransient
 } from "@balancer-labs/v3-solidity-utils/contracts/openzeppelin/ReentrancyGuardTransient.sol";
+import { PackedTokenBalance } from "@balancer-labs/v3-solidity-utils/contracts/helpers/PackedTokenBalance.sol";
 
 import { CowSwapFeeBurner } from "./CowSwapFeeBurner.sol";
 
@@ -63,8 +64,9 @@ contract ERC4626CowSwapFeeBurner is CowSwapFeeBurner {
         // the underlying token.
         feeToken.safeTransferFrom(msg.sender, address(this), exactFeeTokenAmountIn);
 
-        uint256 minTargetTokenAmountOut = uint128(encodedMinAmountsOut >> 128);
-        uint256 minERC4626AmountOut = uint128(encodedMinAmountsOut & type(uint128).max);
+        (uint256 minTargetTokenAmountOut, uint256 minERC4626AmountOut) = PackedTokenBalance.fromPackedBalance(
+            bytes32(encodedMinAmountsOut)
+        );
 
         IERC4626 erc4626Token = IERC4626(address(feeToken));
         // Redeem and overwrite inputs with new asset and unwrapped amount.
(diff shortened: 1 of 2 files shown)


pkg/standalone-utils/contracts/ERC4626CowSwapFeeBurner.sol
```diff
@@ -52,19 +55,34 @@ contract ERC4626CowSwapFeeBurner is CowSwapFeeBurner {
...
         address recipient,
         uint256 deadline
     ) external override onlyProtocolFeeSweeper nonReentrant {
         // In this case we first pull the wrapped token, unwrap, and then proceed to burn by creating an order for
         // the underlying token.
         feeToken.safeTransferFrom(msg.sender, address(this), exactFeeTokenAmountIn);
 
-        // Redeem and overwrite inputs with new asset and unwrapped amount.
+        uint256 minTargetTokenAmountOut = uint128(encodedMinAmountsOut >> 128);
+        uint256 minERC4626AmountOut = uint128(encodedMinAmountsOut & type(uint128).max);
```
joaobrunoah, line 67: Not sure if it's worth it, but a function to decode the minAmountsOut in target and ERC4626 would make this part of the code clearer.

joaobrunoah (approved):
LGTM, only a minor comment.

pkg/standalone-utils/contracts/ERC4626CowSwapFeeBurner.sol
```diff
@@ -52,19 +55,34 @@ contract ERC4626CowSwapFeeBurner is CowSwapFeeBurner {
...
         exactFeeTokenAmountIn = erc4626Token.redeem(exactFeeTokenAmountIn, address(this), address(this));
 
+        uint256 feeTokenBalanceAfter = feeToken.balanceOf(address(this));
+        uint256 redeemAmount = feeTokenBalanceAfter - feeTokenBalanceBefore;
+
+        if (exactFeeTokenAmountIn < minERC4626AmountOut) {
+            revert AmountOutBelowMin(feeToken, exactFeeTokenAmountIn, minERC4626AmountOut);
+        } else if (redeemAmount < minERC4626AmountOut) {
+            revert AmountOutBelowMin(feeToken, redeemAmount, minERC4626AmountOut);
+        }
```
EndymionJkb, line 84:
```suggestion
        if (redeemAmount < minERC4626AmountOut) {
            revert AmountOutBelowMin(feeToken, redeemAmount, minERC4626AmountOut);
        }
```
It seems like this top one is checked twice? It's a little hard to follow (could use more comments). The tests pass when this is removed, so either it's really redundant, or we're missing a test.

EndymionJkb (commented):
A couple comments and suggestions

EndymionJkb (approved):
LGTM! (Pending CI.)

pkg/standalone-utils/contracts/ERC4626CowSwapFeeBurner.sol
```diff
@@ -52,32 +55,45 @@ contract ERC4626CowSwapFeeBurner is CowSwapFeeBurner {
...
+        uint256 feeTokenBalanceBefore = underlyingToken.balanceOf(address(this));
+
+        erc4626Token.redeem(exactFeeTokenAmountIn, address(this), address(this));
+
+        uint256 feeTokenBalanceAfter = underlyingToken.balanceOf(address(this));
+        exactFeeTokenAmountIn = feeTokenBalanceAfter - feeTokenBalanceBefore;
+
+        if (exactFeeTokenAmountIn < minERC4626AmountOut) {
+            revert AmountOutBelowMin(underlyingToken, exactFeeTokenAmountIn, minERC4626AmountOut);
+        }
```
jubeira, line 82: Two comments here:
- ERC4626 should return the assets redeemed. It is true that measuring the balances should be accurate, but unless the ERC4626 is non-compliant it shouldn't be strictly necessary. I'm fine with leaving it either way.
- If `minERC4626AmountOut` is 0 and `exactFeeTokenAmountIn` is 0 this should revert as well. The code as is wouldn't prevent the issue we've seen yesterday unless the `minERC4626AmountOut` is encoded, which should be optional

PR #1390, opened 2025-05-27
LBP Migration Router

elshan-eth:
# Description

This PR adds a router for migrating an LBP pool to a Weighted Pool. The owner of the LBP can initiate the migration, during which a new Weighted Pool with custom parameters will be created. The user must specify the amount to be deposited into the new pool, the remaining tokens will be sent to the treasury.

## Type of change

- [ ] Bug fix 
- [ ] New feature 
- [ ] Breaking change 
- [ ] Dependency changes
- [ ] Code refactor / cleanup
- [ ] Other

## Checklist:

### DIFF
--- a/pkg/interfaces/contracts/vault/ILBPMigrationRouter.sol
+++ b/pkg/interfaces/contracts/vault/ILBPMigrationRouter.sol
@@ -1,0 +1,88 @@
+// SPDX-License-Identifier: GPL-3.0-or-later
+
+pragma solidity ^0.8.24;
+
+import { IERC20 } from "@openzeppelin/contracts/token/ERC20/IERC20.sol";
+
+import { PoolRoleAccounts } from "@balancer-labs/v3-interfaces/contracts/vault/VaultTypes.sol";
+import { IWeightedPool } from "@balancer-labs/v3-interfaces/contracts/pool-weighted/IWeightedPool.sol";
+
+import { ILBPool } from "../pool-weighted/ILBPool.sol";
+
+/// @notice Interface for migrating liquidity from a Liquidity Bootstrapping Pool (LBP) to new Weighted Pool with custom parameters.
+interface ILBPMigrationRouter {
+    /**
+     * @notice Thrown when trying to migrate liquidity, but the LBP weights are not yet finalized.
+     * @param lbp The Liquidity Bootstrapping Pool with unfinalized weights
+     */
+    error LBPWeightsNotFinalized(ILBPool lbp);
+
+    /**
+     * @notice Thrown when the actual input amount of a token is less than required.
+     * @param token The token with insufficient input amount
+     * @param actualAmount The actual amount of the token provided
+     */
+    error InsufficientInputAmount(IERC20 token, uint256 actualAmount);
+
+    /**
+     * @notice Thrown when the caller is not the owner of the LBP.
+     * @param lbpOwner The actual owner of the LBP
+     */
+    error SenderIsNotLBPOwner(address lbpOwner);
+
+    struct MigrationHookParams {
+        ILBPool lbp;
+        IWeightedPool weightedPool;
+        IERC20[] tokens;
+        address sender;
+        uint256[] exactAmountsIn;
+        uint256 minAddBptAmountOut;
+        uint256[] minRemoveAmountsOut;
+    }
+
+    struct WeightedPoolParams {
+        string name;
+        string symbol;
+        uint256[] normalizedWeights;
+        PoolRoleAccounts roleAccounts;
+        uint256 swapFeePercentage;
+        address poolHooksContract;
+        bool enableDonation;
+        bool disableUnbalancedLiquidity;
+        bytes32 salt;
+    }
+
+    /**
+     * @notice Migrates liquidity from an LBP to a new weighted pool with custom parameters.
+     * @param lbp Liquidity Bootstrapping Pool
+     * @param exactAmountsIn The exact amounts of each token to add to the weighted pool
+     * @param minAddBptAmountOut Minimum amount of BPT tokens expected to receive
+     * @param minRemoveAmountsOut Minimum token amounts expected when removing from the LBP
+     * @param params Parameters for creating the new weighted pool
+     * @return weightedPool The newly created weighted pool
+     * @return bptAmountOut The amount of BPT tokens received from the weighted pool after migration
+     */
+    function migrateLiquidity(
+        ILBPool lbp,
+        uint256[] memory exactAmountsIn,
+        uint256 minAddBptAmountOut,
+        uint256[] memory minRemoveAmountsOut,
+        WeightedPoolParams memory params
+    ) external returns (IWeightedPool weightedPool, uint256 bptAmountOut);
+
+    /**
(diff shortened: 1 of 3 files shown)


pkg/pool-weighted/contracts/LBPMigrationRouter.sol
```diff
@@ -0,0 +1,172 @@
...
+        {
+            LBPoolImmutableData memory lbpImmutableData = lbp.getLBPoolImmutableData();
+
+            // solhint-disable-next-line not-rely-on-time
+            if (block.timestamp <= lbpImmutableData.endTime) {
+                revert LBPWeightsNotFinalized(lbp);
+            }
+        }
+
+        IERC20[] memory tokens = _vault.getPoolTokens(address(lbp));
```
EndymionJkb, line 124:
```suggestion
```
Can use lbpImmutableData.tokens (without scope).

pkg/pool-weighted/contracts/LBPMigrationRouter.sol
```diff
@@ -0,0 +1,172 @@
...
+
+    /// @inheritdoc ILBPMigrationRouter
+    function queryMigrateLiquidity(
+        ILBPool lbp,
+        uint256[] memory exactAmountsIn,
+        address sender,
+        WeightedPoolParams memory params
+    ) external returns (IWeightedPool, uint256) {
+        return _migrateLiquidity(lbp, exactAmountsIn, 0, new uint256[](exactAmountsIn.length), sender, params, true);
+    }
```
EndymionJkb, line 70:
```suggestion
    function queryMigrateLiquidity(
        ILBPool lbp,
        uint256[] memory exactAmountsIn,
        address sender,
        WeightedPoolParams memory params
    ) external returns (uint256 bptAmountOut) {
        (, bptAmountOut) = _migrateLiquidity(
            lbp,
            exactAmountsIn,
            0,
            new uint256[](exactAmountsIn.length),
            sender,
            params,
            true
        );
    }
```
Seems strange to return an ephemeral weighted pool address here. Maybe do this?

EndymionJkb (commented):
First pass. Was this "remainder to the treasury" feature requested? Is it intended to remove dust through donation, or as a fee mechanism (e.g., they would agree to withdraw 90%)?

Since the withdrawal is proportional, should we allow the new weighted pool to have arbitrary weights? Maybe we should always use the LBP ending weights, or it seems there would be a mismatch if they were very different.

Does this even make sense if the sale ends at 20/80 (with 20% their token) or 10/90? Wouldn't they want their token to be the higher weight in the regular weighted pool? Or is this intended to be used with pools that end at or close to 50/50?

I think we need more documentation on the intended use case here.

pkg/pool-weighted/contracts/lbp/LBPMigrationRouter.sol
```diff
@@ -0,0 +1,193 @@
...
+
+        return
+            _vault.initialize(
+                address(params.weightedPool),
+                params.sender,
+                params.tokens,
+                params.exactAmountsIn,
+                params.minAddBptAmountOut,
+                bytes("")
+            );
```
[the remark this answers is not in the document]
EndymionJkb: reCLAMM ideas are migrating to LBPs :) There were in fact some "accidents" with LBPs, most notably one where the creator got the weights backwards, and the pool was immediately arbed for a big loss.

pkg/interfaces/contracts/pool-weighted/ILBPMigrationRouter.sol
```diff
@@ -0,0 +1,104 @@
...
+     * @param excessReceiver Address to receive excess tokens after migration
+     * @param params Parameters for creating the new weighted pool
+     * @return weightedPool The newly created weighted pool
+     * @return bptAmountOut The amount of BPT tokens received from the weighted pool after migration
+     */
+    function migrateLiquidity(
+        ILBPool lbp,
+        uint256[] memory exactAmountsIn,
+        uint256 minAddBptAmountOut,
+        uint256[] memory minRemoveAmountsOut,
```
jubeira, line 83: As discussed off-thread, the migration parameter will be a maximum liquidity percentage.

About the slippage, it shouldn't really be necessary, and this is going to be called by another contract in the end. So maybe a single FP-18 number representing the maximum BPT percentage to migrate should do.

The router will exit all of the liquidity, and initialize the new pool with a percentage of it keeping the same price. The rest goes to the `excessReceiver` address.

pkg/pool-weighted/contracts/lbp/LBPMigrationRouter.sol
```diff
@@ -0,0 +1,199 @@
...
+                userData: new bytes(0)
+            })
+        );
+
+        for (uint256 i = 0; i < removeAmountsOut.length; i++) {
+            if (params.exactAmountsIn[i] > removeAmountsOut[i]) {
+                revert InsufficientInputAmount(params.tokens[i], removeAmountsOut[i]);
+            }
+
+            uint256 remainingBalance = removeAmountsOut[i] - params.exactAmountsIn[i];
```
jubeira, line 120: Here we'll need to apply the percentage considering the new pool's weight, and computing the input amounts accordingly to preserve the price considering the difference between the LBP end weights and weighted pool's starting weights.

pkg/pool-weighted/test/foundry/LBPMigrationRouter.t.sol
```diff
@@ -0,0 +1,409 @@
...
+                    "Rate provider address mismatch"
+                );
+                assertEq(tokenInfo[i].paysYieldFees, lbpTokenInfo[i].paysYieldFees, "Pays yield fees mismatch");
+            }
+
+            PoolConfig memory poolConfig = vault.getPoolConfig(address(weightedPool));
+            assertEq(poolConfig.staticSwapFeePercentage, DEFAULT_SWAP_FEE_PERCENTAGE, "Incorrect swap fee percentage");
+            assertEq(
+                poolConfig.liquidityManagement.disableUnbalancedLiquidity,
+                true,
```
jubeira, line 151: This one should actually be false. It's just a regular weighted pool now, so there's no reason to disable unbalanced liquidity.

pkg/pool-weighted/test/foundry/LBPMigrationRouter.t.sol
```diff
@@ -0,0 +1,409 @@
...
+                "Disable unbalanced liquidity should be true"
+            );
+            assertEq(poolConfig.liquidityManagement.enableDonation, true, "Enable donation should be true");
```
jubeira, line 154: And this one should probably be false.

jubeira (commented):
Great work!

I've done a pass over the tests and it's very reasonable. Just a few minor comments (mostly config flags for enable donation and disable unbalanced liquidity for the new pool). I think we should only address those two things; changing the migration parameters from exact inputs to a liquidity percentage is out of scope of this PR, as that approach was discussed after the first version was out.

Great job!

jubeira (approved):
LGTM, great work!

I have just a few comments to address in this PR; the rest can be iterated upon later. 
Tests also look pretty good! Thanks for pushing this!

PR #1391, opened 2025-05-28
Change auth logic in cow burner

elshan-eth:
# Description

This PR changes the authorization system in the Cow Fee Burner. One of the options was to leave the retry and cancel methods without authorization, but the downside of this approach is that it could allow griefing attacks in very rare cases. For example, we might want to call retry, but someone could spam cancel. This is an extremely unlikely scenario, but I’d prefer to avoid it. 

## Type of change

- [ ] Bug fix 
- [X] New feature 
- [ ] Breaking change 
- [ ] Dependency changes
- [ ] Code refactor / cleanup
- [ ] Other

## Checklist:


jubeira (approved):
LGTM, I think this is pretty reasonable. Thanks for pushing it!

PR #1379, opened 2025-05-20
[DRAFT] add lp oracle for weighted pool

elshan-eth:
# Description

## Type of change

- [ ] Bug fix 
- [ ] New feature 
- [ ] Breaking change 
- [ ] Dependency changes
- [ ] Code refactor / cleanup
- [ ] Optimization: [ ] gas / [ ] bytecode
- [ ] Documentation or wording changes
- [ ] Other

## Checklist:

- [ ] The diff is legible and has no extraneous changes
- [ ] Complex code has been commented, including external interfaces
- [ ] Tests have 100% code coverage
- [ ] The base branch is either `main`, or there's a description of how to merge

PR #1395, opened 2025-06-04
LP Oracle for Stable Pools

joaobrunoah:
# Description

This PR creates an LP oracle for Stable pools, it means, a way to safely price the BPT of stable pools, so it can be used as a collateral.

The method is based on https://www.notion.so/Pricing-stable-pool-tokens-202e395e2281807f9021fdcafe376aa7?pvs=5 and is built on top of the PR #1379 .

## Type of change

- [ ] Bug fix 
- [x] New feature 
- [ ] Breaking change 
- [ ] Dependency changes
- [ ] Code refactor / cleanup
- [ ] Other

## Checklist:


pkg/standalone-utils/contracts/LPOracleBase.sol
```diff
@@ -0,0 +1,255 @@
...
+        uint256 tvl = calculateTVL(prices);
+        uint256 totalSupply = _vault.totalSupply(address(pool));
+
+        uint256 lpPrice = tvl.divUp(totalSupply);
+
+        return (0, lpPrice.toInt256(), 0, _updatedAt, 0);
+    }
+
+    /// @inheritdoc ILPOracleBase
+    function calculateTVL(int256[] memory) public view virtual returns (uint256) {
```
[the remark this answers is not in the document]
joaobrunoah: I've tried, but then I'd need to define `latestRoundData()` function in every child oracle, and they'd basically be the same.

pkg/standalone-utils/test/foundry/StableLPOracle.t.sol
```diff
@@ -0,0 +1,424 @@
...
+            uint256[] memory poolInitAmounts = new uint256[](totalTokens);
+            address[] memory _tokens = new address[](totalTokens);
+
+            for (uint256 i = 0; i < totalTokens; i++) {
+                _tokens[i] = address(sortedTokens[i]);
+                uint256 decimalsToken = IERC20Metadata(address(sortedTokens[i])).decimals();
+                // poolInitAmounts will only define the invariant of the pool. Since K does not depend on the
+                // invariant, we don't need to fuzz it. Besides, fuzzing it causes the Stable Invariant to don't
+                // converge during initialization, which introduces unnecessary complexities in the test.
+                poolInitAmounts[i] = defaultAccountBalance() / 100 / (10 ** (18 - decimalsToken));
```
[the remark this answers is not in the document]
joaobrunoah: The balances are only used to calculate the invariant, and what unbalances the pool are the token prices. If I randomly pick a starting point, the Stable Invariant does not converge.  What I can do is randomize the invariant itself, and initialize the pool with "invariant/n" amount on each token.

This would just be an artifact to don't deal with Stable invariant not converging during the initialization, since that's not the point of the test and the balances are only used to compute the invariant, anyway.

pkg/standalone-utils/contracts/StableLPOracle.sol
```diff
@@ -0,0 +1,205 @@
...
+
+        tvl = 0;
+        for (uint256 i = 0; i < _totalTokens; i++) {
+            tvl += prices[i].toUint256().mulDown(balancesForPricesScaled18[i]);
+        }
+
+        return tvl;
+    }
+
+    function _computeBalancesForPrices(
```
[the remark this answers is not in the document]
joaobrunoah: I asked Claude to get some suggestions, and he suggested `_computePriceConsistentBalances`. Is it better?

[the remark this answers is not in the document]
jubeira: :sweat_smile: yeah why not?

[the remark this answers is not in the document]
jubeira: `balancedBalances`, `marketBalances`, `marketBalancedBalances`, `marketPriceBalances` :thinking:

pkg/standalone-utils/contracts/StableLPOracle.sol
```diff
@@ -0,0 +1,212 @@
...
+        //
+        // where `D` is the invariant, `n` is the number of tokens, `xi` is the balance of each token,
+        // `a = A*(n^2n)` and `b = a - n^n`.
+        //
+        // The gradient in terms of xj (the balance of the j-th token) is:
+        //
+        // a \prod{xi} + a \sum{xi} \prod_i!=j{xi} + bD \prod_i!=j{xi}
+        //
+        // We can make this gradient equal to k*pj, where pj is the price of the j-th token and k is a constant.
+        // Then, solving this system of equations for every pj, we will have an array of balances that respect the
```
EndymionJkb, line 83:
```suggestion
        // Then, solving this system of equations for every pj, we will have an array of balances that reflect the
```
Or "that correspond to the price vector"; we mean that a pool with these balances would have the expected token prices.

pkg/standalone-utils/contracts/StableLPOracle.sol
```diff
@@ -0,0 +1,212 @@
...
+    }
+
+    /**
+     * @dev Computes a set of balances for a given price vector and a given invariant. The set of balances should meet
+     * two conditions:
+     * 1. The invariant of the computed balances should be equal to the given invariant.
+     * 2. The spot prices of the computed balances should be equal to the given price vector. In other words:
+     *    `spotPrice(i, j) == prices[i] / prices[j]`
+     */
+    function _computeMarketPriceBalances(
```
EndymionJkb, line 66: This is only called once, during TVL calculation. Consider having this also return the TVL, since it already takes the prices as an argument, and tvl is just the sum of the product of this argument and what this function is already computing. Would save a little gas by eliminating a second loop.

pkg/standalone-utils/contracts/StableLPOracle.sol
```diff
@@ -0,0 +1,212 @@
...
+        //
+        // f(k) = T^(n+1)P - alpha = 0
+        //
+        // where `T=1+\sum{a/(a-k*pi)}`, `P = \prod{(a - k*pi)/a}` and `alpha=a*(b/a)^(n+1)`.
+        // Notice that P*T^n can be a very big number. To avoid math overflows, but keep the precision, we divide
+        // f(k) and f'(k) by P*T^n.
+
+        k = _findInitialGuessForK(a, b, prices);
+        for (uint256 i = 0; i < 255; i++) {
+            // dTdk and dPdk are the derivatives of T and P with respect to k.
```
EndymionJkb, line 114: These names are a little "Curvy" and hard to follow.

Using the convention of using prime for derivatives; e.g.,  a' (1st derivative) and a'' (2nd derivative), which we don't have here:

// tPrimeOverK = first derivative of T(totalSomething) with respect to K
T = totalSomething (descriptive name); dTdk = tPrimeOverK; dPdk = pPrimeOverK

It's not easy, but slightly longer names would be less cryptic and easier to follow. Otherwise your eyes just glaze over. Descriptive names plus ASCII art on top of it would help a lot.

pkg/standalone-utils/contracts/StableLPOracle.sol
```diff
@@ -0,0 +1,212 @@
...
+import { StableMath } from "@balancer-labs/v3-solidity-utils/contracts/math/StableMath.sol";
+
+import { LPOracleBase } from "./LPOracleBase.sol";
+
+contract StableLPOracle is LPOracleBase {
+    using FixedPoint for uint256;
+    using SafeCast for *;
+
+    // Thrown when the k parameter did not converge to the positive root.
+    error KDidntConverge();
```
EndymionJkb, line 25:
```suggestion
    error KDidNotConverge();
```
We don't usually use contractions in error names.

pkg/standalone-utils/contracts/StableLPOracle.sol
```diff
@@ -0,0 +1,212 @@
...
+        IStablePool pool_,
+        AggregatorV3Interface[] memory feeds,
+        uint256 version_
+    ) LPOracleBase(vault_, IBasePool(address(pool_)), feeds, version_) {
+        // solhint-disable-previous-line no-empty-blocks
+    }
+
+    /// @inheritdoc ILPOracleBase
+    function calculateTVL(int256[] memory prices) public view override returns (uint256 tvl) {
+        (, , , uint256[] memory lastBalancesLiveScaled18) = _vault.getPoolTokenInfo(address(pool));
```
EndymionJkb, line 41:
```suggestion
        (, , , uint256[] memory lastBalancesLiveScaled18) = _vault.getPoolTokenInfo(address(pool));
        InputHelpers.ensureInputLengthMatch(prices.length, lastBalancesLiveScaled18.length);
```
Should have some input validation here, as it's a public function.

EndymionJkb (commented):
This is really dense; first pass, not looking super deep at the math. If we have publicly accessible docs on this (vs. an internal notion), it would be good to link to them in the code. Maybe make a Notion doc public? (Do we have a public area of Notion?)

pkg/standalone-utils/contracts/StableLPOracle.sol
joaobrunoah: I wanted to split them so I can use the balances to compute the current invariant and current spot prices in a test. I mean, the TVL is not predictable, but the spot prices and the invariant are.

jubeira: yeah I think that's fine.

pkg/standalone-utils/contracts/StableLPOracle.sol
```diff
@@ -0,0 +1,213 @@
...
+        // dPdkDivP = P'/P, where P' is the derivative of P with respect to k.
+        dPdkDivP = 0;
+        // alphaDivPTn = alpha/(PT^n), where alpha is the constant term of the stable invariant.
+        alphaDivPTn = b;
+        for (i = 0; i < _totalTokens; i++) {
+            den = _mulDownInt(k, prices[i]) - a;
+            T += _divDownInt(a, den);
+            dTdk -= _divDownInt((prices[i] * a) / den, den);
+            dPdkDivP += _divDownInt(prices[i], _mulDownInt(k, prices[i]) - a);
+            alphaDivPTn = (alphaDivPTn * b) / a;
```
jubeira, line 186: I have some doubts here.

The denominator according to the notion doc is `k * prices[i] / a - 1`, so if you say here that `den = k * prices[i] - a`, you're actually saying that you need to multiply the numerator by `a`.

This is reflected correctly when computing `T`.

But for `dTdK`...
Nvm got it as I was writing this :) you multiply the numerator by `a^2` and divide by `a`, so it's fine.

pkg/standalone-utils/test/foundry/StableLPOracle.t.sol
```diff
@@ -0,0 +1,425 @@
...
+        (, , , uint256[] memory liveBalancesScaled18) = vault.getPoolTokenInfo(address(pool));
+        uint256 invariant = pool.computeInvariant(liveBalancesScaled18, Rounding.ROUND_DOWN);
+
+        uint256[] memory marketPriceBalancesScaled18 = oracle.computeMarketPriceBalances(invariant, pricesInt);
+        uint256 invariantForPrices = pool.computeInvariant(marketPriceBalancesScaled18, Rounding.ROUND_DOWN);
+
+        assertApproxEqRel(invariantForPrices, invariant, 1e4, "Invariant does not match");
+    }
+
+    function testComputeMarketPriceBalances__Fuzz(
```
jubeira, line 272: wow, this thing really works :grinning:

jubeira (commented):
Awesome job. 

The new e2e test looks pretty convincing. I'm amazed this works with this degree of precision.

I only have a few minor comments to improve readability just a bit, plus a few questions. Other than that this is already pretty solid. Thanks for pushing it!

pkg/standalone-utils/contracts/StableLPOracle.sol
```diff
@@ -0,0 +1,213 @@
...
+            int256 ri = _divDownInt(prices[i], a);
+            // P = \prod(k*ri - 1) . So, to divide alpha by PT^n, we can iteratively divide alpha by T and (k*ri - 1).
+            den = _mulDownInt(k, ri) - _ONE_INT;
+            alphaDivPTn = _divDownInt(alphaDivPTn, _mulDownInt(den, T));
+        }
+    }
+
+    function _computeAAndBForPool(IStablePool pool) internal view returns (int256 a, int256 b) {
+        (uint256 amplificationFactor, , ) = pool.getAmplificationParameter();
+        // a = A * n^2n, but A = ampParameter * n^(n-1). So, a = ampParameter * n^(2n)/n^(n-1) = ampParameter * n^(n+1).
```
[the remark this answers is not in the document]
joaobrunoah: It came from a conversation between Sergio and I, but also it comes from how the stable invariant is defined in the Stable Math. If you recreate the math of the stable invariant from the contract, you'll notice that `ampTimesTotal = amplificationParameter * numTokens`, but actually it represents `A*n^n`. So, there's a `n^(n-1)` missing, which must be inside the `amplificationParameter`.

pkg/standalone-utils/contracts/StableLPOracle.sol
```diff
@@ -0,0 +1,212 @@
...
+     *    `spotPrice(i, j) == prices[i] / prices[j]`
+     */
+    function _computeMarketPriceBalances(
+        uint256 invariant,
+        int256[] memory prices
+    ) internal view returns (uint256[] memory balancesForPrices) {
+        // To compute the balances for a given price vector, we need to compute the gradient of the stable invariant.
+        // The stable invariant is:
+        //
+        // a \sum{xi} \prod{xi} +bD \prod{xi} - D^(n+1) = 0
```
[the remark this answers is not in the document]
joaobrunoah: I used the same notation as Stable Math library now, using S as Sum and P as Product. (Couldn't reproduce the Sum symbol in 3 lines, and that way is more straightforward)

pkg/standalone-utils/contracts/StableLPOracle.sol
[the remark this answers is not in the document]
joaobrunoah: By the way, there's no mistake. `ampParameter = A * n^(n-1)`, so `A = ampParameter / n^(n-1)`

pkg/standalone-utils/contracts/StableLPOracle.sol
```diff
@@ -147,67 +158,65 @@ contract StableLPOracle is LPOracleBase {
...
 
     function _computeAAndBForPool(IStablePool pool) internal view returns (int256 a, int256 b) {
-        (uint256 amplificationFactor, , ) = pool.getAmplificationParameter();
-        // a = A * n^2n, but A = ampParameter * n^(n-1). So, a = ampParameter * n^(2n)/n^(n-1) = ampParameter * n^(n+1).
-        a = int256((amplificationFactor * (_totalTokens ** (_totalTokens + 1))).divDown(StableMath.AMP_PRECISION));
+        (uint256 amplificationParameter, , ) = pool.getAmplificationParameter();
+        // In the StableMath library, `amplificationParameter = A*n^(n-1)` (For more information, check the
+        // `computeInvariant` natspec of the StableMath library).
+        // a = A * n^2n, but A = ampParameter / n^(n-1). So, a = ampParameter * n^(2n)/n^(n-1) = ampParameter * n^(n+1).
+        a = int256((amplificationParameter * (_totalTokens ** (_totalTokens + 1))).divDown(StableMath.AMP_PRECISION));
```
jubeira, line 211: Let's maybe add a small comment explaining why the unsafe cast is actually safe. It should be because we're dividing by AMP_PRECISION.

Also, why is it a FP operation instead of a raw one?

jubeira (commented):
This is looking pretty good. Sorry for the back and forth about naming, I liked some of the old more compact ones better.

pkg/standalone-utils/contracts/StableLPOracle.sol
joaobrunoah: When I do an FP division, I scale `a` to 18 decimals. Same with `b`. So, I can use these variables as scaled 18 variables, which is prety convenient when I'm doing divisions like `a/b`, where I need floating points. I'll add a comment about it.

pkg/standalone-utils/contracts/StableLPOracle.sol
```diff
@@ -0,0 +1,225 @@
...
+     * computations (e.g. `b/a`) we need FP precision, so return these variables as scaled-18 is convenient.
+     */
+    function _computeAAndBForPool(IStablePool pool) internal view returns (int256 a, int256 b) {
+        (uint256 amplificationParameter, , ) = pool.getAmplificationParameter();
+        // In the StableMath library, `amplificationParameter = A*n^(n-1)` (For more information, check the
+        // `computeInvariant` natspec of the StableMath library). `a = A * n^2n` and `A = ampParameter / n^(n-1)`. So,
+        // `a = ampParameter * n^(2n)/n^(n-1) = ampParameter * n^(n+1)`.
+        // Since `2 <= totalTokens <= 5`, and 1 <= amplificationParameter/AMP_PRECISION <= 50000, the max value of `a`
+        // is 5^6 * 50000 = 781,250,000, which multiplied by 1e18 is far from the int256 max positive value of 5.78e76.
+        // Since there's no risk of Underflow/Overflow, we don't need to use SafeCast.
```
EndymionJkb, line 213:
```suggestion
        // In the StableMath library, `amplificationParameter` = A * n^(n-1)
        // (For more information, check the `computeInvariant` NatSpec of the StableMath library.)
        //
        // Given that:
        // `A` = ampParameter / n^(n-1); and
        // `a` = A * n^2n
        //
        // We have: `a` = ampParameter * n^(2n) / n^(n-1) = ampParameter * n^(n+1)
        //
        // Since `2 <= totalTokens <= 5`; and 
        // 1 <= amplificationParameter / AMP_PRECISION <= 50,000
        //
        // The max value of `a` is given by: 5^6 * 50000 = 781,250,000
        // This value, multiplied by 1e18, is far less than the int256 max positive value of 5.78e76.
        // Since there's no risk of Underflow/Overflow, we don't need to use SafeCast.
```
This is kind of squished together. Consider this for legibility.

pkg/standalone-utils/contracts/WeightedLPOracle.sol
```diff
@@ -8,27 +8,21 @@ import { SafeCast } from "@openzeppelin/contracts/utils/math/SafeCast.sol";
 import { IERC20 } from "@openzeppelin/contracts/token/ERC20/IERC20.sol";
```
EndymionJkb, line 8:
```suggestion
```
Can also remove IERC20Metadata (not in diff)

pkg/standalone-utils/contracts/WeightedLPOracle.sol
```diff
@@ -8,27 +8,21 @@ import { SafeCast } from "@openzeppelin/contracts/utils/math/SafeCast.sol";
...
 
 import { IWeightedLPOracle } from "@balancer-labs/v3-interfaces/contracts/standalone-utils/IWeightedLPOracle.sol";
+import { ILPOracleBase } from "@balancer-labs/v3-interfaces/contracts/standalone-utils/ILPOracleBase.sol";
 import { IWeightedPool } from "@balancer-labs/v3-interfaces/contracts/pool-weighted/IWeightedPool.sol";
 import { Rounding } from "@balancer-labs/v3-interfaces/contracts/vault/VaultTypes.sol";
+import { IBasePool } from "@balancer-labs/v3-interfaces/contracts/vault/IBasePool.sol";
 import { IVault } from "@balancer-labs/v3-interfaces/contracts/vault/IVault.sol";
 
 import { InputHelpers } from "@balancer-labs/v3-solidity-utils/contracts/helpers/InputHelpers.sol";
```
EndymionJkb, line 17:
```suggestion
```
Unused import

pkg/standalone-utils/test/foundry/StableLPOracle.t.sol
```diff
@@ -0,0 +1,425 @@
...
+import { IERC20 } from "@openzeppelin/contracts/token/ERC20/IERC20.sol";
+
+import { PoolRoleAccounts, Rounding } from "@balancer-labs/v3-interfaces/contracts/vault/VaultTypes.sol";
+import { IStablePool } from "@balancer-labs/v3-interfaces/contracts/pool-stable/IStablePool.sol";
+import { IVault } from "@balancer-labs/v3-interfaces/contracts/vault/IVault.sol";
+
+import { CastingHelpers } from "@balancer-labs/v3-solidity-utils/contracts/helpers/CastingHelpers.sol";
+import { InputHelpers } from "@balancer-labs/v3-solidity-utils/contracts/helpers/InputHelpers.sol";
+import { StablePoolFactory } from "@balancer-labs/v3-pool-stable/contracts/StablePoolFactory.sol";
+import { ArrayHelpers } from "@balancer-labs/v3-solidity-utils/contracts/test/ArrayHelpers.sol";
```
EndymionJkb, line 18:
```suggestion
```
Unused

EndymionJkb (approved):
Everything I saw (mostly docs and nits) looks good, and I think the math has been pretty thoroughly covered.

elshan-eth:
Oh, I’m late with the review. This looks amazing! You’re a real math genius, when I was working on the oracle for the Weighted Pool, I couldn’t figure out how to do something like this for StablePool.

joaobrunoah:
> Oh, I’m late with the review. This looks amazing! You’re a real math genius, when I was working on the oracle for the Weighted Pool, I couldn’t figure out how to do something like this for StablePool.

@elshan-eth kkkkkk, me neither! Sergio made a huge lift to reach something like this, I only reproduced what he thought :)

PR #1399, opened 2025-06-06
Add weighted lp oracle factory

elshan-eth:
# Description

This PR adds a factory for easier deployment of oracles for weighted pools.

## Type of change

- [ ] Bug fix 
- [X] New feature 
- [ ] Breaking change 
- [ ] Dependency changes
- [ ] Code refactor / cleanup
- [ ] Other

## Checklist:


pkg/standalone-utils/contracts/WeightedLPOracleFactory.sol
```diff
@@ -0,0 +1,54 @@
...
+    uint256 internal _oracleVersion;
+    mapping(IWeightedPool => IWeightedLPOracle) internal _oracles;
+
+    constructor(IVault vault, uint256 oracleVersion) SingletonAuthentication(vault) {
+        _oracleVersion = oracleVersion;
+    }
+
+    function create(
+        IWeightedPool pool,
+        AggregatorV3Interface[] memory feeds
```
jubeira, line 28: Let's check that the pool is registered in the vault.

Not sure whether there's any good check for the feeds, but let's check at least that the length matches the pool tokens and that they are not null.

pkg/standalone-utils/contracts/WeightedLPOracleFactory.sol
```diff
@@ -0,0 +1,54 @@
...
+            revert OracleAlreadyExists();
+        }
+
+        oracle = new WeightedLPOracle(getVault(), pool, feeds, _oracleVersion);
+
+        _oracles[pool] = oracle;
+        emit WeightedLPOracleCreated(pool, oracle);
+    }
+
+    function removeOracle(IWeightedPool pool) external authenticate {
```
jubeira, line 40: Usually for factories we use a `disable` pattern instead of outright removing it, where `disable` just prevents the creation of new oracles.

I don't think we need the capability to remove them from here.

pkg/interfaces/contracts/standalone-utils/IWeightedLPOracleFactory.sol
```diff
@@ -0,0 +1,52 @@
...
+import { AggregatorV3Interface } from "@chainlink/contracts/src/v0.8/shared/interfaces/AggregatorV3Interface.sol";
+
+import { IWeightedPool } from "../pool-weighted/IWeightedPool.sol";
+import { IWeightedLPOracle } from "./IWeightedLPOracle.sol";
+
+/**
+ * @title Weighted LP Oracle Factory
+ * @notice Factory contract for deploying and managing lp weighted pool oracles.
+ */
+interface IWeightedLPOracleFactory {
```
jubeira, line 14: Let's also add a getter that returns true given an oracle address if it comes from this factory, and false otherwise (similar to `isPoolFromFactory`).

Unlike the factories, I don't think we need to store the list of deployed oracles in this case.

jubeira (commented):
Nice job @elshan-eth ! Pretty clean.

Just a few minor comments, otherwise LGTM!

jubeira (approved):
LGTM, thanks for pushing this!

PR #1404, opened 2025-06-17
LBP Migration Router Refactoring

elshan-eth:
# Description

This PR contains a refactoring of the LBP migration contracts. The migration parameters have been moved to the pool, the factory has an additional method for deployment pool with migration, the migration router has been simplified, The and the timelock is implemented using the ERC6909 standard (to support it, OpenZeppelin version in npm modules had to be updated).

## Type of change

- [ ] Bug fix 
- [ ] New feature 
- [ ] Breaking change 
- [ ] Dependency changes
- [X] Code refactor / cleanup
- [ ] Other

## Checklist:


EndymionJkb (commented):
See #1405. Was too much to do through comments.

pkg/pool-weighted/contracts/lbp/LBPool.sol
```diff
@@ -156,6 +164,11 @@ contract LBPool is ILBPool, WeightedPool, Ownable2Step, BaseHooks {
             lbpParams.reserveTokenEndWeight
         );
 
+        _lockDurationAfterMigration = lockDurationAfterMigration;
+        _bptPercentageToMigrate = bptPercentageToMigrate;
+        _migrationWeightProjectToken = migrationWeightProjectToken;
+        _migrationWeightReserveToken = migrationWeightReserveToken;
```
jubeira, line 170: Let's also check these parameters here.

Migration will fail if `_migrationWeightProjectToken + _migrationWeightReserveToken != FixedPoint.ONE`, and that would lock the liquidity if I'm not mistaken since we won't allow regular exits (well, we would still have recovery mode but that's not ideal).

So it's better to fail early and disallow creating the LBP if the migration parameters are incorrect.

Also the lock duration should have a maximum duration (1 year? 4 years? But not more than that).
BPT percentage to migrate should be < FP(1) too.

pkg/pool-weighted/contracts/lbp/LBPoolFactory.sol
```diff
@@ -120,13 +195,45 @@ contract LBPoolFactory is IPoolVersion, ReentrancyGuardTransient, BasePoolFactor
...
 
+        address migrationRouterOrZero = hasMigration ? _migrationRouter : address(0);
+        if (
+            hasMigration &&
+            (migrationWeightProjectToken + migrationWeightReserve != FixedPoint.ONE ||
+                migrationWeightProjectToken == 0 ||
+                migrationWeightReserve == 0)
+        ) {
+            revert InvalidMigrationWeights();
+        }
```
jubeira, line 206: :+1:

pkg/pool-weighted/contracts/lbp/LBPool.sol
```diff
@@ -168,12 +181,30 @@ contract LBPool is ILBPool, WeightedPool, Ownable2Step, BaseHooks {
...
+     * @dev The migration parameters are set at deployment and cannot be changed later.
+     * @return migrationRouter The address of the migration router allowed to migrate this LBP
+     * @return lockDuration The duration for which the BPTs will be locked after migration
+     * @return bptPercentageToMigrate The percentage of the BPT to migrate from the LBP to the new weighted pool
+     * @return weightProjectToken The weight of the project token in the new weighted pool
+     * @return weightReserveToken The weight of the reserve token in the new weighted pool
      */
-    function getMigrationRouter() external view returns (address) {
-        return _migrationRouter;
+    function getMigrationParams()
```
jubeira, line 192: We might as well add these to the pool's immutable params

pkg/pool-weighted/contracts/lbp/Timelocker.sol
```diff
@@ -0,0 +1,91 @@
...
+pragma solidity ^0.8.24;
+
+import { ERC6909Metadata } from "@openzeppelin/contracts/token/ERC6909/extensions/draft-ERC6909Metadata.sol";
+import { IERC20Metadata } from "@openzeppelin/contracts/token/ERC20/extensions/IERC20Metadata.sol";
+import { SafeERC20 } from "@openzeppelin/contracts/token/ERC20/utils/SafeERC20.sol";
+import { ERC6909 } from "@openzeppelin/contracts/token/ERC6909/draft-ERC6909.sol";
+import { IERC20 } from "@openzeppelin/contracts/token/ERC20/IERC20.sol";
+import { Multicall } from "@openzeppelin/contracts/utils/Multicall.sol";
+
+contract Timelocker is ERC6909, ERC6909Metadata, Multicall {
```
jubeira, line 12: how about `BPTTimeLocker` or `BPTLocker`?

jubeira (commented):
This is pretty cool! I like the changes a lot.

First pass is done. I've left a handful of mostly minor comments, other than that it's pretty close to ready.

On top of that, can we restore the test introduced here? I think it got lost in the refactor: https://github.com/balancer/balancer-v3-monorepo/pull/1390/commits/209227badb71bf9d60a8d511e820851f6b2a4a30

Thanks for pushing this again! I like the outcome of the design decisions you had to take along the way too.

pkg/pool-weighted/contracts/lbp/LBPool.sol
elshan-eth: Sorry, I didn't quite get this comment. These params are immutable

EndymionJkb (commented):
Getting closer! Some more naming and documentation suggestions.

pkg/pool-weighted/contracts/lbp/BPTTimeLocker.sol
```diff
@@ -0,0 +1,92 @@
...
+    /// @notice The caller has no balance of the locked BPT.
+    error NoLockedBPT();
+
+    mapping(uint256 => uint256) internal _unlockTimestamps;
+
+    /**
+     * @notice Burn the locked tokens for the caller.
+     * @param bptAddress The address of the BPT to burn
+     */
+    function burn(address bptAddress) public {
```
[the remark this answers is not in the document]
elshan-eth: About Reentrancy, I think no, because the check-effects-interactions pattern is used here

PR #1405, opened 2025-06-19
LBP migration edits

EndymionJkb:
# Description

The review of #1404 got a bit out of hand :)

It was easier to just make the changes, as many were outside the diff. No semantic changes, just some cleanup/docs/renames. This can be merged into #1404.

// This really doesn't need to be generalized; it's used for one purpose, so can be specific
ContractIsNotActiveInRegistry(string contractName) -> NoRegisteredWeightedPoolFactory();

// Finalized sounds like V1 (at least to those who go back that far); I reused the name of the underlying LBPool error
LBPWeightsNotFinalized -> RemovingLiquidityNotAllowed

// "share" isn't wrong, but perhaps bptPercentage is clearer
shareToMigrate -> bptPercentageToMigrate

// Big one responsible for a lot of the diff: as in ReCLAMM, using 0/1 is confusing, especially when we have two well-defined tokens (project and reserve). Using the names also makes things like price calculations easier to follow.

Timelocker introduces a lot of new OpenZeppelin libraries. It would be helpful to document what they are, why they're used, etc.

Most of the imports were unused, and there were some additional renames. "Amount" isn't really involved in the error; the issue is the timelock expiration, so renamed to focus on that. Events have the verb at the end, and using a regular address for the BPT is clearer than a numerical value (even though they're the same).

SetUnlockTimestamp -> BPTTimelockSet (and use address in the event for clarity)
AmountNotUnlockedYet -> BPTStillLocked
NoLockedAmount -> NoLockedBPT

## Type of change

- [ ] Bug fix 
- [ ] New feature 
- [ ] Breaking change 
- [ ] Dependency changes
- [X] Code refactor / cleanup
- [ ] Other

## Checklist:

### DIFF
--- a/pkg/interfaces/contracts/pool-weighted/ILBPMigrationRouter.sol
+++ b/pkg/interfaces/contracts/pool-weighted/ILBPMigrationRouter.sol
@@ -5,8 +5,8 @@ pragma solidity ^0.8.24;
 import { IERC20 } from "@openzeppelin/contracts/token/ERC20/IERC20.sol";
 
 import { PoolRoleAccounts } from "../vault/VaultTypes.sol";
-import { ILBPool } from "./ILBPool.sol";
 import { IWeightedPool } from "./IWeightedPool.sol";
+import { ILBPool } from "./ILBPool.sol";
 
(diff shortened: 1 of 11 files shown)


elshan-eth:
Thanks a lot!

PR #1423, opened 2025-07-07
Add unbalanced liquidity support for pools that allow only proportional adds

elshan-eth:
# Description

This PR adds a combined operation that allows adding unbalanced liquidity using addProportional + swap. The combination needs to be calculated off-chain, and the parameters should be sent to this router extension.

## Type of change

- [ ] Bug fix 
- [X] New feature 
- [ ] Breaking change 
- [ ] Dependency changes
- [ ] Code refactor / cleanup
- [ ] Other

## Checklist:


EndymionJkb (commented):
See #1428 and #1429 :)

I didn't try to change the name or the diff would be useless, but I'm not sure of the contract name here. We have only used "Extension" with the Proxy pattern, where this is only a conceptual extension. We've just been naming routers <function>Router. Maybe just AddUnbalanced(ViaSwap)Router? (Or just UnbalancedLiquidityRouter, as below, especially if we add removing.)

Are we also going to have a remove unbalanced? Then it could do both, and be called something like UnbalancedLiquidityRouter.

elshan-eth:
@jubeira This PR including ETH logic since the solution turned out to be simple

pkg/interfaces/contracts/vault/IAddUnbalancedLiquidityViaSwapRouter.sol
```diff
@@ -0,0 +1,110 @@
...
+
+/// @notice Router interface for adding unbalanced liquidity via a combination of a proportional add and a swap.
+interface IAddUnbalancedLiquidityViaSwapRouter {
+    struct AddLiquidityProportionalParams {
+        uint256[] maxAmountsIn;
+        uint256 exactBptAmountOut;
+        bytes userData;
+    }
+
+    struct SwapExactInParams {
```
EndymionJkb, line 22: In the main contract you're using a combined struct. Consider defining that up here and using one struct instead of three. I tried it; it's not a big change and works.

pkg/vault/contracts/AddUnbalancedLiquidityViaSwapRouter.sol
```diff
@@ -0,0 +1,228 @@
...
+        (amountsIn, , ) = _addLiquidityHook(params.addLiquidityParams);
+        swapAmountCalculated = _swapSingleTokenHook(params.swapParams);
+    }
+
+    function queryAddUnbalancedLiquidityViaSwapHook(
+        AddLiquidityAndSwapHookParams calldata params
+    ) external nonReentrant onlyVault returns (uint256[] memory amountsIn, uint256 swapAmountCalculated) {
+        (amountsIn, , ) = _queryAddLiquidityHook(params.addLiquidityParams);
+        swapAmountCalculated = _querySwapHook(params.swapParams);
+    }
```
jubeira, line 164: I'm a bit on the fence with this interface. It's better than before, but I'm leaning towards one more twist unless you think I'm missing something.

From the outside this is just a way to add liquidity, so the swap amount calculated doesn't really matter.
Perhaps it's fine to return these values in the internal hooks, but maybe I'd modify the external interface. See comment below.

pkg/vault/contracts/AddUnbalancedLiquidityViaSwapRouter.sol
```diff
@@ -0,0 +1,228 @@
...
+
+    /// @inheritdoc IAddUnbalancedLiquidityViaSwapRouter
+    function addUnbalancedLiquidityViaSwapExactIn(
+        address pool,
+        uint256 deadline,
+        bool wethIsEth,
+        AddLiquidityProportionalParams calldata addLiquidityParams,
+        SwapExactInParams calldata swapParams
+    ) external payable saveSenderAndManageEth returns (uint256[] memory amountsIn, uint256 swapAmountOut) {
```
jubeira, line 51: Here and elsewhere (applies to exact out and queries):
I think it'd be better to just return the actual amounts in, where:
- `amountsIn'[swapTokenIn] = amountsIn[swapTokenIn] + swapAmountIn`
- `amountsIn'[swapTokenOut] = amountsIn[swapTokenOut] - swapAmountOut`

Not sure whether it's really needed to split the two quantities for the caller. WDYT?

pkg/vault/contracts/AddUnbalancedLiquidityViaSwapRouter.sol
```diff
@@ -0,0 +1,228 @@
...
+        return
+            SwapParams({
+                tokenIn: swapParams.tokenIn,
+                tokenOut: swapParams.tokenOut,
+                kind: SwapKind.EXACT_OUT,
+                amountGiven: swapParams.exactAmountOut,
+                limit: swapParams.maxAmountIn,
+                userData: swapParams.userData
+            });
+    }
```
jubeira, line 227: This only adds the kind right?
Perhaps you can build `_buildAddLiquidityParams` directly from `SwapExactInParams` or `SwapExactOutParams` + kind? Or does that cause stack too deep?

(just to clarify, this looks fine as is and there's no need to make changes; I'm just curious).

jubeira (approved):
LGTM, very clean!

Just a few super minor comments. I really like the end result; great job! :100:

pkg/vault/contracts/AddUnbalancedLiquidityViaSwapRouter.sol
```diff
@@ -0,0 +1,228 @@
...
+
+    function queryAddUnbalancedLiquidityViaSwapHook(
+        AddLiquidityAndSwapHookParams calldata params
+    ) external nonReentrant onlyVault returns (uint256[] memory amountsIn, uint256 swapAmountCalculated) {
+        (amountsIn, , ) = _queryAddLiquidityHook(params.addLiquidityParams);
+        swapAmountCalculated = _querySwapHook(params.swapParams);
+    }
+
+    /***************************************************************************
+                                   Others functions
```
EndymionJkb, line 167:
```suggestion
                                   Helper functions
```

pkg/vault/test/foundry/AddUnbalancedLiquidityViaSwapRouter.t.sol
```diff
@@ -0,0 +1,250 @@
...
+
+import { IERC20 } from "@openzeppelin/contracts/token/ERC20/IERC20.sol";
+
+import { SwapKind, PoolRoleAccounts } from "@balancer-labs/v3-interfaces/contracts/vault/VaultTypes.sol";
+import {
+    IAddUnbalancedLiquidityViaSwapRouter
+} from "@balancer-labs/v3-interfaces/contracts/vault/IAddUnbalancedLiquidityViaSwapRouter.sol";
+import { IVault } from "@balancer-labs/v3-interfaces/contracts/vault/IVault.sol";
+
+import { WeightedPoolFactory } from "@balancer-labs/v3-pool-weighted/contracts/WeightedPoolFactory.sol";
```
EndymionJkb, line 15:
```suggestion
```
Unused import

pkg/interfaces/contracts/vault/IAddUnbalancedLiquidityViaSwapRouter.sol
```diff
@@ -0,0 +1,110 @@
...
+pragma solidity ^0.8.24;
+
+import { IERC4626 } from "@openzeppelin/contracts/interfaces/IERC4626.sol";
+import { IERC20 } from "@openzeppelin/contracts/token/ERC20/IERC20.sol";
+
+import { IVault } from "./IVault.sol";
+import { IRouterQueries } from "./IRouterQueries.sol";
+
+import { AddLiquidityKind, RemoveLiquidityKind, SwapKind } from "./VaultTypes.sol";
+import "./RouterTypes.sol";
```
EndymionJkb, line 12:
```suggestion
import { IERC20 } from "@openzeppelin/contracts/token/ERC20/IERC20.sol";

import { AddLiquidityKind, RemoveLiquidityKind, SwapKind } from "./VaultTypes.sol";
import { IVault } from "./IVault.sol";
import "./RouterTypes.sol";
```
Unused and ordering

EndymionJkb (commented):
Minor comments.

pkg/vault/contracts/AddUnbalancedLiquidityViaSwapRouter.sol
```diff
@@ -0,0 +1,228 @@
...
+
+import { RouterHooks } from "./RouterHooks.sol";
+
+/**
+ * @notice Enable adding and removing liquidity unbalanced on pools that do not support it natively.
+ * @dev This extends the standard `Router` in order to call shared internal hook implementation functions.
+ * It factors out the unbalanced adds into two operations: a proportional add and a swap, executes them using
+ * the standard router, then checks the limits.
+ */
+contract AddUnbalancedLiquidityViaSwapRouter is RouterHooks, IAddUnbalancedLiquidityViaSwapRouter {
```
jubeira, line 24: Late comment: in the tests we're using the regular router to query swaps and adds individually.

Maybe it's worth inheriting from `RouterQueries` here instead, making those entrypoints available?

pkg/interfaces/contracts/vault/IAddUnbalancedLiquidityViaSwapRouter.sol
EndymionJkb: I mean you have SwapParams defined in the `AddUnbalancedLiquidityViaSwapRouter` contract. You could move that definition up to the interface `AddUnbalancedLiquidityViaSwapRouter`, and use it instead of defining the two structs: `SwapExactInParams` and `SwapExactOutParams`. The functions would still be separate; you're just using one struct instead of three.

PR #1424, opened 2025-07-07
Refactoring the Router contract

elshan-eth:
# Description

This PR extracts logic into the BaseRouter so that it can be reused in other routers. It does not yet integrate BaseRouter into the AggregatorRouter to keep the number of changes minimal.

## Type of change

- [ ] Bug fix 
- [ ] New feature 
- [ ] Breaking change 
- [ ] Dependency changes
- [x] Code refactor / cleanup
- [ ] Other

## Checklist:


pkg/vault/contracts/Router.sol
```diff
@@ -2,36 +2,28 @@
...
+ * @notice Base contract for routing swaps and liquidity operations through custom logic.
+ * @dev Provides shared functionality for derived routers, such as managing approvals, performing token transfers,
+ * and interacting with external pools or Vaults. Designed to be extended with specific routing behavior.
  */
-contract Router is IRouter, RouterCommon {
-    using Address for address payable;
-    using RouterWethLib for IWETH;
-    using SafeCast for *;
-
+contract Router is IRouter, BaseRouter {
```
jubeira, line 20: This is fine, but I'm thinking it'd be great to preserve the `IRouter` interface as an all-in-one interface.

In other words, if I wanted to use the router for something I should be able to just import and interface and get the ball rolling. Same that happens with `IVault`, which is a wide net that catches everything.

Can we achieve the same here? Maybe if `IRouter` extends `IBaseRouter`?

pkg/interfaces/contracts/vault/IBaseRouter.sol
```diff
@@ -0,0 +1,257 @@
...
+import { IVault } from "./IVault.sol";
+
+import { AddLiquidityKind, RemoveLiquidityKind, SwapKind } from "./VaultTypes.sol";
+
+interface IBaseRouter {
+    /**
+     * @notice The sender does not transfer the correct amount of tokens to the Vault.
+     * @param token The address of the token that was expected to be transferred.
+     */
+    error InsufficientPayment(IERC20 token);
```
jubeira, line 17: I think this is from the aggregator router right?

pkg/interfaces/contracts/vault/IBaseRouter.sol
```diff
@@ -0,0 +1,257 @@
...
+pragma solidity ^0.8.24;
+
+import { IERC4626 } from "@openzeppelin/contracts/interfaces/IERC4626.sol";
+import { IERC20 } from "@openzeppelin/contracts/token/ERC20/IERC20.sol";
+
+import { IVault } from "./IVault.sol";
+
+import { AddLiquidityKind, RemoveLiquidityKind, SwapKind } from "./VaultTypes.sol";
+
+interface IBaseRouter {
```
jubeira, line 12: This could also be `IRouterQueries`, except that you'll also need the data structs afterwards right?

Do you think it'd be worth splitting this into `IRouterQueries` and `RouterTypes`, like we do with the vault, or is it too much?

pkg/vault/contracts/Router.sol
```diff
@@ -553,12 +373,12 @@ contract Router is IRouter, RouterCommon {
         uint256 deadline,
         bool wethIsEth,
         bytes calldata userData
-    ) external payable saveSender(msg.sender) returns (uint256) {
+    ) public payable saveSender(msg.sender) returns (uint256) {
```
jubeira, line 376: can we keep this just external?

pkg/vault/contracts/Router.sol
```diff
@@ -587,12 +407,12 @@ contract Router is IRouter, RouterCommon {
...
```
jubeira, line 410: (same as above)

jubeira (commented):
Thank you very much for pushing this; it really made my life much easier.

I only have very minor questions / comments, otherwise LGTM. My main comment is that I'd like to preserve a single interface for the whole router if possible.

Imagine building reclamm as an example over the monorepo. If I was using `IRouter`, ideally it shouldn't change because of this refactor, and I should be able to just use the same interface as usual (ideally, if it's not possible let's discuss it).

pkg/interfaces/contracts/vault/IBaseRouter.sol
joaobrunoah: At first I thought that separating Types and Queries into two different contracts would be an overkill, since no contract would call "RouterTypes" separated from IRouterQueries, but thinking twice it can improve the UX from developers integrating with Balancer routers.

pkg/vault/contracts/RouterHooks.sol
```diff
@@ -14,24 +14,27 @@ import "@balancer-labs/v3-interfaces/contracts/vault/VaultTypes.sol";
...
+    /**
+     * @notice The sender does not transfer the correct amount of tokens to the Vault.
+     * @param token The address of the token that was expected to be transferred.
+     */
+    error InsufficientPayment(IERC20 token);
+
+    bool internal immutable _isAggregator;
+
+    constructor(bool isAggregator) {
+        _isAggregator = isAggregator;
```
jubeira, line 37: Sorry for being annoying, but I'd much rather leave everything related with this flag for another PR.

The current `Router` only knows how to settle with permit2 in the main branch, and that should stay the same in this PR. This PR was meant to be a refactor only so that we can focus on restructuring the code specifically, without adding anything. The diff is already quite big, and I want to change a few things around settlement when `isAggregator` is true. But we'll be able to discuss that more efficiently in a different scope.

Can we keep the router's features the same for the time being?

pkg/vault/contracts/RouterHooks.sol
```diff
@@ -0,0 +1,402 @@
...
+    /**
+     * @notice Hook for initialization.
+     * @dev Can only be called by the Vault.
+     * @param params Initialization parameters (see IRouter for struct definition)
+     * @return bptAmountOut BPT amount minted in exchange for the input tokens
+     */
+    function initializeHook(
+        InitializeHookParams calldata params
+    ) external nonReentrant onlyVault returns (uint256 bptAmountOut) {
+        return _initializeHook(params);
```
jubeira, line 46: (here and elsewhere): is there a particular reason to have an internal version of the hooks if it's used only once, as opposed to just inline the code of the internal function here?

I'm not strongly against it, but if the internal function is only used once I'd say inlining it is better. Less boilerplate code around.

jubeira (approved):
Just one small final comment, otherwise LGTM!

I'm not a huge fan of adding yet another layer in the middle of the inheritance chain, but this does maximize code re-utilization. So in this case it seems to be the right choice.

pkg/vault/contracts/CompositeLiquidityRouter.sol
```diff
@@ -10,6 +10,7 @@ import { IVaultErrors } from "@balancer-labs/v3-interfaces/contracts/vault/IVaul
 import { IWETH } from "@balancer-labs/v3-interfaces/contracts/solidity-utils/misc/IWETH.sol";
 import { IVault } from "@balancer-labs/v3-interfaces/contracts/vault/IVault.sol";
 import "@balancer-labs/v3-interfaces/contracts/vault/VaultTypes.sol";
+import "@balancer-labs/v3-interfaces/contracts/vault/RouterTypes.sol";
```
EndymionJkb, line 13:
```suggestion
import "@balancer-labs/v3-interfaces/contracts/vault/RouterTypes.sol";
import "@balancer-labs/v3-interfaces/contracts/vault/VaultTypes.sol";
```
Nit: reorder for line-length

pkg/vault/contracts/Router.sol
```diff
@@ -2,36 +2,30 @@
...
-import { IERC4626 } from "@openzeppelin/contracts/interfaces/IERC4626.sol";
-import { SafeCast } from "@openzeppelin/contracts/utils/math/SafeCast.sol";
 import { IERC20 } from "@openzeppelin/contracts/token/ERC20/IERC20.sol";
-import { Address } from "@openzeppelin/contracts/utils/Address.sol";
 import { IPermit2 } from "permit2/src/interfaces/IPermit2.sol";
 
 import { IWETH } from "@balancer-labs/v3-interfaces/contracts/solidity-utils/misc/IWETH.sol";
 import { IRouter } from "@balancer-labs/v3-interfaces/contracts/vault/IRouter.sol";
 import { IVault } from "@balancer-labs/v3-interfaces/contracts/vault/IVault.sol";
 import "@balancer-labs/v3-interfaces/contracts/vault/VaultTypes.sol";
```
EndymionJkb, line 11:
```suggestion
```
Unused import.

pkg/vault/contracts/Router.sol
```diff
@@ -2,36 +2,30 @@
...
+import "@balancer-labs/v3-interfaces/contracts/vault/RouterTypes.sol";
 
-import { RouterWethLib } from "./lib/RouterWethLib.sol";
-import { RouterCommon } from "./RouterCommon.sol";
+import { RouterHooks } from "./RouterHooks.sol";
+import { RouterQueries } from "./RouterQueries.sol";
```
EndymionJkb, line 15:
```suggestion
import { RouterQueries } from "./RouterQueries.sol";
import { RouterHooks } from "./RouterHooks.sol";
```
Order nit

pkg/vault/contracts/RouterCommon.sol
```diff
@@ -298,7 +298,7 @@ abstract contract RouterCommon is IRouterCommon, SenderGuard, VaultGuard, Reentr
      * result in locked funds, but are not otherwise a security or soundness issue. This check only exists as an attempt
      * to prevent user error.
      */
-    receive() external payable {
+    receive() external payable virtual {
```
EndymionJkb, line 301: Not in diff, but unused imports + ordering:

```
import { IERC20Permit } from "@openzeppelin/contracts/token/ERC20/extensions/IERC20Permit.sol";
import { SafeCast } from "@openzeppelin/contracts/utils/math/SafeCast.sol";
import { Address } from "@openzeppelin/contracts/utils/Address.sol";
import { IPermit2 } from "permit2/src/interfaces/IPermit2.sol";

import { IRouterCommon } from "@balancer-labs/v3-interfaces/contracts/vault/IRouterCommon.sol";
import { IWETH } from "@balancer-labs/v3-interfaces/contracts/solidity-utils/misc/IWETH.sol";
import { IAllowanceTransfer } from "permit2/src/interfaces/IAllowanceTransfer.sol";

import { StorageSlotExtension } from "@balancer-labs/v3-solidity-utils/contracts/openzeppelin/StorageSlotExtension.sol";
import { InputHelpers } from "@balancer-labs/v3-solidity-utils/contracts/helpers/InputHelpers.sol";
import { RevertCodec } from "@balancer-labs/v3-solidity-utils/contracts/helpers/RevertCodec.sol";
import {
    ReentrancyGuardTransient
} from "@balancer-labs/v3-solidity-utils/contracts/openzeppelin/ReentrancyGuardTransient.sol";
import { Version } from "@balancer-labs/v3-solidity-utils/contracts/helpers/Version.sol";
import {
    TransientStorageHelpers
} from "@balancer-labs/v3-solidity-utils/contracts/helpers/TransientStorageHelpers.sol";

import { RouterWethLib } from "./lib/RouterWethLib.sol";
import { SenderGuard } from "./SenderGuard.sol";
import { VaultGuard } from "./VaultGuard.sol";
```

pkg/vault/contracts/RouterHooks.sol
```diff
@@ -0,0 +1,402 @@
...
+import { IWETH } from "@balancer-labs/v3-interfaces/contracts/solidity-utils/misc/IWETH.sol";
+import { IVault } from "@balancer-labs/v3-interfaces/contracts/vault/IVault.sol";
+import "@balancer-labs/v3-interfaces/contracts/vault/VaultTypes.sol";
+import "@balancer-labs/v3-interfaces/contracts/vault/RouterTypes.sol";
+
+import { RouterWethLib } from "./lib/RouterWethLib.sol";
+import { RouterCommon } from "./RouterCommon.sol";
+
+/**
+ * @notice Base Router with hooks for swaps and liquidity operations via Vault.
```
EndymionJkb, line 20:
```suggestion
 * @notice Base Router with hooks for swaps and liquidity operations via the Vault.
```

pkg/vault/contracts/RouterHooks.sol
```diff
@@ -0,0 +1,402 @@
...
+            params.pool,
+            params.sender,
+            params.tokens,
+            params.exactAmountsIn,
+            params.minBptAmountOut,
+            params.userData
+        );
+
+        for (uint256 i = 0; i < params.tokens.length; ++i) {
+            IERC20 token = params.tokens[i];
```
EndymionJkb, line 60: Could move the token assignment lower so that it's not executed in the "continue" case.

pkg/vault/contracts/RouterHooks.sol
```diff
@@ -0,0 +1,402 @@
...
+                kind: params.kind,
+                userData: params.userData
+            })
+        );
+
+        // maxAmountsIn length is checked against tokens length at the Vault.
+        IERC20[] memory tokens = _vault.getPoolTokens(params.pool);
+
+        for (uint256 i = 0; i < tokens.length; ++i) {
+            IERC20 token = tokens[i];
```
EndymionJkb, line 120: Same as above; could move assignment below "continue" (like it's done in remove liquidity)

pkg/vault/contracts/RouterHooks.sol
```diff
@@ -0,0 +1,402 @@
...
+
+import { RouterWethLib } from "./lib/RouterWethLib.sol";
+import { RouterCommon } from "./RouterCommon.sol";
+
+/**
+ * @notice Base Router with hooks for swaps and liquidity operations via Vault.
+ * @dev Implements hooks for init, add liquidity, remove liquidity, and swaps.
+ */
+contract RouterHooks is RouterCommon {
+    using Address for address payable;
```
EndymionJkb, line 24:
```suggestion
abstract contract RouterHooks is RouterCommon {
```
Could be abstract, and unused "using" directive

EndymionJkb (approved):
Looks reasonable, and sets things up well for the future changes.

I'd suggest putting more detail on the changes in the description for future reference. Something like:

* Moves structs defined in IRouter to a new RouterTypes file (similar to VaultTypes).
* Introduces IRouterQueries as a base contract for IRouter, moving all query definitions to it.
* Moves query and hook implementations from derived contracts to separate RouterQueries and RouterHooks contracts.
* Router now descends from RouterCommon through RouterQueries and RouterHooks as intermediaries, vs. directly as before.

See new diagram:
https://drive.google.com/file/d/1UV_UXPRI3zlHbqS7FPVUW_03uzHPAjqD/view

pkg/vault/contracts/RouterHooks.sol
jubeira: note that we're not doing any changes to the router logic in this PR. This is just copying and re-arranging existing logic to make the review easier.

pkg/vault/contracts/RouterHooks.sol
jubeira: (same, let's not modify the code in this PR)

pkg/vault/contracts/RouterHooks.sol
jubeira: That's fine then!

PR #1426, opened 2025-07-08
Clean boosted pool interface

EndymionJkb:
# Description

Before the introduction of the wrapped flags, the CLR checked for a buffer and automatically wrapped/unwrapped, so we had to return the tokens in proportional boosted pool operations. This code was still in there, even though at this point the caller would need to know the tokens in order to decide how to set the wrap flags.

This does change the interface for some boosted pool operations, but simplifies the code - also making it more symmetrical (proportional vs. unbalanced) - and also saves ~350 bytes.

## Type of change

- [ ] Bug fix 
- [ ] New feature 
- [X] Breaking change 
- [ ] Dependency changes
- [ ] Code refactor / cleanup
- [ ] Other

## Checklist:


joaobrunoah (approved):
LGTM! I thought it could conflict with #1416, but that's not the case.

elshan-eth (approved):
LGTM!

pkg/vault/contracts/CompositeLiquidityRouter.sol
```diff
@@ -535,15 +532,16 @@ contract CompositeLiquidityRouter is ICompositeLiquidityRouter, BatchRouterCommo
...
             if (callParams.isStaticCall == false) {
                 _sendTokenOut(callParams.sender, IERC20(token), actualAmountOut, callParams.wethIsEth);
             }
         }
 
         if (actualAmountOut < minAmountOut) {
-            revert IVaultErrors.AmountOutBelowMin(IERC20(tokenOut), actualAmountOut, minAmountOut);
+            IERC20 tokenOut = needToUnwrap ? underlyingToken : IERC20(token);
+
+            revert IVaultErrors.AmountOutBelowMin(tokenOut, actualAmountOut, minAmountOut);
```
jubeira, line 544: Hm.

I think there's a small bug in `_processTokenInExactOut`, and this asymmetry made me realize about it.
Two things:
- This is fine, but alternatively you could keep `tokenOut` as a local before the first `if (needToUnwrap)`, and assign it accordingly. That way you would save the branch here, which checks whether you need to unwrap for a second time in the function.
- For the event in `_processTokenInExactOut` I think we're always emitting the same token, no matter whether we're actually dealing with the wrapped or the underlying. I think we need to emit `tokenIn`, not just `token` in there (commenting here because it's outside the diff).

jubeira (commented):
Looking good!
Just one comment about the events we're emitting when we hit the limits, otherwise LGTM.

Great job, very clean!

pkg/vault/contracts/CompositeLiquidityRouter.sol
```diff
@@ -474,19 +477,18 @@ contract CompositeLiquidityRouter is ICompositeLiquidityRouter, BatchRouterCommo
...
+                _takeTokenIn(callParams.sender, IERC20(token), actualAmountIn, callParams.wethIsEth);
             }
-
-            actualAmountIn = amountIn;
-            tokenIn = token;
         }
 
         if (actualAmountIn > maxAmountIn) {
-            revert IVaultErrors.AmountInAboveMax(IERC20(token), amountIn, maxAmountIn);
+            IERC20 tokenIn = needToWrap ? underlyingToken : IERC20(token);
```
jubeira, line 489: I would do this the other way around.
`underlyingToken` only gets assigned in one of the `if / else` branches, and we're checking `needToWrap` twice in the function. It's not a huge deal, but it's unnecessary complexity IMO. Also this required moving `underlyingToken` outside of the loop.

Instead, you can declare `tokenIn` outside the loop as it was before this PR (not as a return value, but as a local variable at the same level where `underlyingToken` is right now. `tokenIn` is used in both branches, and assigned accordingly to either `underlyingToken` in the `if` branch, or to `token` in the `else` branch. And finally, here you just use `tokenIn` without checking `needToWrap` anymore, since it's already assigned correctly

pkg/vault/contracts/CompositeLiquidityRouter.sol
jubeira: Same as above. Yeah, it's an extra assignment in the normal path, but now you need a larger scope for underlying token and you need to check `needToUnwrap` twice. IMO the other option is simpler.

jubeira (commented):
I'd suggest the other option. IMO it's a bit more readable, and I don't think that saving one local address assignment in one of the paths (i.e. when no wrap / unwrap is needed) is worth the extra complexity.

pkg/vault/contracts/CompositeLiquidityRouter.sol
```diff
@@ -507,17 +509,19 @@ contract CompositeLiquidityRouter is ICompositeLiquidityRouter, BatchRouterCommo
...
         RouterCallParams memory callParams
-    ) private returns (address tokenOut, uint256 actualAmountOut) {
+    ) private returns (uint256 actualAmountOut) {
+        IERC20 underlyingToken;
+        IERC20 tokenOut;
+
         if (needToUnwrap) {
             IERC4626 wrappedToken = IERC4626(token);
-            IERC20 underlyingToken = IERC20(_vault.getERC4626BufferAsset(wrappedToken));
+            underlyingToken = IERC20(_vault.getERC4626BufferAsset(wrappedToken));
```
jubeira, line 518:
```suggestion
            IERC20 underlyingToken = IERC20(_vault.getERC4626BufferAsset(wrappedToken));
```

PR #1432, opened 2025-07-11
Create Stable Oracle Factory

### LINKED ISSUE #1431: Create Factory for Stable LP Oracle

joaobrunoah:
# Description

This PR creates a factory for the Stable Oracle. Notice that it refactors the Weighted oracle to reuse the same code for both factories.

## Type of change

- [ ] Bug fix 
- [x] New feature 
- [ ] Breaking change 
- [ ] Dependency changes
- [ ] Code refactor / cleanup
- [ ] Other

## Checklist:

## Issue Resolution

Closes #1431


pkg/interfaces/contracts/standalone-utils/ILPOracleFactoryBase.sol
```diff
@@ -0,0 +1,40 @@
...
+pragma solidity ^0.8.24;
+
+import { AggregatorV3Interface } from "@chainlink/contracts/src/v0.8/shared/interfaces/AggregatorV3Interface.sol";
+
+import { IBasePool } from "@balancer-labs/v3-interfaces/contracts/vault/IBasePool.sol";
+
+import { ILPOracleBase } from "./ILPOracleBase.sol";
+
+/**
+ * @title LP Oracle Factory
```
EndymionJkb, line 12:
```suggestion
```
Not wrong, but we haven't used titles elsewhere in the codebase. (Only in a couple of the fee burners.)

pkg/interfaces/contracts/standalone-utils/ILPOracleFactoryBase.sol
```diff
@@ -0,0 +1,40 @@
...
+
+    /**
+     * @notice Gets the oracle for the given pool.
+     * @param pool The address of the pool
+     * @return oracle The address of the oracle for the pool
+     */
+    function getOracle(IBasePool pool) external view returns (ILPOracleBase oracle);
+
+    /**
+     * @notice Checks if the given oracle was created by this factory.
```
EndymionJkb, line 35:
```suggestion
     * @notice Checks whether the given oracle was created by this factory.
```

pkg/standalone-utils/contracts/LPOracleFactoryBase.sol
```diff
@@ -0,0 +1,56 @@
...
+
+import { ILPOracleFactoryBase } from "@balancer-labs/v3-interfaces/contracts/standalone-utils/ILPOracleFactoryBase.sol";
+import { ILPOracleBase } from "@balancer-labs/v3-interfaces/contracts/standalone-utils/ILPOracleBase.sol";
+import { IBasePool } from "@balancer-labs/v3-interfaces/contracts/vault/IBasePool.sol";
+import { IVault } from "@balancer-labs/v3-interfaces/contracts/vault/IVault.sol";
+
+import { InputHelpers } from "@balancer-labs/v3-solidity-utils/contracts/helpers/InputHelpers.sol";
+import { SingletonAuthentication } from "@balancer-labs/v3-vault/contracts/SingletonAuthentication.sol";
+
+abstract contract LPOracleFactoryBase is ILPOracleFactoryBase, SingletonAuthentication {
```
EndymionJkb, line 16: Would be good to add a contract header, saying this is a common base contract for all BPT oracles (e.g., weighted, stable).

pkg/standalone-utils/contracts/LPOracleFactoryBase.sol
```diff
@@ -0,0 +1,56 @@
...
+    uint256 internal _oracleVersion;
+    mapping(IBasePool => ILPOracleBase) internal _oracles;
+    mapping(ILPOracleBase => bool) internal _isOracleFromFactory;
```
EndymionJkb, line 19:
```suggestion

    mapping(IBasePool pool => ILPOracleBase) oracle internal _oracles;
    mapping(ILPOracleBase oracle => bool creationFlag) internal _isOracleFromFactory;
```
Name the mapping parameters (even though they're kind of obvious). Would also separate the mappings from other variables.

pkg/standalone-utils/contracts/LPOracleFactoryBase.sol
```diff
@@ -0,0 +1,56 @@
...
+import { ILPOracleFactoryBase } from "@balancer-labs/v3-interfaces/contracts/standalone-utils/ILPOracleFactoryBase.sol";
+import { ILPOracleBase } from "@balancer-labs/v3-interfaces/contracts/standalone-utils/ILPOracleBase.sol";
+import { IBasePool } from "@balancer-labs/v3-interfaces/contracts/vault/IBasePool.sol";
+import { IVault } from "@balancer-labs/v3-interfaces/contracts/vault/IVault.sol";
+
+import { InputHelpers } from "@balancer-labs/v3-solidity-utils/contracts/helpers/InputHelpers.sol";
+import { SingletonAuthentication } from "@balancer-labs/v3-vault/contracts/SingletonAuthentication.sol";
+
+abstract contract LPOracleFactoryBase is ILPOracleFactoryBase, SingletonAuthentication {
+    uint256 internal _oracleVersion;
```
EndymionJkb, line 17: Is there a specific reason to use a numerical version? Should we just use Version?

pkg/standalone-utils/contracts/LPOracleFactoryBase.sol
```diff
@@ -0,0 +1,56 @@
...
+    mapping(IBasePool => ILPOracleBase) internal _oracles;
+    mapping(ILPOracleBase => bool) internal _isOracleFromFactory;
+
+    constructor(IVault vault, uint256 oracleVersion) SingletonAuthentication(vault) {
```
EndymionJkb, line 21: Is the only reason for `SingletonAuthentication` to support `getVault`? (vs. granting the ability to have permissioned functions in derived oracles - doesn't seem there are any in weighted or stable, so oracles that need that could include it at a lower level).

If so, consider storing it here: `IVault private immutable _vault`, define a public `getVault`, and just use `_vault` internally.

pkg/standalone-utils/contracts/WeightedLPOracleFactory.sol
```diff
@@ -3,57 +3,40 @@
...
+import { LPOracleFactoryBase } from "./LPOracleFactoryBase.sol";
 import { WeightedLPOracle } from "./WeightedLPOracle.sol";
 
-contract WeightedLPOracleFactory is IWeightedLPOracleFactory, SingletonAuthentication {
-    uint256 internal _oracleVersion;
-    mapping(IWeightedPool => IWeightedLPOracle) internal _oracles;
-    mapping(IWeightedLPOracle => bool) internal _isOracleFromFactory;
+contract WeightedLPOracleFactory is LPOracleFactoryBase {
+    /**
+     * @notice New oracle is created for a Weighted Pool.
```
EndymionJkb, line 18:
```suggestion
     * @notice A new Weighted Pool oracle was created.
```

EndymionJkb (commented):
Look straightforward and clean. Just raised a couple questions.

pkg/standalone-utils/contracts/LPOracleFactoryBase.sol
jubeira: +1, let's just shift to `Version`

pkg/standalone-utils/contracts/LPOracleFactoryBase.sol
```diff
@@ -0,0 +1,56 @@
...
+        _oracleVersion = oracleVersion;
+    }
+
+    /// @inheritdoc ILPOracleFactoryBase
+    function create(IBasePool pool, AggregatorV3Interface[] memory feeds) external returns (ILPOracleBase oracle) {
+        if (address(_oracles[pool]) != address(0)) {
```
jubeira, line 27: I'm thinking that this could be an issue.
You might create an oracle with the wrong price feeds for a given pool, and this wouldn't allow you to create another one.

We should hash the feeds and the pool to make a unique internal identifier, and maybe disallow creating an oracle for a given pool with the same configuration twice. But more than one oracle for a given pool could be OK if the configuration is not the same.

pkg/standalone-utils/contracts/LPOracleFactoryBase.sol
jubeira: having `disable` could be a good idea (same as pool factories: do not allow creating more oracles if disabled; existing oracles keep working as usual). Let's add it too.

jubeira (commented):
First pass done!
Just a few minor comments on top of @EndymionJkb 's; otherwise LGTM!

pkg/standalone-utils/contracts/LPOracleFactoryBase.sol
joaobrunoah: So, AggregatorV3Intervace has a `version` function that requires uint256, while our Version library is of type `string`. To be compatible with chainlink's interace, I'd recommend leaving it as is and don't adopting the Version library.

jubeira: fair point, agreed.

pkg/standalone-utils/contracts/LPOracleFactoryBase.sol
joaobrunoah: So, after the offline discussion, moving on with the hash of pool and feeds.

pkg/standalone-utils/contracts/LPOracleFactoryBase.sol
```diff
@@ -0,0 +1,101 @@
...
+
+    constructor(IVault vault, uint256 oracleVersion) SingletonAuthentication(vault) {
+        _oracleVersion = oracleVersion;
+    }
+
+    /// @inheritdoc ILPOracleFactoryBase
+    function create(IBasePool pool, AggregatorV3Interface[] memory feeds) external returns (ILPOracleBase oracle) {
+        _ensureEnabled();
+
+        bytes32 oracleId = _computeOracleId(pool, feeds);
```
jubeira, line 36: Simplified to `oracleId`; we do this elsewhere when keys are not directly addresses.

pkg/standalone-utils/contracts/LPOracleFactoryBase.sol
```diff
@@ -0,0 +1,101 @@
...
+    }
+
+    /// @inheritdoc ILPOracleFactoryBase
+    function getOracle(
+        IBasePool pool,
+        AggregatorV3Interface[] memory feeds
+    ) external view returns (ILPOracleBase oracle) {
+        bytes32 oracleId = _computeOracleId(pool, feeds);
+        oracle = ILPOracleBase(address(_oracles[oracleId]));
+    }
```
jubeira, line 59: Removed revert from here to keep it simple.

pkg/standalone-utils/contracts/LPOracleFactoryBase.sol
```diff
@@ -0,0 +1,101 @@
...
+
+    /// @inheritdoc ILPOracleFactoryBase
+    function isOracleFromFactory(ILPOracleBase oracle) external view returns (bool success) {
+        success = _isOracleFromFactory[oracle];
+    }
+
+    /// @inheritdoc ILPOracleFactoryBase
+    function disable() external authenticate {
+        _ensureEnabled();
```
jubeira, line 68: Same as elsewhere, revert in case we try to disable twice.

pkg/standalone-utils/contracts/LPOracleFactoryBase.sol
```diff
@@ -0,0 +1,101 @@
...
+
+        _isDisabled = true;
+        emit OracleFactoryDisabled();
+    }
+
+    function _computeOracleId(IBasePool pool, AggregatorV3Interface[] memory feeds) internal pure returns (bytes32) {
+        address[] memory feedAddresses = asAddress(feeds);
+        return keccak256(abi.encode(pool, feedAddresses));
```
jubeira, line 76: Simplified this a bit to avoid unnecessary copies.

Also skipped sorting the feeds. You can create an oracle with the fees in the wrong order, and therefore need to create it again. If we sort it here, two oracles with the same feeds but in different order would be considered as exactly the same, whereas that's not the case really. Each feed corresponds to a token, in whatever order they come.

jubeira (approved):
Sending my comments inline to speed things up :) LGTM!

PTAL and feel free to merge when ready.

PR #1449, opened 2025-07-23
Refactoring CompositeLiquidityRouter - Final Part 

elshan-eth:
# Description

Continuation of the refactoring from PR https://github.com/balancer/balancer-v3-monorepo/pull/1447. This PR adds AggregatorCompositeLiquidityRouter.

## Type of change

- [ ] Bug fix 
- [X] New feature 
- [ ] Breaking change 
- [ ] Dependency changes
- [ ] Code refactor / cleanup
- [ ] Optimization: [ ] gas / [ ] bytecode
- [ ] Documentation or wording changes
- [ ] Other

## Checklist:

- [ ] The diff is legible and has no extraneous changes
- [ ] Complex code has been commented, including external interfaces
- [ ] Tests have 100% code coverage
- [ ] The base branch is either `main`, or there's a description of how to merge

PR #1452, opened 2025-07-29
Suggestion: CLR last stage

jubeira:
# Description

Small suggestion on top of #1449.

The diff in #1449 looks correct, and I know that this version is more verbose, but I'm slightly inclined to stick with the original structure. This is mostly because of two reasons:
- It's more symmetrical with `_processTokenOut...` functions
- It clearly separates what we do for the wrapped and the underlying case

Furthermore, the pattern of 

```
if (isStatic == false) {
    if (isAggregator) {
        settle();
    } else {
        takeTokenIn();
    }
}
```
is starting to be used everywhere, so perhaps we can even do a second helper in a future PR (e.g. `takeOrSettle`).

@elshan-eth PTAL. The diff looks a bit messy, but it should be the original code adapted to the `isAggregator` logic.

## Type of change

- [ ] Bug fix 
- [ ] New feature 
- [ ] Breaking change 
- [ ] Dependency changes
- [x] Code refactor / cleanup
- [ ] Other

## Checklist:

## Issue Resolution

N/A

### DIFF
--- a/pkg/vault/contracts/CompositeLiquidityRouterHooks.sol
+++ b/pkg/vault/contracts/CompositeLiquidityRouterHooks.sol
@@ -210,33 +210,43 @@ contract CompositeLiquidityRouterHooks is BatchRouterCommon {
         bool needToWrap,
         RouterCallParams memory callParams
     ) private returns (uint256 actualAmountIn) {
-        bool isAggregator = _isAggregator;
+        if (needToWrap) {
+            IERC4626 wrappedToken = IERC4626(token);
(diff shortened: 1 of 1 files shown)


elshan-eth (approved):
I like my version because it's more compact, but I think your arguments make sense, so LGTM.

jubeira:
I now understood the nuances involved around this bit of code much better. Let's stick to the original version; I've left a few more minor comments in there, but let's go with that one. Thanks for checking this one out and for explaining the details to me!

PR #1463, opened 2025-08-04
Onchain token pair registry (alt requirements)

### LINKED ISSUE #1439: Build onchain token pair registry
We want a registry of (token pair) --> pool address onchain.

The interface should allow adding a pool to the registry, and the registry should parse all the possible pairs (including boosted pools with wrapped / underlying variants). The function to add new pools should be permissioned in principle to avoid spam.

Then, we should have a getter that returns the list of pools that allow swaps through a given pair (with pagination, etc).

jubeira:
# Description

Add a small registry that maps token pairs --> swap path.
A buffer can also be considered a path for (underlying, wrapped) tokens; pools and buffers are treated equally.

Modifying the registry is permissioned (with owner + governance), and one pair might have multiple paths.

This version, unlike #1453, allows setting multi-hop paths. And as such, paths from (tokenA --> tokenB) are different from (tokenB --> tokenA), which requires storing both options individually.

## Type of change

- [ ] Bug fix 
- [x] New feature 
- [ ] Breaking change 
- [ ] Dependency changes
- [ ] Code refactor / cleanup
- [ ] Other

## Checklist:

  - Some coverage missing

## Issue Resolution

Closes #1439

### DIFF
--- a/pkg/standalone-utils/contracts/TokenPairRegistry.sol
+++ b/pkg/standalone-utils/contracts/TokenPairRegistry.sol
@@ -1,6 +1,6 @@
 // SPDX-License-Identifier: GPL-3.0-or-later
 
-pragma solidity ^0.8.27;
+pragma solidity ^0.8.24;
 
 import { IERC4626 } from "@openzeppelin/contracts/interfaces/IERC4626.sol";
 import { IERC20 } from "@openzeppelin/contracts/token/ERC20/IERC20.sol";
@@ -41,7 +41,7 @@ contract TokenPairRegistry is ITokenPairRegistry, OwnableAuthentication {
         return _pairsToPaths[tokenId];
     }
 
-    function addPath(address tokenIn, IBatchRouter.SwapPathStep[] calldata steps) external authenticate {
+    function addPath(address tokenIn, IBatchRouter.SwapPathStep[] memory steps) external authenticate {
         address tokenOut = address(steps[steps.length - 1].tokenOut);
         bytes32 tokenId = _getTokenId(tokenIn, tokenOut);
 
@@ -58,7 +58,11 @@ contract TokenPairRegistry is ITokenPairRegistry, OwnableAuthentication {
             stepTokenIn = address(step.tokenOut);
         }
 
-        _pairsToPaths[tokenId].push(steps);
+        IBatchRouter.SwapPathStep[][] storage paths = _pairsToPaths[tokenId];
+        paths.push();
+        for (uint256 i = 0; i < steps.length; ++i) {
+            paths[paths.length - 1].push(steps[i]);
+        }
         emit PathAdded(tokenIn, tokenOut, _pairsToPaths[tokenId].length);
     }
 
@@ -66,7 +70,7 @@ contract TokenPairRegistry is ITokenPairRegistry, OwnableAuthentication {
         bytes32 tokenId = _getTokenId(tokenIn, tokenOut);
         IBatchRouter.SwapPathStep[][] storage paths = _pairsToPaths[tokenId];
         uint256 pathsLength = paths.length;
-        
+
         if (index >= pathsLength) {
             revert IndexOutOfBounds();
         }
@@ -75,6 +79,8 @@ contract TokenPairRegistry is ITokenPairRegistry, OwnableAuthentication {
             paths[index] = paths[pathsLength - 1];
         }
         paths.pop();
+
+        emit PathRemoved(tokenIn, tokenOut, paths.length);
     }
 
     function addSimplePath(address path) external authenticate {
@@ -103,7 +109,7 @@ contract TokenPairRegistry is ITokenPairRegistry, OwnableAuthentication {
         uint256 tokenPairs = tokens.length - 1;
         for (uint256 i = 0; i < tokenPairs; ++i) {
             for (uint256 j = i + 1; j < tokens.length; ++j) {
-                _addTokenPair(pool, address(tokens[i]), address(tokens[j]));
+                _addTokenPair(pool, address(tokens[i]), address(tokens[j]), false);
             }
         }
     }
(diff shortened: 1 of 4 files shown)


joaobrunoah (approved):
LGTM!

pkg/standalone-utils/contracts/TokenPairRegistry.sol
```diff
@@ -0,0 +1,242 @@
...
+
+        if (index >= pathsLength) {
+            revert IndexOutOfBounds();
+        }
+
+        if (pathsLength > 1 && index != pathsLength - 1) {
+            paths[index] = paths[pathsLength - 1];
+        }
+        // pop() can be used to clear dynamic arrays: it deletes every element and removes the inner array entirely.
+        paths.pop();
```
EndymionJkb, line 82: Do we need to validate that the path still makes sense after the removal, like we do when adding a path? In that case, could factor out the "path validation" code and call it both places.

pkg/standalone-utils/contracts/TokenPairRegistry.sol
```diff
@@ -0,0 +1,242 @@
...
+
+        emit PathRemoved(tokenIn, tokenOut, paths.length);
+    }
+
+    function addSimplePath(address poolOrBuffer) external authenticate {
+        if (vault.isPoolRegistered(poolOrBuffer)) {
```
EndymionJkb, line 88: Do we want to factor this repeated code out into `_getTokenType` (pool or buffer) as in the routers?

pkg/standalone-utils/contracts/TokenPairRegistry.sol
```diff
@@ -0,0 +1,242 @@
...
+                emit PathRemoved(tokenIn, tokenOut, paths.length);
+                return;
+            }
+        }
+
+        // If we didn't return at this point, we revert with an error indicating the invalid removal attempt.
+        revert InvalidRemovePath(poolOrBuffer, tokenIn, tokenOut);
+    }
+
+    function _removeBuffer(IERC4626 wrappedToken) internal {
```
EndymionJkb, line 191: Same comment as above in _addBuffer; always unwrap?

pkg/standalone-utils/contracts/TokenPairRegistry.sol
```diff
@@ -0,0 +1,242 @@
...
+        for (uint256 i = 0; i < tokens.length; ++i) {
+            if (tokens[i] == IERC20(tokenIn)) {
+                foundIn = true;
+            }
+            if (tokens[i] == IERC20(tokenOut)) {
+                foundOut = true;
+            }
+        }
+
+        if (!foundIn || !foundOut) {
```
EndymionJkb, line 238:
```suggestion
        if (foundIn == false || foundOut == false) {
```

pkg/standalone-utils/contracts/TokenPairRegistry.sol
```diff
@@ -0,0 +1,242 @@
...
+            if (tokenOut != address(buffer)) {
+                revert InvalidBufferPath(buffer, tokenIn, tokenOut);
+            }
+        } else if (tokenIn == buffer) {
+            // This is an unwrap
+            if (tokenOut != address(underlying)) {
+                revert InvalidBufferPath(buffer, tokenIn, tokenOut);
+            }
+        } else {
+            // Token in must be either wrapped or underlying
```
EndymionJkb, line 218:
```suggestion
            // Token in must be either wrapped or underlying.
```

pkg/standalone-utils/contracts/TokenPairRegistry.sol
```diff
@@ -0,0 +1,242 @@
...
+
+    function _checkBufferStep(address buffer, address tokenIn, address tokenOut) internal view {
+        address underlying = address(vault.getBufferAsset(IERC4626(buffer)));
+
+        if (underlying == address(0)) {
+            revert BufferNotInitialized(buffer);
+        }
+
+        if (tokenIn == underlying) {
+            // This is a wrap
```
EndymionJkb, line 208:
```suggestion
            // This is a wrap.
```

pkg/standalone-utils/contracts/TokenPairRegistry.sol
```diff
@@ -0,0 +1,242 @@
...
+            if (tokenOut != address(buffer)) {
+                revert InvalidBufferPath(buffer, tokenIn, tokenOut);
+            }
+        } else if (tokenIn == buffer) {
+            // This is an unwrap
```
EndymionJkb, line 213:
```suggestion
            // This is an unwrap.
```

pkg/standalone-utils/contracts/TokenPairRegistry.sol
```diff
@@ -0,0 +1,242 @@
...
+
+import { IBatchRouter } from "@balancer-labs/v3-interfaces/contracts/vault/IBatchRouter.sol";
+import { ITokenPairRegistry } from "@balancer-labs/v3-interfaces/contracts/standalone-utils/ITokenPairRegistry.sol";
+import { IVault } from "@balancer-labs/v3-interfaces/contracts/vault/IVault.sol";
+import { EnumerableSet } from "@balancer-labs/v3-solidity-utils/contracts/openzeppelin/EnumerableSet.sol";
+import { InputHelpers } from "@balancer-labs/v3-solidity-utils/contracts/helpers/InputHelpers.sol";
+
+import { OwnableAuthentication } from "./OwnableAuthentication.sol";
+
+contract TokenPairRegistry is ITokenPairRegistry, OwnableAuthentication {
```
EndymionJkb, line 16: Could use a contract header with the purpose, use cases, and some examples (e.g., addPath vs. addSimplePath).

jubeira: Added. Test cases should be self documenting on how to use this

pkg/standalone-utils/contracts/TokenPairRegistry.sol
```diff
@@ -0,0 +1,242 @@
...
+    using EnumerableSet for EnumerableSet.AddressSet;
+
+    mapping(bytes32 pairId => IBatchRouter.SwapPathStep[][] paths) internal _pairsToPaths;
+
+    constructor(IVault vault, address initialOwner) OwnableAuthentication(vault, initialOwner) {
+        // solhint-disable-previous-line no-empty-blocks
+    }
+
+    function getPathAt(
+        address tokenA,
```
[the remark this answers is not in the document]
jubeira: yeah thanks, that was a leftover from the original specification where paths were unique and there was no input / output, just a pair.

pkg/standalone-utils/contracts/TokenPairRegistry.sol
```diff
@@ -0,0 +1,242 @@
...
+
+        IBatchRouter.SwapPathStep[][] storage paths = _pairsToPaths[tokenId];
+        paths.push();
+        for (uint256 i = 0; i < steps.length; ++i) {
+            paths[paths.length - 1].push(steps[i]);
+        }
+        emit PathAdded(tokenIn, tokenOut, _pairsToPaths[tokenId].length);
+    }
+
+    function removePathAtIndex(address tokenIn, address tokenOut, uint256 index) external authenticate {
```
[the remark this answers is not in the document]
jubeira: (re: headings: the contract is fairly short; I don't think that's really needed in this case).

PR #1465, opened 2025-08-07
Adapt Gyro Library to Surge Hook

joaobrunoah:
# Description

To avoid code duplication, the new Surge Hook needs the offset parameters exposed in the function that computes the swap. It doesn't affect the pool math.

## Type of change

- [ ] Bug fix 
- [ ] New feature 
- [ ] Breaking change 
- [ ] Dependency changes
- [x] Code refactor / cleanup
- [ ] Other

## Checklist:

### DIFF
--- a/pkg/interfaces/contracts/pool-gyro/IGyroECLPPool.sol
+++ b/pkg/interfaces/contracts/pool-gyro/IGyroECLPPool.sol
@@ -81,6 +81,7 @@ interface IGyroECLPPool is IBasePool {
      * @param symbol Pool symbol
      * @param eclpParams Parameters to configure the E-CLP pool, with 18 decimals
      * @param derivedEclpParams Parameters calculated off-chain based on eclpParams. 38 decimals for higher precision
+     * @param version Pool version
      */
     struct GyroECLPPoolParams {
         string name;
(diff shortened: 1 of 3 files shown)


pkg/interfaces/contracts/pool-gyro/IGyroECLPPool.sol
```diff
@@ -153,4 +154,11 @@ interface IGyroECLPPool is IBasePool {
      * @return data A struct containing all immutable stable pool parameters
      */
     function getGyroECLPPoolImmutableData() external view returns (GyroECLPPoolImmutableData memory data);
+
+    /**
+     * @notice Get ECLP parameters relevant to swap/add/remove calculations.
+     * @return params A struct containing all ECLP parameters
+     * @return d A struct containing all derived ECLP parameters
+     */
+    function getECLPParams() external view returns (EclpParams memory params, DerivedEclpParams memory d);
```
joaobrunoah, line 163: This function already existed, but it wasn't in the interface.

pkg/pool-gyro/contracts/GyroECLPPool.sol
```diff
@@ -139,17 +139,25 @@ contract GyroECLPPool is IGyroECLPPool, BalancerPoolToken, PoolInfo, Version {
...
-                    .toUint256();
+            (newBalanceInt, , ) = GyroECLPMath.calcYGivenX(
+                balancesLiveScaled18[0].toInt256(),
+                eclpParams,
+                derivedECLPParams,
+                invariant
+            );
         }
+
+        newBalance = newBalanceInt.toUint256();
```
jubeira, line 160: I think I'd prefer `return newBalanceInt.toUint256();` given that this function is a bit long.

For the other cases where we now return `a` and `b` I think implicit returns are acceptable.

jubeira (approved):
This looks correct!

I'm not a huge fan of functions returning multiple values, but in this case it looks justified. Also the functions that now return `a` and `b` are not that long, so it should be acceptable.

EndymionJkb (approved):
LGTM!

PR #1473, opened 2025-08-08
Constant Chainlink price feed

EndymionJkb:
# Description

Useful for LP oracles where the rate provider already reflects the market price (e.g., XAUt).

All LP oracle pool tokens need a price feed, but if you had both the rate provider and price feed they would interfere, so you can only use one. If there is a rate provider available, but no feed, the oracle could be deployed with this one.

## Type of change

- [ ] Bug fix 
- [X] New feature 
- [ ] Breaking change 
- [ ] Dependency changes
- [ ] Code refactor / cleanup
- [ ] Other

## Checklist:


joaobrunoah (approved):
LGTM! I wonder if we could use price = 1 if the feed address is equal to 0 in the LPBaseOracle, it'd save some gas by not interacting with an external contract.

Anyway, using a constant price oracle looks more organized, not sure which path we will choose.

jubeira (approved):
:ship:

PR #1481, opened 2025-08-13
docs: correct broken link

radik878:
corrected broken link https://docs.gyro.finance/gyroscope-protocol/concentrated-liquidity-pools/2-clps to
https://docs.gyro.finance/pools/2-clps.html


jubeira (approved):
Thanks @radik878 !

PR #1500, opened 2025-08-19
Update Foundry.toml to 0.8.27

joaobrunoah:
# Description

When Gyro pools were created in the monorepo, we decided to support `require` with custom errors, which was only supported by 0.8.27.

So, hooks and oracles using gyro pool also needed to be compiled using 0.8.27, which made us compile the whole `standalone-utils` and `pool-hooks` packages as 0.8.27 (and dependencies). In short, everything in the monorepo is implicitly compiled with 0.8.27 version but the Vault, which is explicitly specified to compile with 0.8.26 solc in hardhat.config.ts.

This PR makes this situation more explicit by declaring all packages compilers as 0.8.27.

## Type of change

- [ ] Bug fix 
- [ ] New feature 
- [ ] Breaking change 
- [x] Dependency changes
- [ ] Code refactor / cleanup
- [ ] Other

## Checklist:

### DIFF
--- a/pkg/vault/foundry.toml
+++ b/pkg/vault/foundry.toml
@@ -23,7 +23,7 @@ remappings = [
 ]
 optimizer = true
 optimizer_runs = 999
-solc_version = '0.8.26'
+solc_version = '0.8.27'
 auto_detect_solc = false
 evm_version = 'cancun'
 ignored_error_codes = [2394, 5574, 3860] # Transient storage, code size


jubeira (approved):
LGTM!

It should be fine to change the `foundry.toml` file in the vault package as well; hardhat will compile it in 0.8.26 which is the version used for the release, but the test will compile with 0.8.27. The test reuses hardhat artifacts anyway so it should be consistent.

EndymionJkb (approved):
LGTM

PR #1508, opened 2025-08-21
Document feed/token correspondence in LP oracles

### LINKED ISSUE #1506: [L-05 - Oracle audit] Missing order checks can lead to faulty oracle
When an Oracle is deployed, an array of AggregatorV3Interface is given to correspond to the
tokens in the tokenPool. We can see in the constructor that there is a check to ensure that the
length or the feeds array is the same as the totalTokens.

However, there is no check to ensure that the AggregatorV3 feed and the token actually
correspond.

Note: there's really not much we can do here; a registry sounds overkill and oracles can be created again with the right order. Let's review the documentation and make sure this is crystal clear.

EndymionJkb:
# Description

Per audit issue L-05, comment the importance of properly aligning feeds and tokens, and the inherent limitations on what the contract can verify. Refer to the public docs, which also elaborate on this.

## Type of change

- [ ] Bug fix 
- [ ] New feature 
- [ ] Breaking change 
- [ ] Dependency changes
- [ ] Code refactor / cleanup
- [ ] Other

## Checklist:

## Issue Resolution

Resolves #1506

### DIFF
--- a/pkg/interfaces/contracts/standalone-utils/ILPOracleFactoryBase.sol
+++ b/pkg/interfaces/contracts/standalone-utils/ILPOracleFactoryBase.sol
@@ -36,6 +36,10 @@ interface ILPOracleFactoryBase {
 
     /**
      * @notice Creates a new oracle for the given pool.
+     * @dev Note that the caller must ensure that the given `feeds` are correct for the corresponding pool tokens.
+     * The contract checks that the array lengths match, but cannot independently verify correctness. See the docs for
+     * notes on how to pair feeds (and rate providers) with tokens, as there are many subtleties.
+     *
      * @param pool The address of the pool
      * @param feeds The array of price feeds for the tokens in the pool
      * @return oracle The address of the newly created oracle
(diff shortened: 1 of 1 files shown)


pkg/interfaces/contracts/standalone-utils/ILPOracleFactoryBase.sol
```diff
@@ -36,6 +36,10 @@ interface ILPOracleFactoryBase {
 
     /**
      * @notice Creates a new oracle for the given pool.
+     * @dev Note that the caller must ensure that the given `feeds` are correct for the corresponding pool tokens.
+     * The contract checks that the array lengths match, but cannot independently verify correctness. See the docs for
+     * notes on how to pair feeds (and rate providers) with tokens, as there are many subtleties.
```
jubeira, line 41: Should we also clarify that oracles can be recreated with the right feed / token order in case there's a mistake?

jubeira (approved):
LGTM!

pkg/interfaces/contracts/standalone-utils/ILPOracleFactoryBase.sol
EndymionJkb: Sure, I can add that.

PR #1510, opened 2025-08-22
[DRAFT] Merge AggregatorBatchRouterHooks with BatchRouterHooks

elshan-eth:
# Description

## Type of change

- [ ] Bug fix 
- [ ] New feature 
- [ ] Breaking change 
- [ ] Dependency changes
- [ ] Code refactor / cleanup
- [ ] Other

## Checklist:


pkg/vault/contracts/BatchRouterHooks.sol
```diff
@@ -23,13 +23,11 @@ import {
...
-
 /**
- * @notice Base BatchRouter contract with hooks for swaps and liquidity operations via the Vault.
+ * @notice Entrypoint for batch swaps, and batch swap queries.
+ * @dev The external API functions unlock the Vault, which calls back into the corresponding hook functions.
+ * These interpret the steps and paths in the input data, perform token accounting (in transient storage, to save gas),
+ * settle with the Vault, and handle wrapping and unwrapping ETH.
  */
 contract BatchRouterHooks is BatchRouterCommon {
     using CastingHelpers for *;
```
EndymionJkb, line 33:
```suggestion
```
Unused. Can also remove CastingHelpers import above.

pkg/vault/contracts/BatchRouterHooks.sol
```diff
@@ -23,13 +23,11 @@ import {
...
-}
-
 /**
- * @notice Base BatchRouter contract with hooks for swaps and liquidity operations via the Vault.
+ * @notice Entrypoint for batch swaps, and batch swap queries.
+ * @dev The external API functions unlock the Vault, which calls back into the corresponding hook functions.
+ * These interpret the steps and paths in the input data, perform token accounting (in transient storage, to save gas),
+ * settle with the Vault, and handle wrapping and unwrapping ETH.
  */
 contract BatchRouterHooks is BatchRouterCommon {
```
EndymionJkb, line 32:
```suggestion
abstract contract BatchRouterHooks is BatchRouterCommon {
```
Should be abstract

EndymionJkb (dismissed):
Looks great! The diagram now looks like [this](https://drive.google.com/file/d/17nucg-fSm_5rzutk1ukdFay-6kmYnpZF/view?usp=sharing)

This is outside the diff, but we can eliminate the double inheritance of IRouterQueries, as revealed by the diagram (which is why I make these diagrams). Just remove "is IRouterQueries" from the IRouter interface.

I looked at whether we might consolidate RouterQueries and RouterHooks (like on the batch side), but maybe that's not worth it, unless we're sure we don't want to allow separate inheritance of IRouterQueries (would be eliminated if we combined them).

jubeira (commented):
Halfway there; checking swap exact out.

pkg/vault/contracts/BatchRouterHooks.sol
```diff
@@ -550,8 +588,8 @@ contract BatchRouterHooks is BatchRouterCommon {
 
         (uint256[] memory exactAmountsOut, ) = _getSingleInputArrayAndTokenIndex(pool, tokenOut, stepExactAmountOut);
 
-        // The Router is always an intermediary in this case. The Vault will burn tokens from the Router, so
-        // the Router is both owner and spender, which doesn't require approval.
+        // Router is always an intermediary in this case. The Vault will burn tokens from the Router, so
+        // Router is both owner and spender (which doesn't need approval).
```
joaobrunoah, line 592: I never know when we use the article and when we don't, but I had the impression that "The Router" here was correct (especially looking at other comments that always start with "The user, The last step, The buffer", etc.)

joaobrunoah (approved):
LGTM! Nice that only minor changes were needed.

PR #1519, opened 2025-08-25
E-CLP Surge Hook: Rotation Angle Guardrail

### LINKED ISSUE #1516: Surge ECLP: Document parameter boundaries, revert if invalid
Revert on registration when the peak price approximation is not accurate (current code should just work for relevant use cases; no need to adjust the logic itself).
Document boundaries appropriately.

joaobrunoah:
# Description

E-CLP Surge Hook relies on the fact that the peak price can be approximated by sine/cosine of the rotation angle. That's valid when the peak price is close to 1, but when the peak is not 1 (and there's no rate provider) there's an imprecision.

We evaluated that between 30 and 60 degrees of rotation, the approximation works fine. Also, we've noticed that most of E-CLP pools are created in this range. The ones that are not have a rate provider, and then the rotation angle becomes 45 degrees.

So, this PR makes sure the rotation angle of an E-CLP pool registered in the E-CLP Surge Hook is in the range [30deg, 60deg].

## Type of change

- [ ] Bug fix 
- [x] New feature 
- [ ] Breaking change 
- [ ] Dependency changes
- [ ] Code refactor / cleanup
- [ ] Other

## Checklist:

## Issue Resolution

Closes #1516


pkg/pool-hooks/contracts/ECLPSurgeHook.sol
```diff
@@ -25,6 +25,13 @@ contract ECLPSurgeHook is SurgeHookCommon {
     using FixedPoint for uint256;
     using SafeCast for *;
 
+    /**
+     * @notice The rotation angle is too small or too large for the surge hook to be used.
+     * @dev The surge hook accept angles from 30 to 60 degrees. Outside of this range, the computation of the peak
```
EndymionJkb, line 30:
```suggestion
     * @dev The surge hook accepts angles from 30 to 60 degrees. Outside of this range, the computation of the peak
```

pkg/pool-hooks/contracts/ECLPSurgeHook.sol
```diff
@@ -25,6 +25,13 @@ contract ECLPSurgeHook is SurgeHookCommon {
...
+     * price cannot be approximated by sine/cosine.
+     */
+    error InvalidRotationAngleForSurgeHook();
```
EndymionJkb, line 33:
```suggestion
    error InvalidRotationAngle();
```
Maybe we don't need "ForSurgeHook," as we're in that contract.

pkg/pool-hooks/contracts/ECLPSurgeHook.sol
```diff
@@ -53,6 +60,14 @@ contract ECLPSurgeHook is SurgeHookCommon {
         TokenConfig[] memory tokenConfig,
         LiquidityManagement calldata liquidityManagement
     ) public override onlyVault returns (bool success) {
+        (IGyroECLPPool.EclpParams memory eclpParams, ) = IGyroECLPPool(pool).getECLPParams();
+
+        // The surge hook only works for pools with a rotation angle between 30 and 60 degrees. Outside of this range,
+        // the computation of the peak price cannot be approximated by sine/cosine.
```
EndymionJkb, line 66: I guess it's kind of obvious from context, but we could note that sin 30 degrees = cos 60 degrees = 0.5; sine monotonically increases and cosine monotonically decreases (from 0-90); so they're only *both* below 0.5 between 30 and 60. Above 60, sine will be > 0.5, and below 30, cosine will be > 0.5.

EndymionJkb (approved):
The code change is small and looks correct. Minor suggestions.

jubeira (approved):
I've checked onchain and all of the pools that are most relevant by TVL have the same angle and would meet this condition. It seems a sensible guardrail.

LGTM!

Issue #1515, opened 2025-08-25
Move oracles to independent packages


jubeira:
`standalone-utils` --> `pool-oracles` inside monorepo. Same code, same settings, etc.

PR #1521, opened 2025-08-26
Oracle uptime

### LINKED ISSUE #1505: [L-04 - Oracles audit] Missing sequencer uptime check for L2 deployment
Balancer is deployed on multiple L2s, which means that the oracles will also be used on those
chains. However, the getFeedData is missing a check for sequencer uptime check:

```
function getFeedData() public view returns (int256[] memory prices, uint256 updatedAt) {
    uint256 totalTokens = _totalTokens;
    AggregatorV3Interface[] memory feeds = _getFeeds(totalTokens);
    uint256[] memory feedDecimalScalingFactors =
    _getFeedTokenDecimalScalingFactors(totalTokens);
    prices = new int256[](totalTokens);
    updatedAt = type(uint256).max;
    for (uint256 i = 0; i < totalTokens; i++) {
        (, int256 answer,, uint256 feedUpdatedAt,) = feeds[i].latestRoundData();
```

`LPOracleBase.latestRoundData()` calls external Chainlink price feeds but does not verify that the L2 sequencer is operational.
On L2 rollups, Chainlink recommends integrating the Sequencer Uptime Feed to detect:
●​ When the sequencer is offline (status = 1)
●​ The initial “grace period” after recovery, when prices can still be stale or inconsistent due to delayed state sync

Without this, the oracle may return incorrect LP prices during sequencer downtime or
immediately after recovery, potentially causing issues.

Recommendation: It is recommended to implement a sequencer uptime check in the
LPOracleBase contract for L2 deployments by adding a function to verify the sequencer status.

EndymionJkb:
# Description

Optimistic rollups (e.g., Arbitrum, Optimism) and many ZK-rollups rely on sequencers to efficiently manage transaction ordering, execution, and batching before submitting them to Layer 1. If this sequencer goes down, L2 applications will not be updated in a timely fashion, which is particularly important for price oracles. To mitigate this risk, Chainlink provides a sequencer uptime feed to track the status of the sequencer, including the time since the last restart (see https://docs.chain.link/data-feeds/l2-sequencer-feeds).

This PR adds a sequencer uptime feed address (and associated "grace period") to the factory deployment, which get stored in immutable storage, and passed down to immutable storage in the oracles as new ones are created.

The grace period is the minimum time the sequencer uptime feed must have been operating for the oracles to be reliable. This could vary by chain. (They suggest one hour.) We'll want to research good values for this, as all oracles would effectively stop working for that long after a sequencer outage.

Every time the oracle price feeds are queried, we first query the sequencer to make sure it's running. If it is down, or has been up for less than the grace period, we revert. If the sequencer feed itself is unavailable and can't report its status, it likewise reverts with an explicit error message. (This might be overkill.)

I took the names of the events and errors from the Chainlink docs, but I don't think they're great names. The "grace period" is more like a "minimum resync period" or something like that. And we don't say "NotOver" anywhere else. Maybe NotExpired? Is it best to stick with the Chainlink-suggested names?

## Type of change

- [ ] Bug fix 
- [X] New feature 
- [ ] Breaking change 
- [ ] Dependency changes
- [ ] Code refactor / cleanup
- [ ] Other

## Checklist:

## Issue Resolution

Resolves #1505


joaobrunoah (approved):
LGTM! Very clear. This will impacgt #1515 , maybe we should merge this one first and I can adjust that one.

pkg/standalone-utils/contracts/LPOracleBase.sol
```diff
@@ -276,6 +307,30 @@ abstract contract LPOracleBase is ILPOracleBase, AggregatorV3Interface {
...
+            if (answer == _SEQUENCER_STATUS_DOWN) {
+                revert SequencerDown();
+            }
+
+            // solhint-disable-next-line not-rely-on-time
+            if (block.timestamp - startedAt < _uptimeGracePeriod) {
+                revert GracePeriodNotOver();
+            }
+        } catch {
+            revert SequencerFeedUnavailable();
```
jubeira, line 330: Is there any known reason why these contracts would revert, or is this just defensive programming?

pkg/standalone-utils/test/foundry/WeightedLPOracle.t.sol
```diff
@@ -247,6 +257,47 @@ contract WeightedLPOracleTest is BaseVaultTest, WeightedPoolContractsDeployer {
...
+    }
+
+    function testGracePeriodNotOver() public {
+        (IWeightedPool pool, ) = createAndInitPool(2);
+        (WeightedLPOracleMock oracle, ) = deployOracle(pool);
+        uptimeFeed.setStartedAt(block.timestamp - 100);
+
+        vm.expectRevert(ISequencerUptimeFeed.GracePeriodNotOver.selector);
+        oracle.latestRoundData();
+    }
```
jubeira, line 299: WDYT about moving these tests to a base layer?

`createAndInit` would be overridden anyway, so each specific test would run its own variant properly. Do you think it's worth it?

pkg/standalone-utils/test/gas/.hardhat-snapshots/WeightedLPOracle - 2 tokens
```diff
@@ -1,3 +1,3 @@
-Min: 92.8k
-Max: 93.2k
-Avg: 93.0k
\ No newline at end of file
+Min: 101.6k
```
jubeira, line 1: were these outdated?

We're using `address(0)` as the sequencer uptime interface in the gas tests, so the overhead in this case should be minimal. Can you explain this difference?

jubeira (commented):
Just a few questions, otherwise LGTM! Great job!

jubeira:
> I took the names of the events and errors from the Chainlink docs, but I don't think they're great names. The "grace period" is more like a "minimum resync period" or something like that. And we don't say "NotOver" anywhere else. Maybe NotExpired? Is it best to stick with the Chainlink-suggested names?

I trust your judgement here :) I don't think we strictly need to stick to chainlink names; as long as the errors are clear I'm happy with it.

pkg/standalone-utils/test/gas/.hardhat-snapshots/WeightedLPOracle - 2 tokens
jubeira: (just to clarify: I don't see anything obviously wrong with this PR, but the gas shouldn't increase by that much, which makes me think I might be missing something if this diff effectively corresponds to this change)

EndymionJkb: They might just be outdated. I will check. I agree it shouldn't change much. Maybe a little, to do the extra check, but it should immediately return.

jubeira: were these outdated in the end?

jubeira (approved):
LGTM! Code and test changes seem reasonable. Great job!

I'm requesting one more review from @joaobrunoah since the code changed quite a bit since his review.

EndymionJkb:
> LGTM! Code and test changes seem reasonable. Great job!
> 
> I'm requesting one more review from @joaobrunoah since the code changed quite a bit since his review.

As I said above, I'm still working on it (took a break to do some other reviews).

Hopefully will have it all by tonight, for final review tomorrow.

pkg/interfaces/contracts/oracles/ISequencerUptimeFeed.sol
```diff
@@ -0,0 +1,39 @@
...
+ * the time since the last restart (see https://docs.chain.link/data-feeds/l2-sequencer-feeds).
+ */
+interface ISequencerUptimeFeed {
+    /// @notice The uptime sequencer has returned a status of "down".
+    error SequencerDown();
+
+    /**
+     * @notice A price feed was accessed while still within the resync window (e.g., after a sequencer outage).
+     * @dev Since outages result in a queue of delayed transactions which must be processed, the stabilization period
+     * (or resync window) allows sufficient time for the L2 network to "catch up," so that the price feed is accurate.
```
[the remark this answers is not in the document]
EndymionJkb: This is a US/British thing. I'm following the US standard (e.g., also not sticking random u's everywhere :)

pkg/oracles/contracts/StableLPOracle.sol
```diff
@@ -33,8 +33,10 @@ contract StableLPOracle is LPOracleBase {
         IVault vault_,
         IStablePool pool_,
         AggregatorV3Interface[] memory feeds,
+        AggregatorV3Interface sequencerUptimeFeed,
+        uint256 uptimeGracePeriod,
```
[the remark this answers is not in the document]
EndymionJkb: I thought I'd caught all of those! Looks like I still missed some. Thanks!

PR #1530, opened 2025-08-29
Add price guardrails to stable oracle

### LINKED ISSUE #1504: [L-03 - Oracles audit] StableOracle math breaks on extreme price drops
StableOracle uses complex math to calculate the TVL. If a price feed reports extremely low prices, these math operations become dysfunctional.
-​ When price < 1e8, the calculations will revert due to division by zero.
-​ When 1e8 < price < 1e12, the calculations will revert due to K non-convergence.

However, in such a case the actual stablePool is likely also dysfunctional due to the extreme
value difference in tokens within the pool.

Recommendations:
It is possible to include a price check for every feed which enforces that the price is higher than
1e12 or it will revert with “Oracle Failure”.

Alternatively, if the gas cost for this is deemed too high, it could also be a good option to clearly
document the cases in which the StableOracle is expected to revert

Note: let's add boundaries and fail gracefully

joaobrunoah:
# Description

To compute the BPT price of a stable pool, the stable oracle relies on a K factor which is computed through a newton method. To ensure the convergence of this function, the prices can't differ too much.

Our fuzz tests have shown that if the price ratio between the max price and min price is above 1e7, K does not converge in certain scenarios. This PR reverts the TVL computation for prices that do not comply with this guardrail.

## Type of change

- [x] Bug fix 
- [ ] New feature 
- [ ] Breaking change 
- [ ] Dependency changes
- [ ] Code refactor / cleanup
- [ ] Other

## Checklist:

## Issue Resolution

Closes #1504


pkg/oracles/test/foundry/StableLPOracle.t.sol
```diff
@@ -218,13 +267,16 @@ contract StableLPOracleTest is BaseLPOracleTest, StablePoolContractsDeployer {
...
                 poolInitAmounts[i] =
                     bound(poolInitAmountsRaw[i], FixedPoint.ONE, 1e9 * FixedPoint.ONE) /
                     (10 ** (18 - tokenDecimals));
-                prices[i] = bound(pricesRaw[i], MIN_PRICE, MAX_PRICE) / (10 ** (18 - tokenDecimals));
+                prices[i] =
+                    bound(pricesRaw[i], priceBase, (priceBase * PRICE_RATIO_LIMIT).mulDown(0.99e18)) /
+                    (10 ** (18 - tokenDecimals));
                 updateTimestamps[i] = block.timestamp - bound(updateTimestampsRaw[i], 1, 100);
 
                 if (updateTimestamps[i] < minUpdateTimestamp) {
```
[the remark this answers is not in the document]
joaobrunoah: I had this test in place in this commit: [be07643e3c6863186ccde43ac5f2868d1fe8082d](https://github.com/balancer/balancer-v3-monorepo/commit/be07643e3c6863186ccde43ac5f2868d1fe8082d)
Maybe I merged something and erased it.

EndymionJkb (approved):
LGTM! All suggestions implemented. As long as you think my "flaky test" fix is good.

pkg/oracles/contracts/StableLPOracle.sol
```diff
@@ -238,4 +252,46 @@ contract StableLPOracle is LPOracleBase {
...
+
+        // The `k` parameter has a high risk of not converging if the ratio between the maximum and minimum prices is
+        // too high.
+        if (maxPrice / minPrice > _PRICE_RATIO_LIMIT) {
+            revert PriceRatioTooHigh();
+        }
+
+        normalizedPrices = new int256[](_totalTokens);
+        for (uint256 i = 0; i < _totalTokens; i++) {
+            normalizedPrices[i] = i == minPriceIndex ? int256(FixedPoint.ONE) : _divDownInt(prices[i], minPrice);
```
jubeira, line 294: `_divDownInt(prices[i], minPrice);` should be exactly equal to `FixedPoint.ONE` for `i = minPriceIndex` anyway right? I don't think there's rounding error there (please double check and correct me if I'm wrong; we can even do a fuzz test if needed).

pkg/oracles/contracts/StableLPOracle.sol
```diff
@@ -42,13 +57,12 @@ contract StableLPOracle is LPOracleBase {
 
     /// @inheritdoc LPOracleBase
     function _computeTVL(int256[] memory prices) internal view override returns (uint256 tvl) {
```
jubeira, line 59: This is outside of the diff, but we should document in `LPOracleBase` that `prices` is a FP-18 (min price would mean different things if `prices` wasn't normalized to 18 decimals).

pkg/oracles/contracts/StableLPOracle.sol
jubeira: And perhaps with that you don't even need to save and update the index.

pkg/oracles/contracts/test/StableLPOracleMock.sol
```diff
@@ -27,8 +27,29 @@ contract StableLPOracleMock is StableLPOracle {
...
+        int256 minPrice = prices[0];
+        for (uint256 i = 1; i < _totalTokens; i++) {
+            if (prices[i] < minPrice) {
+                minPrice = prices[i];
+            }
+        }
+
+        normalizedPrices = new int256[](_totalTokens);
+        for (uint256 i = 0; i < _totalTokens; i++) {
+            normalizedPrices[i] = _divDownInt(prices[i], minPrice);
```
jubeira, line 50: Here we do the same as in the actual contract but we don't fix the min price index to `FP1` :)

pkg/oracles/test/foundry/StableLPOracle.t.sol
```diff
@@ -180,13 +186,18 @@ contract StableLPOracleTest is BaseLPOracleTest, StablePoolContractsDeployer {
...
+            uint256 priceBase = bound(pricesRaw[0], MIN_PRICE, MAX_PRICE / (PRICE_RATIO_LIMIT / 2));
+
             for (uint256 i = 0; i < totalTokens; i++) {
                 _tokens[i] = address(sortedTokens[i]);
                 uint256 tokenDecimals = IERC20Metadata(address(sortedTokens[i])).decimals();
                 poolInitAmounts[i] =
                     bound(poolInitAmountsRaw[i], FixedPoint.ONE, 1e9 * FixedPoint.ONE) /
                     (10 ** (18 - tokenDecimals));
-                prices[i] = bound(pricesRaw[i], MIN_PRICE, MAX_PRICE) / (10 ** (18 - tokenDecimals));
+                prices[i] =
```
jubeira, line 197: Let's leave a small comment here explaining this bound.

pkg/oracles/test/foundry/StableLPOracle.t.sol
```diff
@@ -262,6 +320,55 @@ contract StableLPOracleTest is BaseLPOracleTest, StablePoolContractsDeployer {
...
+
+        uint256[] memory poolInitAmounts = [uint256(563_360_193e18), uint256(10_000_000e18)].toMemoryArray();
+        address[] memory _tokens = [address(sortedTokens[0]), address(sortedTokens[1])].toMemoryArray();
+
+        pool = createAndInitPool(_tokens, poolInitAmounts, amplificationParameter);
+        (oracle, ) = deployOracle(pool);
+
+        int256[] memory pricesInt = new int256[](2);
+        pricesInt[0] = int256(55_529_888e18);
+        pricesInt[1] = int256(1e9);
```
jubeira, line 366: Can we add a small comment explaining these numbers?

jubeira (approved):
Just a few minor comments; otherwise LGTM! All things considered, I think this guardrail has a good cost / benefit relationship. 

Great job!

pkg/oracles/contracts/StableLPOracle.sol
joaobrunoah: Yes, the divDown division would be `number * 1e18 / number`, which is 1e18 and has no rounding issues, no matter which number is used.

This is in place to save gas, as suggested by @EndymionJkb here: https://github.com/balancer/balancer-v3-monorepo/pull/1530#discussion_r2314640468 . But, maybe the cost of storing the index is greater than the division operation.

joaobrunoah: Actually, @EndymionJkb optimization saves 200 gas.

PR #1534, opened 2025-09-02
Timelock authorizer V3: port from V2

jubeira:
# Description

There are a few incompatibilities baked in the timelock authorizer code that won't allow us to use it as is for both V2 and V3, mostly related to the authorizer adaptor (and vault references).

I believe the best way to deal with this is to just port it over to this repository, removing the bits that we won't use for V3. The changes will be fairly simple.

To make this reviewable by humans, I'll be targeting the branch `timelock-authorizer-v3-dev` which is based off the current `main` branch. This PR just brings the code from V2 as is and is not expected to build or do anything.
We don't care about that for now; the point here is to double check that we're not introducing unwanted mistakes along the way. The code will build in a follow-up PR with smaller diff. In other words, to review just check that the files are exactly the same as in V2.

All these PRs shall be merged to the dev branch in order to keep CI healthy on main, and then we'll merge the branch to main once we're done.

## Type of change

- [ ] Bug fix 
- [x] New feature 
- [ ] Breaking change 
- [ ] Dependency changes
- [ ] Code refactor / cleanup
- [ ] Other

## Checklist:

## Issue Resolution

N/A

### DIFF
--- a/pkg/governance-scripts/contracts/TimelockAuthorizerMigrator.sol
+++ b/pkg/governance-scripts/contracts/TimelockAuthorizerMigrator.sol
@@ -1,0 +1,164 @@
+// SPDX-License-Identifier: GPL-3.0-or-later
+// This program is free software: you can redistribute it and/or modify
+// it under the terms of the GNU General Public License as published by
+// the Free Software Foundation, either version 3 of the License, or
+// (at your option) any later version.
+
+// This program is distributed in the hope that it will be useful,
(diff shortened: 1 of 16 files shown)


EndymionJkb (approved):
I confirm this is identical to v2, except AuthorizerWithAdaptorValidation.test.ts is missing, which is expected, as that's the thing v3 doesn't need.

It's going to fail build because of the solidity version, but that is addressed in the next PR.

PR #1542, opened 2025-09-05
Calculation of the exact swap amount on-chain in AddUnbalancedLiquidityViaSwapRouter

elshan-eth:
# Description

This PR introduces a different calculation method for the swap amount in AddUnbalancedLiquidityViaSwapRouter. Instead of being calculated off-chain, the swap amount is now determined after addLiquidity.

## Type of change

- [ ] Bug fix 
- [X] New feature 
- [ ] Breaking change 
- [ ] Dependency changes
- [ ] Code refactor / cleanup
- [ ] Other

## Checklist:


pkg/vault/contracts/AddUnbalancedLiquidityViaSwapRouter.sol
```diff
@@ -31,149 +31,162 @@ contract AddUnbalancedLiquidityViaSwapRouter is RouterQueries, IAddUnbalancedLiq
...
-        (amountsIn, , ) = _queryAddLiquidityHook(params.addLiquidityParams);
-        swapAmountCalculated = _querySwapHook(params.swapParams);
+    ) external nonReentrant onlyVault returns (uint256[] memory amountsIn) {
+        (, amountsIn, ) = _computeAddUnbalancedLiquidityViaSwap(params);
     }
 
-    /***************************************************************************
-                                   Helper functions
-    ***************************************************************************/
+    function _computeAddUnbalancedLiquidityViaSwap(
```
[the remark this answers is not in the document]
elshan-eth: Thanks!

pkg/vault/contracts/AddUnbalancedLiquidityViaSwapRouter.sol
```diff
@@ -98,13 +98,14 @@ contract AddUnbalancedLiquidityViaSwapRouter is RouterQueries, IAddUnbalancedLiq
...
-            uint256 amountIn = amountsIn[i];
-            uint256 amountOut = amountsOut[i];
 
-            _takeTokenIn(hookParams.sender, tokens[i], amountIn, hookParams.wethIsEth);
-            _sendTokenOut(hookParams.sender, tokens[i], amountOut, hookParams.wethIsEth);
+        // In this case, the pools have two tokens,
+        // but we use a 'for' loop instead of getting the needed indexes because it's simpler.
+        for (uint256 i = 0; i < tokens.length; ++i) {
+            _takeTokenIn(hookParams.sender, tokens[i], amountsIn[i], hookParams.wethIsEth);
+            _sendTokenOut(hookParams.sender, tokens[i], amountsOut[i], hookParams.wethIsEth);
```
jubeira, line 106: There's something weird about this conceptually (please let me know if I'm missing something).

The net result of this operation should only have tokens in:
- add proportional
- swap excess for exact (or vice versa depending on the outcome)

Either way, tokens should flow from the LP to the vault, and not vice versa. Why do we need to call `_sendTokenOut`? I'd seem to be unnecessary.

pkg/vault/contracts/AddUnbalancedLiquidityViaSwapRouter.sol
```diff
@@ -138,75 +139,69 @@ contract AddUnbalancedLiquidityViaSwapRouter is RouterQueries, IAddUnbalancedLiq
     function _computeAddUnbalancedLiquidityViaSwap(
         AddLiquidityAndSwapHookParams calldata hookParams
     ) private returns (IERC20[] memory tokens, uint256[] memory amountsIn, uint256[] memory amountsOut) {
+        // maxAmountsIn length is checked against tokens length at the Vault.
+        tokens = _vault.getPoolTokens(hookParams.pool);
+        amountsOut = new uint256[](tokens.length);
+        uint256[] memory maxAmountsIn = new uint256[](tokens.length);
+
+        uint256 exactTokenIndex = hookParams.operationParams.exactToken == tokens[0] ? 0 : 1;
+        uint256 adjustableTokenIndex = exactTokenIndex == 0 ? 1 : 0;
```
jubeira, line 148: nit: you can set these two in the same line

```suggestion
(uint256 exactTokenIndex, uint256 adjustableTokenIndex) = hookParams.operationParams.exactToken == tokens[0] ? (0, 1) : (1, 0);
```

pkg/vault/contracts/AddUnbalancedLiquidityViaSwapRouter.sol
```diff
@@ -138,75 +139,69 @@ contract AddUnbalancedLiquidityViaSwapRouter is RouterQueries, IAddUnbalancedLiq
...
+
+        maxAmountsIn[exactTokenIndex] = hookParams.operationParams.maxAdjustableAmount;
+        maxAmountsIn[adjustableTokenIndex] = hookParams.operationParams.maxAdjustableAmount;
```
jubeira, line 151: As discussed off-thread, I _think_ we should just use _MAX_AMOUNT as the limit for this first operation, and check limits later.

pkg/vault/contracts/AddUnbalancedLiquidityViaSwapRouter.sol
```diff
@@ -138,75 +139,69 @@ contract AddUnbalancedLiquidityViaSwapRouter is RouterQueries, IAddUnbalancedLiq
...
-            if (amountsIn[exactInTokenIndex] >= swapAmountOut) {
-                amountsIn[exactInTokenIndex] -= swapAmountOut;
+            amountsIn[adjustableTokenIndex] += swapAmountIn;
+            if (amountsIn[exactTokenIndex] >= swapAmountOut) {
+                amountsIn[exactTokenIndex] -= swapAmountOut;
             } else {
-                amountsIn[exactInTokenIndex] = 0;
-                amountsOut[exactInTokenIndex] = swapAmountOut - amountsIn[exactInTokenIndex];
+                amountsIn[exactTokenIndex] = 0;
+                amountsOut[exactTokenIndex] = swapAmountOut - amountsIn[exactTokenIndex];
```
jubeira, line 183: I still don't get why we need the `else` clause here. See https://github.com/balancer/balancer-v3-monorepo/pull/1542#discussion_r2369686361 for reference.

There should be no amounts out in this function. 
The swap just adds some amount of the input token to the amounts in from the `add` step, and reduces the amount in for the token out. But the net result for both operations should be > 0 for both tokens (otherwise this is not an add; it's a swap).

I guess technically you can add a lot from one token, get BPT and receive the excess of the other token, but that'd be super weird. Conceptually we are just sending tokens to the vault in exchange for BPT, not tokens in exchange for other tokens. In principle I don't see how we can enter the `else` clause, but if that was possible I'd just revert.

pkg/vault/contracts/AddUnbalancedLiquidityViaSwapRouter.sol
```diff
@@ -31,149 +31,183 @@ contract AddUnbalancedLiquidityViaSwapRouter is RouterQueries, IAddUnbalancedLiq
...
+
+            amountsIn[exactInTokenIndex] += swapAmountIn;
+            if (amountsIn[exactMaxInTokenIndex] >= swapAmountOut) {
+                amountsIn[exactMaxInTokenIndex] -= swapAmountOut;
+            } else {
+                amountsOut[exactMaxInTokenIndex] = swapAmountOut - amountsIn[exactMaxInTokenIndex];
+                amountsIn[exactMaxInTokenIndex] = 0;
+            }
+        }
     }
```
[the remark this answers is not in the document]
jubeira: This should be covered by applying the limit in the swap, although it wouldn't be a terrible idea to enforce it again more explicitly.

jubeira (commented):
Second pass done.

I think we shouldn't be supporting tokens going from the vault to the user in this case, since the operation is add liquidity. My understanding is that this condition should not happen. Am I missing something?

Other than that it's looking good, only minor comments left.

pkg/vault/contracts/AddUnbalancedLiquidityViaSwapRouter.sol
```diff
@@ -96,14 +100,12 @@ contract AddUnbalancedLiquidityViaSwapRouter is RouterQueries, IAddUnbalancedLiq
...
-        uint256[] memory amountsOut;
-        (tokens, amountsIn, amountsOut) = _computeAddUnbalancedLiquidityViaSwap(hookParams);
+        (tokens, amountsIn) = _computeAddUnbalancedLiquidityViaSwap(hookParams);
 
         // In this case, the pools have two tokens,
         // but we use a 'for' loop instead of getting the needed indexes because it's simpler.
         for (uint256 i = 0; i < tokens.length; ++i) {
             _takeTokenIn(hookParams.sender, tokens[i], amountsIn[i], hookParams.wethIsEth);
-            _sendTokenOut(hookParams.sender, tokens[i], amountsOut[i], hookParams.wethIsEth);
         }
```
jubeira, line 109:
```
_takeTokenIn(hookParams.sender, tokens[0], amountsIn[0], hookParams.wethIsEth);
_takeTokenIn(hookParams.sender, tokens[1], amountsIn[1], hookParams.wethIsEth);
```

:D

pkg/vault/contracts/AddUnbalancedLiquidityViaSwapRouter.sol
```diff
@@ -138,28 +140,25 @@ contract AddUnbalancedLiquidityViaSwapRouter is RouterQueries, IAddUnbalancedLiq
...
+        (uint256 exactTokenIndex, uint256 adjustableTokenIndex) = hookParams.operationParams.exactToken == tokens[0]
+            ? (0, 1)
+            : (1, 0);
 
         (amountsIn, , ) = _vault.addLiquidity(
             AddLiquidityParams({
                 pool: hookParams.pool,
                 to: hookParams.sender,
-                maxAmountsIn: maxAmountsIn,
+                maxAmountsIn: [_MAX_AMOUNT, _MAX_AMOUNT].toMemoryArray(),
```
jubeira, line 154: hm, this is a test function.
Perhaps it's ok to use in prod, but we should move it out of the `test` directory if so.

Otherwise just create an array and populate it manually.

pkg/vault/contracts/AddUnbalancedLiquidityViaSwapRouter.sol
```diff
@@ -170,39 +169,39 @@ contract AddUnbalancedLiquidityViaSwapRouter is RouterQueries, IAddUnbalancedLiq
...
+            amountsIn[adjustableTokenIndex] -= swapAmountOut;
+        }
+
+        if (amountsIn[exactTokenIndex] != hookParams.operationParams.exactAmount) {
+            revert AmountInDoesNotMatchExact(amountsIn[exactTokenIndex], hookParams.operationParams.exactAmount);
+        } else if (amountsIn[adjustableTokenIndex] > hookParams.operationParams.maxAdjustableAmount) {
+            revert AmountInAboveMaxAdjustableAmount(
+                amountsIn[adjustableTokenIndex],
+                hookParams.operationParams.maxAdjustableAmount
+            );
```
jubeira, line 204: There's no way to trigger these in practice because the swap will fail with the swap limit before reaching here right?

How about not using the limit in the swap? These error messages are actually better in the context of an `addLiquidity` operation, so maybe it's better to let the swap through and fail here (even though it'll consume some more gas in the failing case). And with that we could actually test these conditions.

WDYT?

jubeira (approved):
Just a few minor comments, otherwise LGTM! Great job! 💯

PR #1546, opened 2025-09-08
Pin error-ex version, etc.

EndymionJkb:
# Description

Even though we're not front-end, I think we should do this just in case (See [this vulnerability](https://x.com/p3b7_/status/1965094840959410230?s=46&t=__ll0ZskxlJ8qG4DuuMK6A)). Sounds like it installs a background process, etc. Don't want to download anything malicious to local machines while running tests.

Had to downgrade a couple of the packages further to not break things. Pinned them to what we were already using, far below the bad versions.

Apparently the snaps were out of date also.

## Type of change

- [ ] Bug fix 
- [ ] New feature 
- [ ] Breaking change 
- [ ] Dependency changes
- [ ] Code refactor / cleanup
- [X] Other

## Checklist:


joaobrunoah (approved):
Makes sense

PR #1555, opened 2025-09-24
CLR audit fixes #2

EndymionJkb:
# Description

The audit agent seemed to have 3 false positives (a couple just wrong or nonsensical; e.g., claiming queries wouldn't work), 3 documented unsupported/accepted cases, and two real ones. The first (wrong argument to an error) I was *sure* we'd already found and fixed. Maybe it crept back in somehow with all the merging.

The second I remember thinking about at one point, but it got lost in the shuffle. It does seem to be valid, if an edge case. Handling it does increase gas a bit (hence all the snap updates), but not terribly so.

Also added some tests for coverage, and removed a redundant check (getting back a little of the gas increase).

## Type of change

- [X] Bug fix 
- [ ] New feature 
- [ ] Breaking change 
- [ ] Dependency changes
- [ ] Code refactor / cleanup
- [ ] Other

## Checklist:


pkg/vault/contracts/BatchRouterCommon.sol
```diff
@@ -117,13 +117,17 @@ abstract contract BatchRouterCommon is RouterCommon {
             // Erases delta, in case more than one batch router operation is called in the same transaction.
             _currentSwapTokenInAmounts().tSet(tokenIn, 0);
             _currentSwapTokensIn().remove(tokenIn);
+            _settledTokenAmounts().tSet(tokenIn, 0);
```
jubeira, line 120: This code is shared by the batch router and the CLR.
This is done at a different level for the batch router though, so if we set it to 0 here unconditionally we'll be setting it twice.

I'm not entirely against that, but maybe we should zero them out after using it (and `_settledTokenAmounts` is not used in this function).

pkg/vault/contracts/BatchRouterCommon.sol
```diff
@@ -117,13 +117,17 @@ abstract contract BatchRouterCommon is RouterCommon {
...
         }
 
         for (int256 i = int256(numTokensOut - 1); i >= 0; --i) {
             address tokenOut = _currentSwapTokensOut().unchecked_at(uint256(i));
+
             _sendTokenOut(sender, IERC20(tokenOut), _currentSwapTokenOutAmounts().tGet(tokenOut), wethIsEth);
+
             // Erases delta, in case more than one batch router operation is called in the same transaction.
             _currentSwapTokenOutAmounts().tSet(tokenOut, 0);
+            _settledTokenAmounts().tSet(tokenOut, 0);
```
jubeira, line 130: (same as above)

pkg/vault/contracts/CompositeLiquidityRouterHooks.sol
```diff
@@ -310,7 +310,7 @@ abstract contract CompositeLiquidityRouterHooks is BatchRouterCommon {
         }
 
         if (actualAmountIn > maxAmountIn) {
-            revert IVaultErrors.AmountInAboveMax(settlementToken, amountIn, maxAmountIn);
+            revert IVaultErrors.AmountInAboveMax(settlementToken, actualAmountIn, maxAmountIn);
```
jubeira, line 313: right; I think we didn't catch it because it's actually impossible to get here in the wrong case (`erc4626BufferWrapOrUnwrap` should revert first).

pkg/vault/contracts/RouterHooks.sol
```diff
@@ -324,12 +324,6 @@ abstract contract RouterHooks is RouterCommon {
     }
 
     function _swapSingleTokenHook(SwapSingleTokenHookParams calldata params) internal returns (uint256) {
-        // The deadline is timestamp-based: it should not be relied upon for sub-minute accuracy.
-        // solhint-disable-next-line not-rely-on-time
-        if (block.timestamp > params.deadline) {
-            revert SwapDeadline();
-        }
-
```
jubeira, line 332: ah, nice.

jubeira (commented):
Thanks for addressing this @EndymionJkb!

I think we should do something a bit different about `_settledTokenAmounts`. Please check out what we're doing in the batch router for reference.

pkg/vault/contracts/BatchRouterCommon.sol
EndymionJkb: Ok; trying to set directly in the CLR. Will be more code, but should only affect the CLR.

jubeira (commented):
comments discussed off-thread; just a few more things and otherwise LGTM!

pkg/vault/contracts/BatchRouterCommon.sol
```diff
@@ -121,7 +121,9 @@ abstract contract BatchRouterCommon is RouterCommon {
 
         for (int256 i = int256(numTokensOut - 1); i >= 0; --i) {
             address tokenOut = _currentSwapTokensOut().unchecked_at(uint256(i));
+
```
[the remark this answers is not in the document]
EndymionJkb: Ok; I think it looks better with the newlines, but I guess we can not gratuitously change whitespace. There was originally a code change here that got (mostly) reverted.

pkg/vault/contracts/CompositeLiquidityRouterHooks.sol
```diff
@@ -414,6 +414,11 @@ abstract contract CompositeLiquidityRouterHooks is BatchRouterCommon {
     ) external nonReentrant onlyVault returns (uint256 exactBptAmountOut) {
         InputHelpers.ensureInputLengthMatch(params.maxAmountsIn.length, tokensIn.length);
 
+        address[] memory processedTokens = _processedTokensIn().values();
+        for (uint256 i = 0; i < processedTokens.length; ++i) {
+            _processedTokensIn().remove(processedTokens[i]);
+        }
```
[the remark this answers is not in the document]
EndymionJkb: Ahhh! Forgot again.

pkg/vault/contracts/CompositeLiquidityRouterHooks.sol
```diff
@@ -653,11 +653,11 @@ abstract contract CompositeLiquidityRouterHooks is BatchRouterCommon {
                 revert IVaultErrors.InvalidTokenType();
             }
 
-            if (swapAmountIn > 0 && childTokenSettledAmount == 0) {
+            if (swapAmountIn > 0 && _processedTokensIn().contains(childPoolToken) == false) {
                 childPoolNeedsLiquidity = true;
 
                 childPoolAmountsIn[i] = swapAmountIn;
-                _settledTokenAmounts().tSet(childPoolToken, swapAmountIn);
+                _processedTokensIn().add(childPoolToken);
```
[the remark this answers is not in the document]
EndymionJkb: Per off-chain discussion, since this would break all the tests, we can leave this alone for now, then do a follow-on where it reverts in unsupported cases - and also changes all the tests so that we don't try to do unsupported things.

PR #1564, opened 2025-10-02
Explicitly reject duplicate nested tokens

EndymionJkb:
# Description

In this hopefully final iteration of the CLR, we revert instead of silently allowing adding tokens that are duplicated at different levels in a nested pool (e.g., in both the parent and a child pool, or in multiple child pools).

This change broke lots of tests, since the main test parent pool had a duplicate token! The regular parent pool now has USDT instead of DAI (which is duplicated in a parent pool).

(As a bonus, this also fixes a latent bug in the Aggregate Router test, which worked accidentally due to random token sorting.)

## Type of change

- [ ] Bug fix 
- [ ] New feature 
- [ ] Breaking change 
- [ ] Dependency changes
- [X] Code refactor / cleanup
- [ ] Other

## Checklist:

### DIFF
--- a/pkg/vault/contracts/CompositeLiquidityRouterHooks.sol
+++ b/pkg/vault/contracts/CompositeLiquidityRouterHooks.sol
@@ -36,7 +36,11 @@ abstract contract CompositeLiquidityRouterHooks is BatchRouterCommon {
         ERC4626
     }
 
-    // Used to keep track of input tokens that have already been processed in nested pools.
+    // Used to keep track of input tokens that have already been processed in nested pools. This is distinct from
+    // `_currentSwapTokensIn`, which keeps track of the input tokens in a batch swap (e.g., preventing
+    // [USDC, DAI, USDC] inputs). `_processedTokensIn` guarantees that the unique tokens in [USDC, DAI] are not
+    // duplicated during nested pool traversal (e.g., parent and child pools both contain DAI).
+    //
     // solhint-disable-next-line var-name-mixedcase
     bytes32 private immutable _PROCESSED_TOKENS_IN_SLOT = _calculateBatchRouterStorageSlot("processedTokensIn");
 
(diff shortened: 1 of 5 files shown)


pkg/vault/test/foundry/CompositeLiquidityRouterNestedPools.t.sol
```diff
@@ -1467,11 +1515,9 @@ contract CompositeLiquidityRouterNestedPoolsTest is BaseERC4626BufferTest {
         );
 
         // Check ChildPoolB.
-        // Since DAI amountOut comes from parentPool and childPoolB, we need to calculate the proportion that comes
-        // from childPoolB.
         assertApproxEqAbs(
             vars.childPoolBAfter.dai,
-            vars.childPoolBBefore.dai - (amountsOut[vars.daiIdx] - poolInitAmount.mulDown(proportionToRemove)),
+            vars.childPoolBBefore.dai - amountsOut[vars.daiIdx],
```
jubeira, line 1520: This was redundant right?

pkg/vault/contracts/CompositeLiquidityRouterHooks.sol
```diff
@@ -36,7 +36,11 @@ abstract contract CompositeLiquidityRouterHooks is BatchRouterCommon {
         ERC4626
     }
 
-    // Used to keep track of input tokens that have already been processed in nested pools.
+    // Used to keep track of input tokens that have already been processed in nested pools. This is distinct from
+    // `_currentSwapTokensIn`, which keeps track of the input tokens in a batch swap (e.g., preventing
```
jubeira, line 40: hmm maybe we can say in a batch swap / add liquidity operation?
This is actually the CLR and there's no batch swap in here, so that might sound a bit confusing.

pkg/vault/test/foundry/CompositeLiquidityRouterNestedPools.t.sol
EndymionJkb: We used to have DAI in there twice, so all the DAI calculations were "doubled". Now tokens are only there once, so the calculations are simpler.

PR #1573, opened 2025-10-28
Generalize LBP

EndymionJkb:
# Description

We're about to be creating different variants of LBPs, so to avoid duplication this factors out the common parts that should be universal: e.g., 2-token, project and reserve, trusted router, migration ability, etc.

The existing LBPool and factory then only contain data and functions specific to that Weighted Pool implementation (e.g., the weights, dynamic and immutable data).

Follow-ons will add new LBPs with different properties, reusing all the common elements. All tests pass - there should be no semantic changes. (This adds a few tests to improve coverage.)

## Type of change

- [ ] Bug fix 
- [ ] New feature 
- [ ] Breaking change 
- [ ] Dependency changes
- [X] Code refactor / cleanup
- [ ] Other

## Checklist:

### DIFF
--- a/pkg/pool-weighted/contracts/lbp/LBPValidation.sol
+++ b/pkg/pool-weighted/contracts/lbp/LBPValidation.sol
@@ -54,9 +54,10 @@ library LBPValidation {
      * Note that the time is also validated here, and the startTime might be "accelerated" to the current time, if it
      * is in the past, per the logic in `GradualValueChange`.
      *
-     * @param lbpCommonParams The common LBP parameters to validate (may mutate startTime)
+     * @param lbpCommonParams The common LBP parameters to validate
+     * @return resolvedStartTime The final start time (might be "fast forwarded" if the start time is in the past)
      */
-    function validateCommonParams(LBPCommonParams memory lbpCommonParams) internal view {
+    function validateCommonParams(LBPCommonParams memory lbpCommonParams) internal view returns (uint256) {
         if (lbpCommonParams.owner == address(0)) {
             revert InvalidOwner();
         }
@@ -73,10 +74,7 @@ library LBPValidation {
             revert TokensMustBeDifferent();
         }
 
-        lbpCommonParams.startTime = GradualValueChange.resolveStartTime(
-            lbpCommonParams.startTime,
-            lbpCommonParams.endTime
-        );
+        return GradualValueChange.resolveStartTime(lbpCommonParams.startTime, lbpCommonParams.endTime);
     }
 
     /**
--- a/pkg/pool-weighted/contracts/lbp/LBPoolFactory.sol
+++ b/pkg/pool-weighted/contracts/lbp/LBPoolFactory.sol
@@ -103,7 +103,7 @@ contract LBPoolFactory is BaseLBPFactory, BasePoolFactory {
         // These validations are duplicated in the pool contract but performed here to surface precise error messages,
         // as create2 would otherwise mask the underlying revert reason.
 
-        LBPValidation.validateCommonParams(lbpCommonParams);
+        lbpCommonParams.startTime = LBPValidation.validateCommonParams(lbpCommonParams);
 
         LBPoolLib.verifyWeightUpdateParameters(
             lbpParams.projectTokenStartWeight,


pkg/pool-weighted/contracts/lbp/LBPool.sol
```diff
@@ -462,11 +247,31 @@ contract LBPool is ILBPool, WeightedPool, Ownable2Step, BaseHooks {
...
+    // Validates weights and adjusts startTime if needed.
+    function _buildLBPCommonParams(
+        LBPCommonParams memory lbpCommonParams,
+        LBPParams memory lbpParams
+    ) private view returns (LBPCommonParams memory finalCommonParams) {
+        finalCommonParams = lbpCommonParams;
+
+        // Checks that the weights are valid and `endTime` is after `startTime`. If `startTime` is in the past,
+        // avoid abrupt weight changes by overriding it with the current block time.
+        finalCommonParams.startTime = LBPoolLib.verifyWeightUpdateParameters(
```
jubeira: perhaps you can validate the start time at the common layer, and the weights inside `_buildWeightedPoolParams`?

jubeira: Rethinking this a bit, and all validations.

We're validating the migration at the factory, and mixing weights and start time here.

We should validate _everything_ at the factory level and either
- Leave comments where the variables are assigned
- Use a library and validate again to be extra careful
 
If there's a problem in the configuration and the pool creation reverts, it's incredibly annoying to debug because it will show as a create2 error everywhere, whereas if the factory reverts it's easy to understand what went wrong. If we validate a second time in the pool constructor we're extra safe, and that should never happen anyway if we do things right. WDYT?

pkg/pool-weighted/contracts/lbp/LBPool.sol
```diff
@@ -164,40 +96,14 @@ contract LBPool is ILBPool, WeightedPool, Ownable2Step, BaseHooks {
...
-    }
-
-    /**
-     * @notice Returns the trusted router, which is used to initialize and seed the pool.
-     * @return trustedRouter Address of the trusted router (i.e., one that reliably reports the sender)
-     */
-    function getTrustedRouter() external view returns (address) {
-        return _trustedRouter;
-    }
+        emit WeightedLBPoolCreated(lbpCommonParams.owner, lbpCommonParams.blockProjectTokenSwapsIn, hasMigration);
```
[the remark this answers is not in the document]
EndymionJkb: Yes; updated.

pkg/pool-weighted/contracts/lbp/LBPValidation.sol
```diff
@@ -0,0 +1,137 @@
...
+
+    /**
+     * @notice Validates common LBP parameters.
+     * @dev This should be called by both factories for early validation, and pools for direct deployment protection.
+     * Note that the time is also validated here, and the startTime might be "accelerated" to the current time, if it
+     * is in the past, per the logic in `GradualValueChange`.
+     *
+     * @param lbpCommonParams The common LBP parameters to validate (may mutate startTime)
+     */
+    function validateCommonParams(LBPCommonParams memory lbpCommonParams) internal view {
```
jubeira: nit (style): let's return the modified `lbpCommonParams` and assign it to itself in the calling function.

Definition:
```
function validateCommonParams(LBPCommonParams memory lbpCommonParams) internal view returns (LBPCommonParams memory);
```

Usage:
```
lbpCommonParams = LBPValidation.validateCommonParams(lbpCommonParams);
```

Alternatively, just return the new start time without assigning it inside the function:

Definition:
```
function validateCommonParams(LBPCommonParams memory lbpCommonParams) internal view returns (uint256 adjustedStartTime);
```

Modification:
```
        return GradualValueChange.resolveStartTime(
            lbpCommonParams.startTime,
            lbpCommonParams.endTime
        );
```

Usage:
```
lbpCommonParams.startTime = LBPValidation.validateCommonParams(lbpCommonParams);
```

Maybe the latter option makes more sense in this case since you're just modifying one param.

Bottom line: if the function alters the memory argument it receives, we should find a way to make it explicit where the function is being called. Since solidity doesn't have a native way of expressing it, consider one of these two options. Preserving memory arguments untouched is preferred whenever possible.

jubeira (approved):
LGTM, looks pretty good! Great job @EndymionJkb 

I only have one comment regarding the code style (please note that the function is used a few times, so please adjust the definition and the usage in all instances before merging). PTAL :)

PR #1574, opened 2025-10-28
Unify prepaid router ux

### LINKED ISSUE #1572: Prepaid batch router: make remove liquidity consistent
Right now the batch router requires the user to transfer tokens to the vault if it's a swap / add, and to the router if it's a remove operation.
This is efficient but confusing. 

In any case, the batch router requires BPT to be held by the router in the first step to call `vault.removeLiquidity` (in the `permit2` version it does indeed pull the tokens).

We can unify the UX being a bit less efficient by:
- sending BPT to the vault (same as with any other token)
- settle + pull tokens to router --> this would be the extra step
- carry on

jubeira:
# Description

See #1572.
With this change, prepaid batch router works the same for any input token, including BPTs (i.e. pay to vault, call operation).

## Type of change

- [ ] Bug fix 
- [ ] New feature 
- [x] Breaking change 
- [ ] Dependency changes
- [ ] Code refactor / cleanup
- [ ] Other

## Checklist:

## Issue Resolution

Closes #1572

### DIFF
--- a/pkg/vault/contracts/BatchRouterHooks.sol
+++ b/pkg/vault/contracts/BatchRouterHooks.sol
@@ -581,7 +581,6 @@ abstract contract BatchRouterHooks is BatchRouterCommon {
                 // Get credit and pull the tokens.
                 _vault.settle(stepTokenIn, stepMaxAmountIn);
                 _vault.sendTo(stepTokenIn, address(this), stepMaxAmountIn);
-
             } else {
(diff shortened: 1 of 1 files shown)


EndymionJkb (approved):
I like it! Always great to remove special-casing. And as we know, gas optimizations are the biggest source of subtle bugs (not here, but would have made it easier for callers to do it wrong).

Simplifies tests too. Maybe odd that it seems to also save gas?

PR #1577, opened 2025-11-02
add swap fee minimizer and factory with tests

gerrrg:
# Description
This PR adds a new contract, `SwapFeeMinimizer` (with accompanying factory), that allows the `owner` to swap out one specific `outputToken` with as low of a swap fee as possible, regardless of the nominal swap fee of the pool. All other operations with the pool are unchanged, and anyone (`owner` included) can swap with the nominal swap fees in the pool via a typical router. There are a few potential use cases for this functionality, though one notable use case is the ability for a DAO to perform token buybacks on a primary liquidity pool w/o needing to pay the typical swap fee.

## Details
- A new `SwapFeeMinimizer` is deployed atomically alongside a new `WeightedPool`
- Swaps through a given `SwapFeeMinimizer` are constrained to its specific `pool` and `outputToken`
- When an `owner` performs a swap, the contract stores the pool's current swap fee, then pool swap fees are set to `_minimalSwapFee`, the swap occurs, and then the fee is returned to its original level
- Interfaces for swaps are identical to those of the router
- A `SwapFeeManager` is registered as the pool's `PoolRoleAccounts.swapFeeManager`
- The `SwapFeeManager` contract has a passthrough function for `vault.setStaticSwapFeePercentage(...)`, though there is an argument to be made for removing this
- `SwapFeeMinimizer` is `Ownable2Step`

## Architectural Context
The initial plan was for this contract itself to _be_ a router. Changing the architecture from an "is a" context to a "has a" context had significant improvements for simplicity and security, but they come with a few small concessions. With this setup, the `SwapFeeMinimizer` itself has to transfer tokens from the `owner` and be an intermediary with the router, which introduces some allowance handling and minor token accounting. It would have been nice to be able to inherit small parts of the broader router contract, but short of fundamentally rearchitecting the core Balancer router, this was impractical, and the realistic options here came down to:
1. inherit the entire router and expand the scope and bytecode of this contract
2. copy and paste the external swap functions and internal helpers I needed
3. have this contract itself call the router's swap functions and handle token transfers from the `owner`

I went with Option 3 to keep the contract as small as practical, offering simplicity and minimized attack surface. The gas for a fee-minimized swap is a little higher, most notably due to the increased token transfers.

## Type of change

- [ ] Bug fix 
- [x] New feature 
- [ ] Breaking change 
- [ ] Dependency changes
- [ ] Code refactor / cleanup
- [ ] Other

## Checklist:

## Issue Resolution
cc: grants committee @mendesfabio

### DIFF
--- a/pkg/standalone-utils/contracts/SwapFeeMinimizer/SwapFeeMinimizer.sol
+++ b/pkg/standalone-utils/contracts/SwapFeeMinimizer/SwapFeeMinimizer.sol
@@ -33,9 +33,10 @@ contract SwapFeeMinimizer is Ownable2Step, ReentrancyGuardTransient {
     IERC20 private immutable _outputToken;
     uint256 private immutable _minimalSwapFee;
 
-    error InvalidOutputToken(IERC20 expected, IERC20 actual);
     error FeeValidationFailed();
+    error InvalidOutputToken(IERC20 expected, IERC20 actual);
     error InvalidPool();
+    error RouterCallFailed();
 
     modifier withSwapMinimized() {
         uint256 originalFee = _vault.getStaticSwapFeePercentage(_pool);
@@ -108,6 +109,11 @@ contract SwapFeeMinimizer is Ownable2Step, ReentrancyGuardTransient {
         _permit2.approve(address(tokenIn), address(_router), uint160(amount), uint48(deadline));
     }
 
+    /**
+     * @notice Swap exact amount of input token for minimum amount of output token with minimized fees
+     * @dev External function encodes router calldata and calls internal function to avoid stack-too-deep errors.
+     *      The internal function contains all security modifiers and business logic.
+     */
     function swapSingleTokenExactIn(
         address poolAddress,
         IERC20 tokenIn,
@@ -117,10 +123,9 @@ contract SwapFeeMinimizer is Ownable2Step, ReentrancyGuardTransient {
         uint256 deadline,
         bool wethIsEth,
         bytes calldata userData
-    ) external payable onlyOwner onlyOutputTokenFromPool(tokenOut, poolAddress) withSwapMinimized returns (uint256) {
-        _pullTokensFromOwnerAndApprove(tokenIn, exactAmountIn, deadline);
-
-        uint256 amountOut = _router.swapSingleTokenExactIn{ value: msg.value }(
+    ) external payable returns (uint256) {
+        bytes memory routerCalldata = abi.encodeWithSelector(
+            IRouter.swapSingleTokenExactIn.selector,
             poolAddress,
             tokenIn,
             tokenOut,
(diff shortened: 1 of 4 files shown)


pkg/standalone-utils/contracts/SwapFeeMinimizer/SwapFeeMinimizer.sol
```diff
@@ -0,0 +1,257 @@
...
+        internal
+        onlyOwner
+        onlyOutputTokenFromPool(tokenOut, poolAddress)
+        withSwapMinimized
+        nonReentrant
+        returns (uint256)
+    {
+        _pullTokensFromOwnerAndApprove(tokenIn, exactAmountIn, deadline);
+
+        (bool success, bytes memory result) = address(_router).call{ value: msg.value }(routerCalldata);
```
Im-Vishwjeet, line 159: The contract doesn't have a `receive()` or a `fallback() payable`, so any router path that sends ETH back to it reverts.

The router refunds with `Address.sendValue`, which is `call{value: amount}("")` with empty calldata. That bypasses `swapSingleTokenExactIn`, since `payable` there only kicks in when the selector matches. The call tries `receive` / `fallback` instead, finds nothing, and reverts. The minimizer catches that as `success = false` and surfaces it as `RouterCallFailed`.

Two cases where this fires:

1. `wethIsEth = true` and `tokenOut == WETH`. `_sendTokenOut` ([RouterCommon.sol#L291-292](https://github.com/balancer/balancer-v3-monorepo/blob/main/pkg/vault/contracts/RouterCommon.sol#L291)) calls `unwrapWethAndTransferToSender`, which ends in `sendValue(sender, amountOut)` ([RouterWethLib.sol#L38](https://github.com/balancer/balancer-v3-monorepo/blob/main/pkg/vault/contracts/lib/RouterWethLib.sol#L38)). Always reverts.

2. `tokenIn == WETH` with any ETH left in the router after the swap. `_swapSingleTokenHook` calls `_returnEth(sender)` ([RouterHooks.sol#L347-L350](https://github.com/balancer/balancer-v3-monorepo/blob/main/pkg/vault/contracts/RouterHooks.sol#L347)) which ends in `sendValue(excess)` ([RouterCommon.sol#L253](https://github.com/balancer/balancer-v3-monorepo/blob/main/pkg/vault/contracts/RouterCommon.sol#L253)). Hits on every `exactOut` where `amountIn < maxAmountIn` (the normal case, since the caller sets `msg.value = maxAmountIn`), and any `exactIn` where `msg.value > exactAmountIn`.

None of the new tests pass `wethIsEth = true`, so this isn't caught.

Fix is either adding `receive() external payable {}` and refunding residual ETH to `msg.sender` after the router call returns (still inside `nonReentrant`, before the fee restore in `withSwapMinimized`), or just dropping `payable` and `wethIsEth` from the swap functions and hardcoding `false` to the router. The second is cleaner if ETH support isn't really intended, but the PR description points at DAO buybacks where ETH output is the obvious shape, so option 1 probably matches intent.

pkg/standalone-utils/contracts/SwapFeeMinimizer/SwapFeeMinimizer.sol
```diff
@@ -0,0 +1,257 @@
...
+    address private immutable _pool;
+    IERC20 private immutable _outputToken;
+    uint256 private immutable _minimalSwapFee;
+
+    error FeeValidationFailed();
+    error InvalidOutputToken(IERC20 expected, IERC20 actual);
+    error InvalidPool();
+    error RouterCallFailed();
+
+    modifier withSwapMinimized() {
```
Im-Vishwjeet, line 41: The modifier sets the pool fee to `_minimalSwapFee` whether or not the current fee is already below that. `_minimalSwapFee` is fixed at construction, but the static fee is mutable via `setSwapFeePercentage`, so the two can end up out of order.

If the owner ever sets the fee below `_minimalSwapFee` (for example during a fee promotion), the next swap through the minimizer reads the lower fee as `originalFee`, raises it to `_minimalSwapFee` for the duration of the swap, then restores. The owner pays more than they would by calling the router directly at the same fee.

The tests don't surface this because they set `_minimalSwapFee == MIN_SWAP_FEE` (the pool's vault-level floor), so the static fee can never go below it. Once `_minimalSwapFee` sits above that floor, the case is reachable.

EndymionJkb:
Paid grant, but de-prioritized. Closing now as not planned.

PR #1578, opened 2025-11-06
Add wrap/unwrap to balancer burner

elshan-eth:
# Description

## Type of change

- [ ] Bug fix 
- [X] New feature 
- [ ] Breaking change 
- [ ] Dependency changes
- [ ] Code refactor / cleanup
- [ ] Other

## Checklist:

### DIFF
--- a/pkg/standalone-utils/contracts/BalancerFeeBurner.sol
+++ b/pkg/standalone-utils/contracts/BalancerFeeBurner.sol
@@ -55,6 +55,7 @@ contract BalancerFeeBurner is IBalancerFeeBurner, ReentrancyGuardTransient, Vaul
                     revert InvalidBufferTokenOut(step.tokenOut, i);
                 }
             } else {
+                // Reverts if pool is not registered.
                 IERC20[] memory poolTokens = _vault.getPoolTokens(step.pool);
                 if (_tokenExists(stepTokenIn, poolTokens) == false) {
                     revert TokenDoesNotExistInPool(stepTokenIn, i);


pkg/interfaces/contracts/standalone-utils/IBalancerFeeBurner.sol
```diff
@@ -14,6 +14,7 @@ interface IBalancerFeeBurner is IProtocolFeeBurner {
     struct SwapPathStep {
         address pool;
         IERC20 tokenOut;
+        bool isBuffer;
     }
```
EndymionJkb:
```suggestion
```
This is now identical to the BatchRouter definition, so we can remove this and just import it:

`import { SwapPathStep } from "../vault/BatchRouterTypes.sol";`

pkg/standalone-utils/contracts/BalancerFeeBurner.sol
```diff
@@ -104,17 +104,33 @@ contract BalancerFeeBurner is IBalancerFeeBurner, ReentrancyGuardTransient, Vaul
         for (uint256 i = 0; i < steps.length; i++) {
             SwapPathStep memory step = steps[i];
```
EndymionJkb: Per the change above, can just use the struct BatchRouterTypes

import { SwapPathStep } from "@balancer-labs/v3-interfaces/contracts/vault/BatchRouterTypes.sol";

EndymionJkb (approved):
LGTM; just a question about struct duplication.

pkg/standalone-utils/test/foundry/BalancerFeeBurner.t.sol
```diff
@@ -217,6 +226,184 @@ contract BalancerFeeBurnerTest is BaseVaultTest {
...
+
+        uint256 snapshot = vm.snapshot();
+        uint256 amountOut = _vaultPreviewRedeem(waDAI, TEST_BURN_AMOUNT);
+        vm.revertTo(snapshot);
+
+        vm.startPrank(address(feeSweeper));
+        IERC20(address(waDAI)).forceApprove(address(feeBurner), TEST_BURN_AMOUNT);
+
+        uint256 minAmountOut = TEST_BURN_AMOUNT - DELTA;
+        feeBurner.burn(address(0), waDAI, TEST_BURN_AMOUNT, waDAI, minAmountOut, alice, orderDeadline);
```
EndymionJkb:
```suggestion
        feeBurner.burn(address(0), waDAI, TEST_BURN_AMOUNT, waDAI, amountOut, alice, orderDeadline);
```
Do you mean this? Otherwise the amount out from the query isn't used.

pkg/interfaces/contracts/standalone-utils/IBalancerFeeBurner.sol
jubeira: +1; if it's exactly the same let's reuse. It's also more clear for a reader if you're already familiar with the router.

pkg/standalone-utils/contracts/BalancerFeeBurner.sol
```diff
@@ -104,17 +104,33 @@ contract BalancerFeeBurner is IBalancerFeeBurner, ReentrancyGuardTransient, Vaul
...
-                    userData: bytes("")
-                })
-            );
+            uint256 amountOut;
+            uint256 minAmountOut = (i == lastStepIndex) ? params.minAmountOut : 0;
+            if (step.isBuffer) {
+                (, , amountOut) = _vault.erc4626BufferWrapOrUnwrap(
+                    BufferWrapOrUnwrapParams({
+                        kind: SwapKind.EXACT_IN,
+                        direction: step.pool == address(stepTokenIn)
```
jubeira: if `step.pool` is `stepTokenIn` then it's an unwrap, but shouldn't we check that `stepTokenIn` is the wrapper's asset if not?

jubeira: And same for regular swaps; we can at least check that the token out is a registered pool token.

pkg/interfaces/contracts/standalone-utils/IBalancerFeeBurner.sol
```diff
@@ -3,10 +3,31 @@
...
+import { WrappingDirection } from "../vault/VaultTypes.sol";
+
 import { SwapPathStep } from "../vault/BatchRouterTypes.sol";
 import { IProtocolFeeBurner } from "./IProtocolFeeBurner.sol";
 
 interface IBalancerFeeBurner is IProtocolFeeBurner {
+    struct StepsInfo {
+        uint256 currentIndex;
+        mapping(uint256 => SwapPathStep[]) steps;
+    }
```
jubeira: I'm not sure I fully follow why we need this.
Is it to avoid deleting existing paths? From what I can see we're always using the latest index.

If that's the case I think I prefer the previous approach. The `delete` operator should clean up the existing array completely, which is fine.

pkg/standalone-utils/contracts/BalancerFeeBurner.sol
```diff
@@ -36,13 +37,41 @@ contract BalancerFeeBurner is IBalancerFeeBurner, ReentrancyGuardTransient, Vaul
...
+                IERC4626 wrappedToken = IERC4626(step.pool);
+                address underlyingToken = wrappedToken.asset();
+
+                if (isUnwrap && step.tokenOut != IERC20(underlyingToken)) {
+                    revert InvalidBufferTokenOut(step.tokenOut, i);
+                } else if (isUnwrap == false && step.tokenOut != IERC20(address(wrappedToken))) {
+                    revert InvalidBufferTokenOut(step.tokenOut, i);
+                }
+            } else {
+                IERC20[] memory poolTokens = _vault.getPoolTokens(step.pool);
```
jubeira:
```suggestion
                // Reverts if pool is not registered.
                IERC20[] memory poolTokens = _vault.getPoolTokens(step.pool);
```

PR #1590, opened 2025-12-11
Seedless LBP

### LINKED ISSUE #1587: Seedless LBP
LBPs, but without actual reserve tokens.

No need to create a separate factory; just reuse weighted LBPs. Buy only to start with.
We can also override the getters in `PoolInfo` to report balances with virtual reserves.

EndymionJkb:
# Description

Rather than create a new contract, we will simply generalize regular Weighted LBPools to allow seedless ones.

Seedless LBPs are defined by not requiring reserve tokens for initialization (and in fact they must be initialized "one-sided." This is accomplished using a "virtual balance" for the reserve token, which serves as an offset to the real balances. This does not affect the price curve, and since withdrawal uses the real balances, still removes all liquidity at the end.

It would be more transparent if we could just pass something like an "isSeedless" flag on deployment, and then intercept the initialization to get the intended reserve balance. It would not ask for any tokens, and would set the virtual balance to the intended reserve balance. However, this can't be done. Since initialization goes directly through the Vault, there is no way to intercept it at the pool level. So we can't be completely transparent, but have to pass in the virtual balance from the beginning, when the pool is created. (Passing 0 means it's a regular LBP.) The plus side is we can then store it immutably (as noted above, it can be fixed for the life of the pool), saving gas.

LBPs do override the onBeforeInitialize hook, so there we require zero reserve tokens if it is seedless. We then use the "effective balances" (i.e., real + virtual) for all calculations: onSwap, and the invariant hooks, which need to be overridden from IBasePool.

This version supports all possible functionality. Depending on the business requirements, this might be overgeneralized.

While initialization must be one-sided, it does let you add reserve tokens later (still before the sale). For instance, the owner might want to seed a small amount of reserve to allow selling tokens back in immediately. It also lets you do single-token liquidity operations.

EDIT: Per offline discussion, we've disabled all unbalanced liquidity operations, and allowed liquidity removal at any time except during the sale.

## Type of change

- [ ] Bug fix 
- [X] New feature 
- [ ] Breaking change 
- [ ] Dependency changes
- [ ] Code refactor / cleanup
- [ ] Other

## Checklist:

## Issue Resolution

Resolves #1587

### DIFF
--- a/pkg/pool-weighted/test/foundry/FixedPriceLBPool.t.sol
+++ b/pkg/pool-weighted/test/foundry/FixedPriceLBPool.t.sol
@@ -721,7 +721,9 @@ contract FixedPriceLBPoolTest is BaseLBPTest, FixedPriceLBPoolContractsDeployer 
         );
     }
 
-    function testOnBeforeRemoveLiquidityBeforeEndTime() public {
+    function testOnBeforeRemoveLiquidityDuringSale() public {
+        vm.warp(block.timestamp + DEFAULT_START_OFFSET + 1);
+
         // Try to remove liquidity before end time.
         vm.prank(address(vault));
         vm.expectRevert(LBPCommon.RemovingLiquidityNotAllowed.selector);
@@ -734,6 +736,24 @@ contract FixedPriceLBPoolTest is BaseLBPTest, FixedPriceLBPoolContractsDeployer 
             new uint256[](0),
             bytes("")
         );
+    }
+
+    function testOnBeforeRemoveLiquidityBeforeStartTime() public {
+        // Warp to just before start time, where removing liquidity is allowed.
+        vm.warp(block.timestamp + DEFAULT_START_OFFSET - 1);
+
+        vm.prank(address(vault));
+        bool success = IHooks(pool).onBeforeRemoveLiquidity(
+            ZERO_ADDRESS,
+            ZERO_ADDRESS,
+            RemoveLiquidityKind.PROPORTIONAL,
+            0,
+            new uint256[](0),
+            new uint256[](0),
+            bytes("")
+        );
+
+        assertTrue(success, "onBeforeRemoveLiquidity should return true after end time");
     }
 
     function testOnBeforeRemoveLiquidityAfterEndTime() public {
--- a/pkg/pool-weighted/test/foundry/LBPool.t.sol
+++ b/pkg/pool-weighted/test/foundry/LBPool.t.sol
@@ -854,7 +854,9 @@ contract LBPoolTest is WeightedLBPTest {
         );
     }
 
-    function testOnBeforeRemoveLiquidityBeforeEndTime() public {
+    function testOnBeforeRemoveLiquidityDuringSale() public {
+        vm.warp(block.timestamp + DEFAULT_START_OFFSET + 1);
+
         // Try to remove liquidity before end time.
         vm.prank(address(vault));
         vm.expectRevert(LBPCommon.RemovingLiquidityNotAllowed.selector);
@@ -867,6 +869,24 @@ contract LBPoolTest is WeightedLBPTest {
             new uint256[](0),
             bytes("")
         );
+    }
+
+    function testOnBeforeRemoveLiquidityBeforeStartTime() public {
+        // Warp to just before start time, where removing liquidity is allowed.
+        vm.warp(block.timestamp + DEFAULT_START_OFFSET - 1);
+
+        vm.prank(address(vault));
+        bool success = IHooks(pool).onBeforeRemoveLiquidity(
+            ZERO_ADDRESS,
+            ZERO_ADDRESS,
+            RemoveLiquidityKind.PROPORTIONAL,
+            0,
+            new uint256[](0),
+            new uint256[](0),
+            bytes("")
+        );
+
+        assertTrue(success, "onBeforeRemoveLiquidity should return true after end time");
     }
 
     function testOnBeforeRemoveLiquidityAfterEndTime() public {


pkg/pool-weighted/contracts/lbp/LBPool.sol
```diff
@@ -186,7 +268,92 @@ contract LBPool is ILBPool, LBPCommon, WeightedPool {
...
+        seedlessRequest.balancesScaled18[_reserveTokenIndex] =
+            request.balancesScaled18[_reserveTokenIndex] +
+            _reserveTokenVirtualBalanceScaled18;
+
+        uint256 calculatedAmountScaled18 = super.onSwap(seedlessRequest);
+
+        // If we are returning reserve tokens, ensure we have enough real balance.
+        if (
+            request.indexOut == _reserveTokenIndex &&
+            calculatedAmountScaled18 > request.balancesScaled18[_reserveTokenIndex]
```
jubeira: If you go with option 1 you can use the local cache here.

pkg/pool-weighted/test/foundry/SeedlessLBPPool.t.sol
```diff
@@ -0,0 +1,926 @@
...
+            false,
+            bytes("")
+        );
+    }
+
+    /*******************************************************************************
+                                    Price Verification
+    *******************************************************************************/
+
+    function testPriceUsesVirtualBalance() public {
```
jubeira: 👍  this is fine.

I'd add one more test: create a seedless LBP and a mirror LBP with real reserves and 0 virtual reserves.
At any given point in time during the sell, swaps exact in and out should return the exact same result.

This indirectly tests that virtual reserves are indistinguishable from real reserves when it comes to pricing. And this can be fuzzed: amounts in / out, sale timestamp, token decimals, etc.

pkg/pool-weighted/test/foundry/SeedlessLBPPool.t.sol
```diff
@@ -0,0 +1,926 @@
...
+        assertApproxEqAbs(reserveSpent, expectedAmountIn, 1, "Reserve amount mismatch");
+
+        // Verify pool balances updated correctly
+        uint256[] memory balancesAfter = vault.getCurrentLiveBalances(pool);
+        assertEq(balancesAfter[projectIdx], realBalances[projectIdx], "Project should be unchanged");
+        assertEq(balancesAfter[reserveIdx], reserveSpent, "Pool reserve balance mismatch");
+        assertGt(balancesAfter[reserveIdx], 0, "Reserve should now be non-zero");
+    }
+
+    function testSingleTokenExactInRemoveProjectToken() public {
```
jubeira: Same, this one can be blocked.

pkg/pool-weighted/test/foundry/SeedlessLBPPool.t.sol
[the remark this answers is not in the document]
jubeira: On a separate note, maybe removing liquidity proportional before the sale doesn't hurt anyone and allows the owner to correct mistakes without pausing the pool.

pkg/pool-weighted/test/foundry/SeedlessLBPPool.t.sol
```diff
@@ -0,0 +1,926 @@
...
+        vm.expectRevert(LBPool.SingleTokenRemovalExceedsRealReserve.selector);
+        vm.prank(bob);
+        router.removeLiquiditySingleTokenExactIn(pool, bptAmount, reserveToken, 1, false, bytes(""));
+    }
+
+    /*******************************************************************************
+                                Invariant Tests
+    *******************************************************************************/
+
+    function testComputeInvariantUsesEffectiveBalances() public view {
```
jubeira: 👍 

Could fuzz balances too

pkg/pool-weighted/test/foundry/SeedlessLBPPool.t.sol
```diff
@@ -0,0 +1,926 @@
...
+        assertEq(
+            balancesRaw[reserveIdx],
+            realBalancesRaw[reserveIdx] + reserveTokenVirtualBalance,
+            "Reserve raw should include virtual"
+        );
+        assertEq(
+            lastBalancesLiveScaled18[reserveIdx],
+            realLastBalancesLiveScaled18[reserveIdx] + reserveTokenVirtualBalance,
+            "Reserve live should include virtual"
+        );
```
jubeira: Might be worth doing the test with a 6-decimal reserve token, JIC. That will force us to see a few edge cases.

pkg/pool-weighted/test/foundry/SeedlessLBPPool.t.sol
```diff
@@ -0,0 +1,926 @@
...
+
+        assertEq(poolBalances[projectIdx], vaultBalances[projectIdx], "Project balance mismatch");
+        assertEq(poolBalances[reserveIdx], vaultBalances[reserveIdx], "Reserve balance should match exactly");
+
+        vm.warp(block.timestamp + DEFAULT_END_OFFSET + 1);
+
+        // Included for 100% test coverage
+        vm.startPrank(bob);
+        IERC20(nonSeedlessPool).approve(address(router), type(uint256).max);
+        router.removeLiquiditySingleTokenExactIn(
```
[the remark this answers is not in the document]
jubeira: as I was writing this I was thinking: there is _one_ reason to remove unbalanced, which is in case the token transfers break down.

Arguably if your token breaks transfers after the sale there's something deeply wrong and the project is beyond saving though. Mentioning it for completion, but probably we can consider that out of scope.

jubeira (commented):
Looking good, great job @EndymionJkb !

Just a bunch of mostly minor comments and test suggestions; otherwise LGTM!

pkg/pool-weighted/test/foundry/SeedlessLBPPool.t.sol
EndymionJkb: Added a "non18" infrastructure (duplicated vs. generalized, to avoid breaking existing tests)

pkg/pool-weighted/test/foundry/SeedlessLBPPool.t.sol
```diff
@@ -835,10 +615,179 @@ contract SeedlessLBPTest is WeightedLBPTest {
...
+        address seededPool;
+        IERC20 projToken;
+        IERC20 resToken;
+        uint256 projIdx;
+        uint256 resIdx;
+        uint256 virtualBalance;
+        uint256 projectInitAmount;
+    }
+
+    function testSeedlessVsSeededSwaps__Fuzz(uint256 swapAmount, uint256 timeOffset, bool useNon18Decimals) public {
```
jubeira: Good stuff.

Maybe one small suggestion here: I think it'd be better to have 2 simpler tests as opposed to one complex test. In particular, I'd remove the `useNon18Decimals` argument, and just do one test for 18-decimal tokens and another one for non-18.

The code will be mostly the same and you can even factor it out to use the same underlying functions after initialization, but you save all the `if / else`s in the test, you'll get more coverage and it'll be simpler to read.

jubeira (approved):
LGTM, I like this. I only have one last suggestion for the added fuzz test.

Out of scope of this PR, but writing it here so that I don't forget: maybe we should enable remove liquidity _before_ the sale to allow the owner to correct mistakes. Let's discuss this off-thread, and maybe address later in a separate PR.

PR #1600, opened 2026-01-07
Unbalanced Add Via Swap Router patch (V3)

EndymionJkb:
# Description

We've gone back and forth on this many times; final state seems to be that the EXACT_IN path *is* reachable, and the only true bug was addressed by the first patch: simply fixing the limit.

Also very slightly refactored/optimized the existing checks. They are technically independent, so should be separate statements, and it's a slight optimization to check the one that *can* happen before the one that can't.

This PR makes those minimal changes, and adds a test for the ExactIn case.
## Type of change

- [X] Bug fix 
- [ ] New feature 
- [ ] Breaking change 
- [ ] Dependency changes
- [ ] Code refactor / cleanup
- [ ] Other

## Checklist:

### DIFF
--- a/pkg/vault/test/foundry/UnbalancedAddViaSwapRouter.t.sol
+++ b/pkg/vault/test/foundry/UnbalancedAddViaSwapRouter.t.sol
@@ -27,6 +27,17 @@ contract UnbalancedAddViaSwapRouterTest is BaseVaultTest {
     using ArrayHelpers for *;
     using FixedPoint for uint256;
 
+    // Needed to avoid stack-too-deep
+    struct TestBalances {
+        uint256 userWeth;
+        uint256 userDai;
+        uint256 userBpt;
+        uint256 userEth;
+        uint256 poolWeth;
+        uint256 poolDai;
+        uint256 totalSupply;
+    }
+
     string constant POOL_VERSION = "Pool v1";
     uint256 constant DEFAULT_AMP_FACTOR = 200;
     uint256 constant DELTA_RATIO = 1e15; // 0.1% delta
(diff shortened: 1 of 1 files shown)


pkg/vault/test/foundry/UnbalancedAddViaSwapRouter.t.sol
```diff
@@ -226,4 +226,52 @@ contract UnbalancedAddViaSwapRouterTest is BaseVaultTest {
...
+                exactBptAmountOut: requestedBpt,
+                exactToken: weth,
+                exactAmount: exactAmount,
+                maxAdjustableAmount: maxAdjustableAmount,
+                addLiquidityUserData: bytes(""),
+                swapUserData: bytes("")
+            });
+
+        vm.prank(alice);
+        uint256[] memory amountsIn = unbalancedAddViaSwapRouter.addLiquidityUnbalanced(
```
jubeira: Should we contrast this against actual balances before and after?
There's actually a test called `testAddProportionalAndSwapExactIn`, but now I'm unsure whether it actually goes through the 'exact in' branch (which we were thinking wasn't used).

Ideally we could have one test that fuzzes through the exact in branch, and another one for exact out branch. The overall result should always be the same: user adds an exact amount of tokens A / up to a max amount of tokens B specified in the request.

pkg/vault/test/foundry/UnbalancedAddViaSwapRouter.t.sol
jubeira: This is fine, but it can also happen for exact out and I don't think it's tested.
If we just take a working scenario and send a lower max adjustable amount, it should revert for the exact out branch as well.

jubeira (approved):
Changes look good! I think this is how it should behave.

I do have a few observations regarding the tests, which are halfway out of scope of this PR.

I'd like to see the following with the tests:
- Have explicit tests for exact in / exact out branches.
  - Compare balances before and after with amounts returned by the router in every case
  - Also compare minted BPT shares (should be trivial, as they are passed directly to add liquidity proportional)
  - If they are fuzz tests, even better
- Check revert reasons for both exact in and exact out branches (adjustable amount in > max adjustable amount in)

Out of scope: `AmountInDoesNotMatchExact` error, which shouldn't ever happen (but keeping the check makes total sense as it makes it very easy to reason about).

pkg/vault/test/foundry/UnbalancedAddViaSwapRouter.t.sol
```diff
@@ -227,51 +300,48 @@ contract UnbalancedAddViaSwapRouterTest is BaseVaultTest {
...
-        proportionalAmounts[0] = (balances[0] * requestedBpt) / totalSupply;
-        proportionalAmounts[1] = (balances[1] * requestedBpt) / totalSupply;
-
-        // Set maxAdjustableAmount just below the proportional add amount
-        // If we took the EXACT_OUT path, it would add to this and exceed the limit
-        // Th EXACT_IN path subtracts from it, so should not revert
-        uint256 maxAdjustableAmount = proportionalAmounts[daiIdx] - 1;
+    // Test helpers
+
+    function _getBalances() private view returns (TestBalances memory balances) {
```
jubeira: FWIW there's already a `getBalances` function in `BaseVaultTest` and a `Balances` struct that you can reuse.

It collects this info and then some, but should be good to use as it is. That will save a few lines here.

jubeira (approved):
LGTM; great job! Just one minor comment regarding code reuse in the tests.

PR #1601, opened 2026-01-07
Weighted pool updates

jubeira:
# Description

Minor correctness improvements.
Credits to @EndymionJkb .

## Type of change

- [ ] Bug fix 
- [ ] New feature 
- [ ] Breaking change 
- [ ] Dependency changes
- [ ] Code refactor / cleanup
- [x] Other

## Checklist:

## Issue Resolution

N/A


EndymionJkb (approved):
LGTM! (Pending CI.)

joaobrunoah (commented):
It's a pity that we can't have immutable arrays to iterate on weights and minBalances, so the code looks a bit uglier. I thought about iterating on weights using assembly and memory pointers, but it's not possible since the variables are immutable.

We could, maybe, store a reduced version of all weights as a bytes32 immutable, so we could access the weight directly, improving the function _getNormalizedWeight and allowing us to use a for loop on _getNormalizedWeights , (and, of course, do the same with min balances) but even so, it looks too cumbersome and the gain is probably marginal.

So, the PR looks good to me as is, I couldn't find any major improvement.

PR #1660, opened 2026-05-25
Fix price impact helper for proportional adds

### LINKED ISSUE #1208: Small price impact helper fix
https://github.com/balancer/b-sdk/pull/316/files#diff-ae161b3769af2b4472a3b48774027670c57f0c4df2a67359a2b6e720d0982946L136

TUPM96:
## Summary
- Return zero price impact for proportional or near-proportional add-liquidity inputs where deltas do not contain both positive and negative sides.
- Avoid running the delta zero-out path when all delta BPT values are zero, and only iterate over non-zero delta BPTs.
- Add unit and integration coverage for the all-zero delta path and proportional add-liquidity price impact.

Closes #1208

## Testing
- `prettier --write pkg/standalone-utils/contracts/PriceImpactHelper.sol pkg/standalone-utils/contracts/test/PriceImpactHelperMock.sol pkg/standalone-utils/test/foundry/PriceImpactUnit.t.sol pkg/standalone-utils/test/foundry/PriceImpact.t.sol`
- `FOUNDRY_ALLOW_PATHS=[...] forge test --match-path test/foundry/PriceImpactUnit.t.sol`
- `FOUNDRY_ALLOW_PATHS=[...] forge test --match-path test/foundry/PriceImpact.t.sol`
- `git diff --check`

### DIFF
--- a/pkg/standalone-utils/contracts/PriceImpactHelper.sol
+++ b/pkg/standalone-utils/contracts/PriceImpactHelper.sol
@@ -42,6 +42,11 @@ contract PriceImpactHelper is CallAndRevert {
             deltas[i] = proportionalAmountsOut[i].toInt256() - exactAmountsIn[i].toInt256();
         }
 
+        (bool hasPositiveDelta, bool hasNegativeDelta) = _hasPositiveAndNegative(deltas);
+        if (!hasPositiveDelta || !hasNegativeDelta) {
+            return 0;
+        }
+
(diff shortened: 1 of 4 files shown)


EndymionJkb:
This is better, but still closing. First, I don't believe we ever deployed this (so likely never will). Also, the fix in #1208 was for the SDK and already shipped in balancer/b-sdk#316, so there's nothing to port here.

PR #1665, opened 2026-07-21
Add E-CLP swap invariant fuzz tests

joaobrunoah:
# Description

Adds fuzz and stateful-fuzz coverage for **one specific E-CLP configuration** — depth on this config rather than breadth across many, since broad parameter robustness is already covered elsewhere.

| Param | Value |
|---|---|
| `alpha` | `0.980392156862745098` |
| `beta` | `1.000000630371029932` |
| `c`, `s` | `0.707106781186547524` (45°) |
| `lambda` | `300` |
| swap fee | `1e14` (0.01%) |

**No vulnerability was found.** No invariant leak, no profitable attack sequence, and no accumulating value extraction.

## What is covered

**Foundry fuzz** — 7 tests, 10 000 runs each:

| File | Property |
|---|---|
| `SwapInvariantECLPSpecific.sol` | invariant never decreases on exact-in / exact-out swaps |
| `SwapRoundTripECLPSpecific.sol` | do-undo round trip cannot profit; N split swaps never out-yield one swap |
| `LiquidityRoundTripECLPSpecific.sol` | add→swap→remove and remove→swap→add cannot profit; proportional add→remove never returns more than deposited; BPT rate does not decrease |

Shared setup lives in `utils/BaseECLPSpecificTest.sol`.

**Medusa stateful campaign** — `fuzz/DrainECLPSpecific.medusa.sol`, pinning this same config rather than `GyroEclpPoolDeployer`'s hardcoded params. Exposes swaps (both directions and kinds), proportional and unbalanced adds, and proportional and single-token-exact-in removes for arbitrary sequencing. Properties: BPT-rate monotonicity, vault solvency, total supply against balances, spot price within `[alpha, beta]`, and atomic round-trip no-profit checks for both swaps and liquidity.

## Initialization at spot price 1

Every run initializes at a spot price of exactly 1.

The E-CLP is homogeneous of degree 1 — scaling both balances scales the invariant and leaves the price unchanged — so spot price depends only on the balance ratio, and the price-1 ratio is a constant for these params. Derived from the virtual-offset form in `GyroECLPMath.virtualOffset0/1`, with `c == s` giving `tau(1) = (0, 1)` exactly:

```
balance1 / balance0 = 9998.999999994488463412
```

So the tests fuzz a single liquidity scalar and derive the counterpart, rather than fuzzing both balances independently (which would land each run at an arbitrary price in `[alpha, beta]`). A post-init assertion checks the realized price via the production `GyroECLPMath.computePrice` path.

**This interval is deliberately lopsided.** `beta` sits only 6.3e-7 above 1.0 while `alpha` is 0.0196 below, so price 1 is essentially at the `beta` edge: the pool holds ~0.00995% of the reachable token-0 range and ~99.99% of the token-1 range. The 9999:1 balance ratio is a symptom of that, and it is why operations are sized against the *opposite* balance — that is the true headroom.

A counterintuitive consequence is worth recording. Exact price 1 and 6-decimal tokens look like they should conflict, since rounding a raw amount at 6 decimals perturbs the ratio and `lambda = 300` on a narrow interval makes price sensitive. The opposite holds: `d(ln p)/d(ln ratio) ~ 6.3e-7`, so the ratio is ~1.6M times more sensitive to price than price is to ratio, and rounding is almost entirely absorbed. Measured worst case over 10 000 runs per threshold:

| Quantity | Worst observed |
|---|---|
| relative balance-ratio deviation | ~6.0e-11 |
| price deviation, `computePrice` (production path) | 45 wei (4.5e-17 relative) |
| price deviation, unclamped `calcSpotPrice0in1` | 45 wei |

Tolerance is 1000 wei (1e-15), ~20x the measured worst case. `computePrice` clamps to `[alpha, beta]`, which could in principle mask upward deviation; cross-checking the unclamped `calcSpotPrice0in1` gives the identical 45-wei worst case, so the clamp hides nothing.

## Findings

**1. `removeLiquidityProportional` drops the BPT rate by exactly 1 wei.** Medusa shrank this to a single call from the initial state. The invariant is recomputed from scratch on the reduced balances and rounds down; the burn pays for it, and the effect is ~3.4e-25 relative. This is not value leaving the pool, and it matches the slack the existing `AddAndRemoveLiquidityECLP` campaign already carries. `property_currentBptRate` therefore uses `+ 1 >=`, pinned at exactly 1 wei so any accumulation still fails.

Over **1 372 783 calls across 68 633 sequences**, the optimizer's worst reachable rate was that same 1 wei — **the deficit does not accumulate**.

**2. The split-swap assertion has exactly zero margin.** Probing with `assertGe(single - split, 1)` fails with `0 < 1` in 23 runs — exact ties occur readily. This is the intended outcome (one wei of extractable value fails the test), but it means a future failure here is signal, not flake, and should not be "fixed" with a tolerance. Expected for a fee-and-round-down AMM: splitting costs N fee round-ups and N output truncations, and at small sizes both paths land on the same raw unit.

**3. Round-trip loss is exactly 2x the swap fee** to five significant figures (`2.0000e-4`), matching theory since the whole price interval is within 2% of 1.0. The shipped bound is `3 *` the fee rather than the tightest passing `2 *`, whose only margin is the `f²` term (~5e-5 of the bound). `2 *` is verified green at 10 000 runs if a tighter assertion is preferred.

## Note on the derived params

The derived params were computed offchain from `pkg/pool-hooks/jupyter/SurgeECLP.ipynb`, but **not by running it**. That notebook does its arithmetic in float64 (`D100 = 1e100` is a float), giving ~16 significant digits for 38-decimal output. Its algorithm was reimplemented in exact arbitrary-precision integer arithmetic.

This is a latent trap for anyone using that notebook with a `beta` close to 1 — `floor(beta * 1e18)` for a beta like `1.00000000001` needs 19 significant digits and is silently corrupted. Worth its own issue; this PR does not fix it.

The constants satisfy every predicate in `validateParams` / `validateDerivedParamsLimits` with ~3 orders of magnitude of headroom (`scnorm2` deviation -2 against a 1e3 tolerance; `|tauAlpha|^2`, `|tauBeta|^2`, `dSq` deviations -1.13e20 against 1e23).

## Type of change

- [ ] Bug fix 
- [ ] New feature 
- [ ] Breaking change 
- [ ] Dependency changes
- [ ] Code refactor / cleanup
- [x] Other 

## Checklist:

- [ ] Changeset is included 

## Verification

Full `pkg/pool-gyro` suite: **261 passed, 0 failed**.

```
FOUNDRY_FUZZ_MAX_TEST_REJECTS=0 forge test --match-path "test/foundry/*ECLPSpecific.sol"

[PASS] testAddSwapRemoveIsNotProfitable__Fuzz         (runs: 10000)
[PASS] testProportionalAddRemoveRoundTrip__Fuzz       (runs: 10000)
[PASS] testRemoveSwapAddIsNotProfitable__Fuzz         (runs: 10000)
[PASS] testSplitSwapNeverBeatsSingleSwap__Fuzz        (runs: 10000)
[PASS] testSwapRoundTripCannotProfit__Fuzz            (runs: 10001)
[PASS] testSwapExactInInvariantNeverDecreases__Fuzz   (runs: 10000)
[PASS] testSwapExactOutInvariantNeverDecreases__Fuzz  (runs: 10000)
7 passed; 0 failed
```

**Non-vacuity was measured, not assumed.** All seven pass with `FOUNDRY_FUZZ_MAX_TEST_REJECTS=0`, where a single rejection aborts the test — **zero runs are silently skipped**. The two round-trip files contain no `vm.assume` and no `try/catch` at all. Separately, tightening every `assertLe` to `assertLt` still passes 3000 runs, so attackers *strictly* lose value rather than breaking even: measured losses of 1.47e-5 and 4.00e-6 relative to the token-0 balance, i.e. fee-scale and orders of magnitude above rounding dust.

Medusa campaign:

```
cd pkg/pool-gyro && medusa fuzz --config medusa-drain-eclp.json
```

Ran 29m59s / 1 372 783 calls / 68 633 sequences / 10 workers / sequence length 20 — 26 tests passed, 0 failed. Coverage saturated at ~8 min (branches flat at 6 083). 2–4 h is the honest number for a serious campaign.

## Scope and known gaps

A non-decreasing invariant is necessary but **not sufficient** to show a pool cannot be drained — it is trivially satisfied by adding liquidity. That is why the BPT-rate property and the round-trip/extraction scenarios are here.

Remaining gaps, stated plainly:

- The BPT-rate check is **low resolution** for single operations: it holds with a bare `assertGe` in the Foundry tests only because those sequences move the invariant by less than one part in 1e18. Medusa is what actually exercises it.
- Not exposed to Medusa: `removeLiquiditySingleTokenExactOut`, `donate` (donation is disabled on this pool), and batch/composite router paths.
- One property, `noActorGainsBothTokens`, was written, found to fail, and **dropped as unsound** — the failure was benign cross-actor arbitrage (one actor moves the price, another profits), not a pool bug. Per-actor profit is not an invariant when a second actor can move the price. It was replaced with an atomic add-then-remove no-profit check.
- Results are bounded to this one configuration by design.

🤖 Generated with [Claude Code](https://claude.com/claude-code)


pkg/pool-gyro/test/foundry/fuzz/DrainECLPSpecific.medusa.sol
```diff
@@ -0,0 +1,571 @@
...
+    /// @notice BPT total supply must equal the sum of every balance that can hold it.
+    function property_totalSupplyMatchesBalances() public view returns (bool) {
+        IERC20 bpt = IERC20(address(pool));
+        uint256 sum = bpt.balanceOf(address(0)) + bpt.balanceOf(alice) + bpt.balanceOf(bob) + bpt.balanceOf(lp);
+
+        return bpt.totalSupply() == sum;
+    }
+
+    /// @notice The spot price must stay inside the configured price interval.
+    function property_spotPriceWithinBounds() public view returns (bool) {
```
EndymionJkb, line 149: The comparison at the end is always true. You're checking whether the price is within [alpha, beta], but `_computeSpotPrice` goes through `GyroECLPMath.computePrice`, which restricts its own return value to exactly that interval.

The unrestricted equivalent is `GyroECLPMath.calcSpotPrice0in1`, which takes the invariant instead of the virtual offsets: `calcSpotPrice0in1(balancesLiveScaled18, params, derivedParams, int256(pool.computeInvariant(balancesLiveScaled18, Rounding.ROUND_DOWN)))`. One caveat: it can revert near the interval endpoints where the denominator approaches zero, and Medusa treats a reverting property function as a failure, so it wants a denominator check or a try/catch around it.

pkg/pool-gyro/test/foundry/LiquidityRoundTripECLPSpecific.sol
```diff
@@ -0,0 +1,282 @@
...
+        _swapExactIn(setup, swapAmountScaled18, swap0To1);
+
+        _addProportional(setup, bptIn);
+
+        assertEq(IERC20(setup.pool).balanceOf(alice), stake, "Attacker did not restore the original BPT position");
+
+        uint256 valueAfter = _attackerValueScaled18(setup);
+
+        assertLe(valueAfter, valueBefore, "Remove -> swap -> add was profitable");
+        assertGe(vault.getBptRate(setup.pool), bptRateBefore, "BPT rate decreased over remove -> swap -> add");
```
EndymionJkb, line 135: This assertion (and the one at 89) are assertGe(vault.getBptRate(setup.pool), bptRateBefore, ...) with no slack, while the companion Medusa property for the same quantity is written as vault.getBptRate(address(pool)) + 1 >= _initialBptRate, because removeLiquidityProportional was shown to drop the rate by exactly one wei in a single call from the initial state. So the two tests are inconsistent.

Easiest thing would be to match the +1 here.

pkg/pool-gyro/medusa-drain-eclp.json
```diff
@@ -0,0 +1,89 @@
+{
+  "fuzzing": {
```
EndymionJkb, line 2: Nothing seems to reference this file? `pkg/pool-gyro/package.json` has one Medusa script, `"test:medusa": "medusa fuzz --config medusa.json"`, and `medusa.json` was edited in this PR to add `DrainECLPSpecificMedusa` to `targetContracts`, so the campaign that actually runs is the shared one, while I think the numbers in the PR description came from this one.

Maybe add `"test:medusa:drain-eclp": "medusa fuzz --config medusa-drain-eclp.json"`

pkg/pool-gyro/test/foundry/SwapInvariantECLPSpecific.sol
```diff
@@ -0,0 +1,106 @@
...
+        vm.assume(amountInRaw > 0);
+
+        uint256 invariantBefore = _computeInvariant(setup.pool);
+
+        vm.prank(lp);
+        try
+            router.swapSingleTokenExactIn(setup.pool, tokenIn, tokenOut, amountInRaw, 0, MAX_UINT256, false, bytes(""))
+        {
+            // Swap succeeded; fall through to the invariant check.
+        } catch {
```
EndymionJkb, line 49: Is this `catch` reachable? (The two other files in the suite call the same swap under the same bounds without failure handling).

I suspect it will still work if you drop the try/catch. Otherwise, consider catching the specific error and re-raise anything else.

pkg/pool-gyro/test/foundry/LiquidityRoundTripECLPSpecific.sol
```diff
@@ -0,0 +1,282 @@
...
+ * The attacker changes the composition of their holdings, so the two tokens have to be valued against each other
+ * somehow. This suite values them 1:1, summing the 18-decimal-scaled balances of both tokens. That choice is
+ * defensible for this specific pool: the whole reachable price interval is
+ * `[alpha, beta] = [0.980392156862745098, 1.000000630371029932]`, so in any state the pool can reach, one unit of
+ * token 0 is worth between 0.9804 and 1.0000007 units of token 1, and the pool starts at exactly 1.
+ *
+ * The residual ambiguity of the 1:1 sum is therefore at most 6.3e-7 relative (`beta - 1`), in the direction that
+ * favours an attacker who ends up long token 0. That is more than two orders of magnitude smaller than the 1e-4 swap
+ * fee every attacker in these sequences has to pay, so the conclusion is not load-bearing on the valuation choice.
+ * No tolerance is added on top: the assertions below are exact.
```
EndymionJkb, line 33: Alternative paragraph. The original one only addresses one direction, and may be inaccurate about why the assertions are safe.

```suggestion
     * The 1:1 sum understates the attacker's gain in two cases. When the price is above 1 and the attacker ends up
     * holding more token 0, the error is at most `beta - 1` (6.3e-7). When the price is below 1 and the attacker ends
     * up holding less token 0, the error is at most `1 - alpha` (1.96e-2), four orders of magnitude larger. The second
     * case is the ordinary one: selling token 0 both reduces the holding and pushes the price down.
     *
     * What keeps the error under the fee is the trade cap, not the width of the price interval. Every operation here
     * is capped at 10% of the token 0 balance, which holds the price within 6.3e-8 of 1, so the error is ~6.3e-4 of
     * the 1e-4 swap fee the attacker pays. Widening that cap breaks the argument: at 20x the token 0 balance the
     * error is an eighth of the fee, and at 2000x it is ten times the fee.
     * No tolerance is added on top: the assertions below are exact.
```

Write the true value of the attacker's holdings in units of token 1 as `V = b0 * p + b1`, and the value this test measures as `M = b0 + b1`. The difference in their changes is `dV - dM = (p - 1) * db0`.

The test passes when `dM` is at most zero, and the attacker really profits when `dV` is above zero, so a false pass needs `(p - 1) * db0` to be positive.

A price above 1 with more token 0 gives that, bounded by `beta - 1`. So does a price below 1 with less token 0, bounded by `1 - alpha`. The current text covers only the first.

pkg/pool-gyro/test/foundry/SwapInvariantECLPSpecific.sol
```diff
@@ -0,0 +1,106 @@
...
+
+        _assertSpotPriceIsOne(setup.pool);
+
+        (IERC20 tokenIn, IERC20 tokenOut) = swap0To1 ? (setup.token0, setup.token1) : (setup.token1, setup.token0);
+
+        // Cap the trade at 10% of the token 0 balance, in both directions. Token 0 is by far the scarcer side here
+        // (the pool starts near `beta`), so it is what limits the reachable trade size whichever way the swap goes.
+        swapAmountScaled18 = bound(swapAmountScaled18, _MIN_SWAP_SCALED18, balance0Scaled18 / 10);
+        uint256 amountInRaw = _toRawAmount(swapAmountScaled18, swap0To1 ? setup.decimals0 : setup.decimals1);
+        vm.assume(amountInRaw > 0);
```
EndymionJkb, line 40:
```suggestion
```
The smallest swap the test allows is `_MIN_SWAP_SCALED18`, which is 1e16. The fewest decimals it allows is `_MIN_DECIMALS`, which is 6. So `_toRawAmount` divides by at most 1e12 and returns at least 1e4. The condition is always true. Same thing below at line 80.

pkg/pool-gyro/test/foundry/utils/BaseECLPSpecificTest.sol
```diff
@@ -0,0 +1,244 @@
...
+
+    uint256 internal constant _SWAP_FEE_PERCENTAGE = 1e14; // 0.01%
+
+    uint8 internal constant _MIN_DECIMALS = 6;
+    uint8 internal constant _MAX_DECIMALS = 18;
+
+    // Balance bounds for token 0 (scaled 18-decimal values). Token 1 is ~9999x larger (see
+    // `_PRICE_ONE_BALANCE_RATIO`), so the upper bound is what keeps the sum of the balances below `_MAX_BALANCES`
+    // (1e34). The lower bound keeps the token 0 raw amount at 6 decimals at or above 1e10 units, so that the
+    // truncation performed by `_toRawAmount` perturbs the balance ratio by at most ~1e-10 in relative terms.
```
EndymionJkb, line 50: Suggest rewording this comment. The bound itself is fine, but it names a limit that is nowhere near being reached.

```suggestion
    // Balance bounds for token 0 (scaled 18-decimal values). Token 1 is ~9999x larger (see
    // `_PRICE_ONE_BALANCE_RATIO`), so the upper bound is set by what the LP can hold: at 1e26 the token 1 balance
    // is ~1e30, against the 1e32 minted by `_TOKEN_MINT_AMOUNT`. That leaves `_MAX_BALANCES` (1e34) four orders of
    // magnitude away, so it is never reached. The lower bound keeps the token 0 raw amount at 6 decimals at or
    // 1e10 units, so that the truncation performed by `_toRawAmount` perturbs the balance ratio by at most ~1e-10
    // in relative terms.
```
At `_MAX_BALANCE0_SCALED18`, which is 1e26, the two balances sum to about 1e30, against a `_MAX_BALANCES` of 1e34. `_TOKEN_MINT_AMOUNT` is what runs out first, and even that would allow roughly 1e28.

pkg/pool-gyro/test/foundry/fuzz/DrainECLPSpecific.medusa.sol
```diff
@@ -0,0 +1,571 @@
...
+        assert(IERC20(address(pool)).totalSupply() == snapshot.totalSupply - exactBptIn);
+        assert(_poolBalance(tokenIndex) == (tokenIndex == 0 ? snapshot.pool0 : snapshot.pool1) - amountOut);
+        assert(_poolBalance(1 - tokenIndex) == (tokenIndex == 0 ? snapshot.pool1 : snapshot.pool0));
+    }
+
+    /**
+     * @notice Add liquidity proportionally and immediately take it back out, by any actor.
+     * @dev An atomic round trip must never return more of either token than it consumed.
+     *
+     * Note this is deliberately checked *atomically* rather than as a standing per-actor property. A global
```
EndymionJkb, line 381: Suggest adding one sentence here (and to `roundTripSwap` below). "atomically" can be confusing in this context.

```suggestion
     * Note this is deliberately checked *atomically* rather than as a standing per-actor property. Atomic here means
     * one fuzz call, not one Vault session: the add and the remove are separate router calls, so the Vault's
     * roundtrip swap fee never applies to this sequence, and the path being tested is the one without it. A global
```
The roundtrip swap fee is recorded in `_addLiquidityCalled`, which is keyed by session and pool together. Two router calls mean two unlock sessions, so that record never carries over from the add to the remove. Testing the path without the fee is more interesting, but as written a reader can easily conclude this function covers the roundtrip fee when it actually doesn't.

pkg/pool-gyro/test/foundry/utils/BaseECLPSpecificTest.sol
```diff
@@ -0,0 +1,244 @@
...
+    // Amount minted to the LP for every freshly created token.
+    uint256 internal constant _TOKEN_MINT_AMOUNT = 1e32;
+
+    /**
+     * @dev Ratio `balance1 / balance0` that puts the pool at a spot price of exactly 1, as an 18-decimal number.
+     * Derived from the virtual offset form used by `GyroECLPMath.virtualOffset0/1`, where
+     * `(A^-1 tau) = (lambda*c*tau.x + s*tau.y, -lambda*s*tau.x + c*tau.y)`, and
+     * `x(p)/r = (A^-1 tauBeta)_x - (A^-1 tau(p))_x`, `y(p)/r = (A^-1 tauAlpha)_y - (A^-1 tau(p))_y`.
+     * Since `c == s`, `tau(1) = (0, 1)`, which gives `x(1)/r = 0.0200582737739962149...` and
+     * `y(1)/r = 200.5626794660776018...`, hence `y/x = 9998.99999999448846341...`.
```
EndymionJkb, line 66:
```suggestion
     * Since `c == s`, `tau(1) = (0, 1)`, which gives `x(1)/r = 0.02005827377399621`, `y(1)/r = 200.56267946607760`
     * and `y/x = 9998.99999999448846`. Note the stored `tauAlpha` and `tauBeta` are normalized so that the squared
     * length of tau equals `dSq` rather than 1, while `tau(1)` above is not, so the last digits of the ratio depend
     * on which convention is applied. The two differ by 2.0e-17 in relative terms, which is 1.26e-5 wei of price:
     * far below the ~45 wei of rounding the initialization already carries.
```
Using one convention throughout gives 9998.999999994488263612. The shipped constant matches the mixed version instead, to within 0.109 wei. The price moves by 6.303e-7 per unit of relative change in the ratio, so a 2.0e-17 difference in the ratio is 1.26e-5 wei of price, six orders of magnitude below the 45 wei the PR already measured. Nothing about the value needs changing, only the claim that the digits are exact.

pkg/pool-gyro/test/foundry/utils/BaseECLPSpecificTest.sol
```diff
@@ -0,0 +1,244 @@
...
+        return (address(0), bytes(""));
+    }
+
+    /// @inheritdoc BaseVaultTest
+    function initPool() internal pure override {
+        // solhint-disable-previous-line no-empty-blocks
+    }
+
+    function createPoolFactory() internal override returns (address) {
+        return address(new GyroECLPPoolFactory(IVault(address(vault)), 365 days, "Factory v1", "Pool v1"));
```
EndymionJkb, line 121:
```suggestion
        return address(deployGyroECLPPoolFactory(IVault(address(vault))));
```
Can this go through the deployer like the other tests? Or is there a reason it has to be done this way here?

pkg/pool-gyro/test/foundry/LiquidityRoundTripECLPSpecific.sol
```diff
@@ -0,0 +1,282 @@
...
+ * ## Valuation of the attacker's holdings
+ *
+ * The attacker changes the composition of their holdings, so the two tokens have to be valued against each other
+ * somehow. This suite values them 1:1, summing the 18-decimal-scaled balances of both tokens. That choice is
+ * defensible for this specific pool: the whole reachable price interval is
+ * `[alpha, beta] = [0.980392156862745098, 1.000000630371029932]`, so in any state the pool can reach, one unit of
+ * token 0 is worth between 0.9804 and 1.0000007 units of token 1, and the pool starts at exactly 1.
+ *
+ * The residual ambiguity of the 1:1 sum is therefore at most 6.3e-7 relative (`beta - 1`), in the direction that
+ * favours an attacker who ends up long token 0. That is more than two orders of magnitude smaller than the 1e-4 swap
```
EndymionJkb, line 31:
```suggestion
 * favors an attacker who ends up long token 0. That is more than two orders of magnitude smaller than the 1e-4 swap
```
'Murica! [Eagle Emoji]

pkg/pool-gyro/test/foundry/LiquidityRoundTripECLPSpecific.sol
```diff
@@ -0,0 +1,282 @@
...
+ * fee every attacker in these sequences has to pay, so the conclusion is not load-bearing on the valuation choice.
```
EndymionJkb, line 32:
```suggestion
 * fee every attacker in these sequences has to pay, so the conclusion does not depend on the valuation choice.
```
I **hate** "load-bearing" - we're not building a bridge here. Kill it with fire!

pkg/pool-gyro/test/foundry/LiquidityRoundTripECLPSpecific.sol
```diff
@@ -0,0 +1,282 @@
...
+            setup.decimals1
+        );
+
+        vm.prank(alice);
+        bptOut = router.addLiquidityUnbalanced(setup.pool, exactAmountsIn, 0, false, bytes(""));
+    }
+
+    /**
+     * @dev Adds liquidity proportionally as the attacker, for an exact BPT amount out.
+     * @dev The `maxAmountsIn` limits are the attacker's full balances, so they are never binding. They have to be
```
EndymionJkb, line 215:
```suggestion
     * The `maxAmountsIn` limits are the attacker's full balances, so they are never binding. They have to be
```
No stuttering :)

pkg/pool-gyro/test/foundry/fuzz/DrainECLPSpecific.medusa.sol
```diff
@@ -0,0 +1,571 @@
...
+        medusa.prank(lp);
+        try router.removeLiquidityProportional(address(pool), exactBptIn, new uint256[](2), false, bytes("")) returns (
+            uint256[] memory amountsOut
+        ) {
+            assert(amountsOut[0] > 0 && amountsOut[1] > 0);
+        } catch {
+            // A proportional exit from a non-degenerate pool must always succeed.
+            assert(false);
+        }
+    }
```
EndymionJkb, line 430:
```suggestion
    /**
     * @notice Liveness: a proportional exit by the LP must never revert, whatever the pool state.
     * @dev Proportional exits do not move the price, so no reachable state may lock an LP out of one. Each amount
     * out is `poolBalance * exactBptIn / totalSupply`, so the guard below skips the states where that rounds to
     * zero for either token, which is the only reason a redeemed amount could come back empty.
     */
    function lpCanAlwaysExitProportionally() external {
        uint256 exactBptIn = IERC20(address(pool)).balanceOf(lp).mulDown(LIVENESS_BPT_RATIO);
        if (exactBptIn < MIN_AMOUNT) return;

        uint256 totalSupply = IERC20(address(pool)).totalSupply();
        if (_poolBalance(0) * exactBptIn < totalSupply || _poolBalance(1) * exactBptIn < totalSupply) return;

        medusa.prank(lp);
        try router.removeLiquidityProportional(address(pool), exactBptIn, new uint256[](2), false, bytes("")) returns (
            uint256[] memory amountsOut
        ) {
            assert(amountsOut[0] > 0 && amountsOut[1] > 0);
        } catch {
            // A proportional exit from a non-degenerate pool must always succeed.
            assert(false);
        }
    }
```
Could test the quantity that actually has to round to something non-zero, rather than the balance on its own. (Then we don't need LIVENESS_MIN_BALANCE anymore.)

pkg/pool-gyro/test/foundry/fuzz/DrainECLPSpecific.medusa.sol
```diff
@@ -0,0 +1,571 @@
...
+            ? (snapshot.pool0 + amountIn, snapshot.pool1 - amountOut)
+            : (snapshot.pool0 - amountOut, snapshot.pool1 + amountIn);
+
+        _assertPoolBalances(expected0, expected1);
+    }
+
+    function _assertPoolBalances(uint256 expected0, uint256 expected1) internal view {
+        assert(_poolBalance(0) == expected0);
+        assert(_poolBalance(1) == expected1);
+    }
```
EndymionJkb, line 510:
```suggestion
    function _assertPoolBalances(uint256 expected0, uint256 expected1) internal view {
        (uint256 balance0, uint256 balance1) = _poolBalances();
        assert(balance0 == expected0);
        assert(balance1 == expected1);
    }

    // New helper; could put down with others.
    function _poolBalances() internal view returns (uint256, uint256) {
        (, , uint256[] memory balancesRaw, ) = vault.getPoolTokenInfo(address(pool));
        return (balancesRaw[0], balancesRaw[1]);
    }
```
Every call to `_poolBalance` runs a full `getPoolTokenInfo` just to return a single number, and `_assertPoolBalances` calls it twice, so each operation pays for two complete pool reads where one would do. `_snapshot` already fetches both in one call.

pkg/pool-gyro/test/foundry/fuzz/DrainECLPSpecific.medusa.sol
```diff
@@ -0,0 +1,571 @@
...
+
+    function _direction(uint256 seed) internal pure returns (uint256 indexIn, uint256 indexOut) {
+        return (seed & 1) == 0 ? (uint256(0), uint256(1)) : (uint256(1), uint256(0));
+    }
+
+    function _boundLocal(uint256 x, uint256 min, uint256 max) internal pure returns (uint256) {
+        if (max <= min) return min;
+        return min + (x % (max - min + 1));
+    }
```
EndymionJkb, line 570:
```suggestion
    function _boundLocal(uint256 x, uint256 min, uint256 max) internal pure returns (uint256) {
        if (max <= min) return min;
        uint256 span = max - min;
        return span == type(uint256).max ? min + x : min + (x % (span + 1));
    }
```
There are lots of copies of this. Consider moving this to `BaseMedusaTest` and deleting all four copies. Maybe in another PR?

There's also an overflow issue. When `max` is `type(uint256).max` and `min` is 0, the expression `max - min + 1` overflows to zero, and the modulo on the next line then panics. (No call site here can reach that, because every `max` passed in is a fraction of a balance or of the total supply.) Easier to fix once in a shared copy than four times.

EndymionJkb (approved):
One substantive comment on `property_spotPriceWithinBounds` (can't fail as written), and other minor and comment things.

Issue #1267, opened 2025-01-31
Create a router to add liquidity with underlying to an initialized buffer


joaobrunoah:
ZenDragon mentioned that it's hard to add liquidity to an already initialized buffer, since you can be frontrun when posting the add liquidity transaction and it's hard to tell how much underlying and wrapped token to have. A solution would be to create a router function that makes it easier to add liquidity using only the underlying.

This router should receive underlying amount and wrapped token as argument, and follow the step-by-step:
* Get the buffer balance;
* get the wrapped token rate;
* Calculate the amount of tokens to wrap, so that the add liquidity is proportional
* Wrap tokens using the wrapped token directly
* Add liquidity to the buffer proportionally
* Return "leftovers" to the sender

Yash-arch-ui:
Is this issue solved ! I am ready to take on if not ? Also tell if external contributors are allowed!

EndymionJkb:
It's a public repo; useful contributions are always welcome.

We don't have a lot of bandwidth now, and wouldn't release new contract versions without a compelling reason, so this kind of task - adding a new router - is probably the only way to contribute at this point (at least at the contract level).

Releasing a new router would still be a business decision, as it would need resources devoted to it (audit cycle, SDK, aggregator support, etc.), so no guarantee it would get deployed or used. But if you made a PR adding this, we would certainly look at it.

PR #1666, opened 2026-07-24
Feat/buffer underlying router

Yash-arch-ui:
## Summary

Implements #1267.

Adding liquidity to an already-initialized buffer is
error-prone: callers can be frontrun, and it's hard to know the right
underlying/wrapped split to hold beforehand. This PR adds
`BufferUnderlyingRouter`, which lets a caller add liquidity to a buffer using
**only the underlying token** — the router handles wrapping the correct
proportional amount internally.

### How it works
1. Reads the buffer's current underlying/wrapped balance and total shares.
2. Computes the proportional underlying/wrapped split needed for the caller's
   `exactSharesToIssue` (rounded up, in the Vault's favor).
3. Pulls the total underlying required from the caller (via Permit2).
4. Wraps the proportional amount directly through the ERC4626 token.
5. Deposits both tokens into the buffer for the exact shares requested.
6. Settles debts with the Vault and refunds any leftover dust (on both the
   underlying and wrapped legs) back to the caller.

### Frontrun / slippage protection
`exactSharesToIssue` fixes the shares minted; `maxAmountUnderlyingIn` bounds
the total underlying spent. If the buffer moves enough between submission and
execution that the true cost exceeds the caller's cap, the call reverts
(`UnderlyingAmountInAboveMax`) instead of silently overcharging.

### Design notes
- **Owner resolution**: `sharesOwner` is always `msg.sender` (via `saveSender`,
  consistent with `BufferRouter`) — no separate payer/owner param, for
  simplicity.
- **Custody**: unlike most routers, this one takes physical custody of the
  pulled underlying (direct Permit2 transfer into the router) rather than
  using `_takeTokenIn`, because it needs funds on hand to call
  `wrappedToken.mint()` mid-flow before anything reaches the Vault.
- **Dust handling**: both legs are reconciled and refunded. The wrapped-refund
  branch is defensive/future-proofing — under the current single-transaction
  flow, the Vault's actual wrapped consumption always equals what was minted,
  so this path is untriggered today but guards against future changes to
  Vault-internal accounting.

### Testing
- Happy path: correct shares issued, correct Vault state.
- Slippage revert above max.
- Deadline expiry revert.
- Buffer-not-initialized revert.
- Zero-shares revert.
- Zero router dust after every call (core invariant).
- Underlying-refund reconciliation against real balance changes.
- Query/mutating consistency fuzzed across rates and share amounts
  (10k runs each).

Opening this as a complete, tested implementation per the issue spec — happy
to iterate on feedback, understanding review bandwidth is limited right now.

### DIFF
--- a/pkg/interfaces/contracts/vault/IBufferUnderlyingRouter.sol
+++ b/pkg/interfaces/contracts/vault/IBufferUnderlyingRouter.sol
@@ -1,0 +1,43 @@
+// SPDX-License-Identifier: GPL-3.0-or-later
+pragma solidity ^0.8.24;
+
+import { IERC4626 } from "@openzeppelin/contracts/interfaces/IERC4626.sol";
+
+/// @title IBufferUnderlyingRouter
+/// @notice Router interface for adding liquidity to ERC4626 buffers using underlying tokens,
+/// with frontrun (slippage) protection per issue #1267.
+interface IBufferUnderlyingRouter {
+    /// @notice Add liquidity to an ERC4626 buffer by depositing underlying tokens.
+    /// @dev The router wraps the underlying into the wrapped token, then deposits into the buffer.
+    /// Reverts if the issued shares fall below `minSharesToIssue` (slippage guard).
+    /// Any dust (excess underlying not consumed) is refunded to the caller.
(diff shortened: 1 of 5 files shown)


EndymionJkb:
Closing as not planned. At this point we're not doing new development of any kind without a compelling business case.

PR #1670, opened 2026-08-18
fix: distribute rounding dust in multi-token operations

angguntrie3-lgtm:
## Summary

Adds `computeProportionalAmountsInWithDustDistribution()` to prevent rounding error accumulation in multi-token pools.

## Problem

The current `computeProportionalAmountsIn()` rounds up each token amount independently:

```solidity
amountsIn[i] = balances[i].mulDivUp(bptAmountOut, bptTotalSupply);
```

For pools with many tokens (up to 64 in V3), this can accumulate significant rounding error. Each token rounds up by up to 1 wei, causing users to overpay.

## Impact

- Users pay more than expected when adding liquidity
- Error scales with number of tokens (up to 64 in V3)
- Worst case: 64 wei extra per operation

## Fix

```solidity
function computeProportionalAmountsInWithDustDistribution(
    uint256[] memory balances,
    uint256 bptTotalSupply,
    uint256 bptAmountOut
) internal pure returns (uint256[] memory amountsIn) {
    // Calculate amounts with rounding up
    // Then distribute dust from last token
}
```

## Files Changed

- `pkg/vault/contracts/BasePoolMath.sol`

---

*Found during security audit by Wulansari Security Research.*

### DIFF
--- a/pkg/vault/contracts/BasePoolMath.sol
+++ b/pkg/vault/contracts/BasePoolMath.sol
@@ -71,6 +71,48 @@ library BasePoolMath {
     }
 
     /**
+     * @notice Computes the proportional amounts of tokens to be deposited into the pool
+     * with rounding dust distribution to prevent accumulation errors.
+     * @dev This is an improved version that distributes rounding dust across tokens
+     * to prevent significant accumulation in multi-token pools (up to 64 tokens).
+     * Without this, each token rounds up independently, causing users to overpay.
+     *
+     * @param balances Array of current token balances in the pool
+     * @param bptTotalSupply Total supply of the pool tokens (BPT)
+     * @param bptAmountOut The amount of pool tokens that need to be minted
+     * @return amountsIn Array of amounts for each token to be deposited
+     */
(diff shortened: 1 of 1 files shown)


EndymionJkb:
That rounding is deliberate, and the "fix" would introduce the problem it is meant to prevent.

Rounding each proportional deposit up is what keeps an add-then-remove round trip from profiting, which is an essential security invariant. The comment directly above the function says so: "For security reasons, to help ensure that for all possible 'round trip' paths the caller always receives the same or fewer tokens than supplied, we have chosen the rounding direction to favor the protocol in all cases." Every rounding direction in BasePoolMath is chosen against the caller for that reason.

The file doesn't compile anyway: `mulDiv` is not a member of `FixedPoint`, and `FixedPoint` is the only library attached to `uint256` in BasePoolMath.sol. And V3 pools hold at most 8 tokens, not 64 (`_MAX_TOKENS` in VaultStorage.sol), so the worst case in the description would be 8 wei rather than 64.

PR #1671, opened 2026-08-19
Refactor CLR Recovery Mode, zero-value ops, and buffer minimum handling

EndymionJkb:
# Description

This PR changes three things about how the Composite Liquidity Router handles proportional operations on ERC4626 (boosted) and nested pools.

* The router no longer detects Recovery Mode or performs recovery withdrawals. Every removal takes the ordinary removal path, whatever state the pool is in.

* A nested removal no longer reverts when an unwrapped token's proportional share is exactly zero. That token is reported as zero of its underlying, and the rest of the withdrawal proceeds.

* When the Vault's ERC4626 buffer refuses an amount as below its minimum wrap amount, and that amount was fixed by the pool rather than chosen by the caller, the router reverts with its own error naming the wrapped token and the amount: `UnwrapAmountTooSmall` on the removal paths, `RequiredWrapAmountTooSmall` on the proportional add path. Every other failure is rethrown unchanged.

Also:
* There are new errors in `ICompositeLiquidityRouterErrors`: `UnwrapAmountTooSmall(address, uint256)` and `RequiredWrapAmountTooSmall(address, uint256)`. No function signatures or events change.

* NatSpec now documents: the denomination of `minAmountsOut`/`maxAmountsIn` and of the returned amounts (the tokens the sender actually receives or pays, so query results can be used directly as limits); the errors above and their remedies; the zero-amount behavior; the absence of Recovery Mode handling and the manual procedure that replaces it; and that the queries are quotes rather than simulations (they pass no limits and do not exercise buffer liquidity).

Finally, there are two new suites (`CompositeLiquidityRouterERC4626PoolMinWrapAmount`, `CompositeLiquidityRouterNestedPoolsMinWrapAmount`) run at the production minimum wrap and trade amounts, which the default test Vault sets too low to reach this behavior.

## Type of change

- [ ] Bug fix 
- [ ] New feature 
- [ ] Breaking change 
- [ ] Dependency changes
- [X] Code refactor / cleanup
- [ ] Other

## Checklist:

- [X] Changeset is included 


pkg/interfaces/contracts/vault/ICompositeLiquidityRouter.sol
```diff
@@ -28,6 +28,11 @@ interface ICompositeLiquidityRouter is ICompositeLiquidityRouterErrors {
      * with the wrapped tokens in the ERC4626 pool have been initialized before initializing or adding liquidity to
      * the "parent" pool, and also make sure limits are set properly.
      *
+     * Here the caller names each amount, so an amount the Vault buffer will not wrap is reported as the Vault's own
+     * `WrapAmountTooSmall`, naming the token, and the remedy is to name a larger amount. The proportional operations
+     * report that condition differently only because there the amount comes from the pool and the caller cannot set
+     * it directly.
```
joaobrunoah:
```suggestion
     * Because `exactAmountsIn` is set by the caller, an amount too small for the Vault buffer to wrap reverts with the
     * Vault's `WrapAmountTooSmall`, which identifies only the token; pass a larger amount for that token. (The
     * proportional entry points wrap a pool-derived amount, so they revert with `RequiredWrapAmountTooSmall` instead.)
```

pkg/interfaces/contracts/vault/ICompositeLiquidityRouter.sol
```diff
@@ -67,7 +72,23 @@ interface ICompositeLiquidityRouter is ICompositeLiquidityRouterErrors {
...
+     * `RequiredWrapAmountTooSmall`, which names the token and the amount, and nothing else is charged in its place.
+     * The buffer applies its minimum to the wrapped amount it is given and again to the underlying amount it
+     * calculates, testing the caller's limit between the two, so a limit that token cannot meet is reported as
+     * `SwapLimit` ahead of the second of those but not of the first. A failed call leaves the sender holding their own
+     * tokens. Requesting more pool tokens always raises the required amount, so it clears this for any wrapper as long
+     * as `maxAmountsIn` covers the larger cost; clearing `wrapUnderlying` for that token pays it wrapped instead.
+     * Requesting no pool tokens at all makes no buffer call and charges nothing.
+     *
+     * Note that each token's whole `maxAmountsIn` entry is taken from the sender up front and the unused part is
+     * returned at the end, so the sender must hold and have approved the limit rather than the eventual cost.
```
joaobrunoah: I'm trying to make these comments a bit less verbose.

```suggestion
     * the "parent" pool, and also make sure limits are set properly. Note that `maxAmountsIn` is denominated in the
     * tokens the sender actually pays: the underlying token (e.g., DAI) wherever `wrapUnderlying` is set, and the
     * wrapped token (e.g., waDAI) elsewhere. The returned `amountsIn` use the same denominations, so limits can be
     * derived directly from a query result.
     *
     * Wherever `wrapUnderlying` is set, the amount wrapped is what the pool requires for `exactBptAmountOut` and not
     * an amount the caller chose, so it can fall below the Vault buffer's minimum and revert with
     * `RequiredWrapAmountTooSmall`, which names the token and the amount. Request more pool tokens to raise the
     * required amount (covering the higher cost in `maxAmountsIn`), or clear `wrapUnderlying` for that token to pay
     * it wrapped instead.
     *
     * Note that each token's whole `maxAmountsIn` entry is taken from the sender up front and the unused part is
     * returned at the end, so the sender must hold and have approved the limit rather than the eventual cost.
```

pkg/interfaces/contracts/vault/ICompositeLiquidityRouter.sol
```diff
@@ -88,7 +109,13 @@ interface ICompositeLiquidityRouter is ICompositeLiquidityRouterErrors {
 
     /**
      * @notice Queries an `addLiquidityProportionalToERC4626Pool` operation without actually executing it.
-     * @dev An "ERC4626 pool" contains IERC4626 yield-bearing tokens (e.g., waDAI).
+     * @dev An "ERC4626 pool" contains IERC4626 yield-bearing tokens (e.g., waDAI). Like the operation it quotes, it
+     * reverts `RequiredWrapAmountTooSmall` for a token whose required amount is too small to wrap. It is a quote and
+     * not a simulation: it passes unlimited maximums, so it will not report a limit the caller would have hit, and it
+     * returns the buffer's calculated amount without exercising the buffer's liquidity, so it cannot report a failure
+     * that arises only where the buffer has to mint through the wrapper itself.
+     * See `addLiquidityProportionalToERC4626Pool`.
```
joaobrunoah:
```suggestion
     * @dev An "ERC4626 pool" contains IERC4626 yield-bearing tokens (e.g., waDAI). See
     * `addLiquidityProportionalToERC4626Pool`. Like the operation it quotes, this reverts with
     * `RequiredWrapAmountTooSmall` when a token's required amount is too small to wrap. It is a quote and not a
     * simulation: it passes unlimited maximums and returns the buffer's calculated amount, so it will not surface
     * limit failures, or wrapper failures that only occur when buffer liquidity is short.
```

pkg/vault/test/foundry/CompositeLiquidityRouterNestedPoolsMinWrapAmount.t.sol
```diff
@@ -0,0 +1,347 @@
...
+
+        tokensOut = new address[](3);
+        tokensOut[0] = childTokensOut[0];
+        tokensOut[1] = childTokensOut[1];
+        tokensOut[2] = address(usdc);
+
+        tokensToUnwrap = unwrapList;
+    }
+
+    /// @dev Raw wa6 the child pool would produce for a proportional burn of `parentBptIn` of the parent pool.
```
joaobrunoah, line 307: wa6 does not follow our naming pattern. We use waDAI, waUSDC, and wa6 was confusing (I first read as wrapped amount 6 decimals, it wasn't obvious to me that wa6 is a wrapped token). Can we rename to waUsdc6, or something like this?

pkg/vault/test/foundry/CompositeLiquidityRouterNestedPoolsMinWrapAmount.t.sol
```diff
@@ -0,0 +1,347 @@
...
+            tokensToUnwrap,
+            false,
+            bytes("")
+        );
+    }
+
+    /**
+     * @dev The traversal reaches this unwrap at two levels, and both report the same way. Here the wrapper sits in a
+     * child pool rather than in the pool the caller names, so the amount is the child's share of what the parent's
+     * own removal produced.
```
joaobrunoah, line 214: These are hard for me to follow. I had to read this one several times to work out that "the traversal reaches this unwrap at two levels" just means it's the child-level counterpart to `testNestedSubMinimumBandRevertsWithRouterError`. Can we say it directly?

```suggestion
     * @dev This test unwraps a token at the child level, which should revert with the same router error raised
     * when the unwrap happens in the parent pool (the pool parameter passed by the caller).
```

pkg/vault/test/foundry/CompositeLiquidityRouterNestedPoolsMinWrapAmount.t.sol
```diff
@@ -0,0 +1,347 @@
...
+
+            vm.revertToState(snapshotId);
+        }
+    }
+
+    /**
+     * @dev This path's boundary is a property of the wrapper, exactly as the flat path's is. At a redeem rate of one
+     * half a raw wrapped amount of 15000 clears the Vault's stated minimum of 10000 and is still refused, because the
+     * minimum is applied again to the underlying it redeems to. A router-side check written against
+     * `getMinimumWrapAmount()` would wave this through.
```
joaobrunoah, line 178:
```suggestion
     * @dev The boundary depends on the wrapper's rate, not on the Vault's constant alone. At a redeem rate of 0.5,
     * 15000 wrapped clears the stated minimum of 10000 but redeems to only ~7498 underlying, which does not, and the
     * Vault applies the minimum to both. A router-side pre-check against `getMinimumWrapAmount()` would let this
     * through, and the Vault would reject it anyway.
```

pkg/vault/test/foundry/CompositeLiquidityRouterNestedPoolsMinWrapAmount.t.sol
```diff
@@ -0,0 +1,347 @@
...
+        assertEq(amountsOut[usdc6Idx], 0, "Zero share should report zero");
+        assertGt(amountsOut[daiIdx], 0, "DAI amount should be non-zero");
+    }
+
+    /**
+     * @dev Everything between zero and the first accepted amount still reverts, and the zero guard does not change
+     * that. Unlike zero, these amounts carry a wrapped credit that has no other consumer, so they cannot simply be
+     * skipped. The failure is reported as `UnwrapAmountTooSmall`, naming the token and the amount the pool produced,
+     * which is what the flat ERC4626 remove path reports for the same condition.
+     */
```
joaobrunoah, line 99: The original comment wasn't stating what the test was actually doing.

```suggestion
    /**
     * @dev Sub-minimum wa6 shares revert with `UnwrapAmountTooSmall(wa6, amount)`, and no balance moves. The sampled
     * amounts cover both rejection paths: 1 and 9999 fail the Vault's check on the amount given, while the two just
     * below `_FIRST_ACCEPTED_RAW` clear that and fail the check on the underlying they redeem to. Unlike a zero
     * share (`testNestedZeroUnwrapAtProductionMinimum`), these carry a wrapped credit with no other consumer, so
     * the router cannot skip them.
     */
```

joaobrunoah (approved):
The logic in the files, and tests, they all make sense.

The thing that made me think a lot in this PR is the size and complexity of the comments. Big and detailed comments are not always good, since they remove my attention from the important part and may make me skip the whole text and jump into the code to iunderstand what's happening.

Moreover, some comments were writing numbers by extension (like 0.5 became `one half`), which makes it even harder to read. I don't know, I'd rather have smaller comments, if possible, and more straight forward ones. It's better to say that `unwrap happens in the child pool` then to say `traversal reaches this unwrap at two levels`.

Another example: the buffer's minimum wrap amount is explained in 13 separate passages in 4 files: 9 in the interface, 1 in the errors file, 2 in the hooks implementation, 1 in the test base. Making comments bigger and more detailed will make the maintainance of the code harder, with more parts to get stale in the future. I'd rather follow a bit of clean code in this case, and leave the code explaining it, if we can. I understand that it may mean more false tickets on immunefy, but we can't compromise the readability and maintainability of the code to avoid those tickets. If anything, these detailed comments could be in a separated readme, targetting AI. And we just mention in the code (Detailed explanation on README-AI.md, or something)

joaobrunoah (approved):
LGTM!

